PR #60, opened 2017-11-23
Add multistage Dockerfile for automated build

lalyos:


matyix:
@lalyos `latest` tag works fine.

PR #81, opened 2017-12-01
Document deps management, with tooling installation

lalyos:


DEPS.md
```diff
@@ -0,0 +1,64 @@
...
+- [glide-vc](github.com/sgotti/glide-vc) vendor cleaner plugin for glide
+
+`make deps` will instal them, in case they are missing
+
+## Add a new dependency
+
+If you write new features which imports a new library, you have to vendor it:
+```
+glide get -v github.com/Masterminds/cookoo/web
+glide-vc --only-code
```
matyix: Shall we not add `--no-tests` as well to remove tests from vendor?

DEPS.md
```diff
@@ -0,0 +1,64 @@
...
+```
+- package: github.com/prometheus/prometheus
+  version: ^2.0.0
+```
+
+than perform an update:
+
+```
+glide up -v
+glide-vc --only-code
```
matyix: Same here re `--no-tests`?

DEPS.md
lalyos: good catch

DEPS.md
lalyos: same here

PR #124, opened 2018-01-10
FetchClusterConfig returns base64 encoded config in case of azure

baluchicken:


cloud/azure.go
```diff
@@ -287,7 +287,8 @@ func GetAzureK8SConfig(cs *banzaiSimpleTypes.ClusterSimple, c *gin.Context) {
 		writeConfig2File(fmt.Sprintf("./statestore/%s", cs.Name), config)
 
 		banzaiUtils.LogInfo(banzaiConstants.TagFetchClusterConfig, "Get k8s config succeeded")
-		SetResponseBodyJson(c, http.StatusOK, config)
+		decodedConfig, _ := base64.StdEncoding.DecodeString(config.Properties.KubeConfig)
```
stoader: Please handle the error as if without a valid kube config we can not proceed further. Also log the content of k8s config in case of error such as we can investigate what caused the error if there is any. If base64.StdEncoding.DecodeString already logs the config causing failure no need to log it again

stoader:
LGMT

PR #128, opened 2018-01-11
Master and node bootstrap scripts came from env variables

colin014:


matyix:
@colin014 for long term generic solution - https://github.com/hashicorp/go-getter

PR #160, opened 2018-01-23
Ability to add arbitrary Helm charts with Pipeline ...

baluchicken:
besides the predefined `spotguides` one.


main.go
```diff
@@ -538,8 +532,8 @@ func updatePrometheusPostHook(_ *banzaiSimpleTypes.ClusterSimple, _ *gin.Context
 
 func installHelmPostHook(createdCluster *banzaiSimpleTypes.ClusterSimple, c *gin.Context) {
 	logTag := "InstallHelmPostHook"
-	retryAttempts := viper.GetInt("dev.retryAttempt")
-	retrySleepSeconds := viper.GetInt("dev.retrySleepSeconds")
+	retryAttempts := viper.GetInt("helm.retryAttempt")
```
stoader: `helm.retryAttempt` and `helm.retrySleepSeconds` should be constant.

matyix: I believe I've asked to make configurable due to some AKS issues, but constants are fine as well,

matyix: @baluchicken how about dealing with the goroutines (for all hooks)? Will open a Jira ticket for that.

stoader: I meant to the `helm.retryAttempt` literal to be a constant like `const HELM_RETRY_ATTEMPT_CONFIG="helm.retryAttempt"`

PR #162, opened 2018-01-24
Reimplemented the deployment status call using the specialized intern…

lpuskas:
…al api call
This PR fixes the behaviour of the rest calls of the form:

```
HEAD /api/v1/clusters/2/deployments/dep_name HTTP/1.1
```
The implementation uses a the specialized internal api call.
HTTP response codes fixed to be consistent with the behaviour (further refinements based on deployment statuses may be required)

### DIFF
--- a/main.go
+++ b/main.go
@@ -945,6 +945,7 @@ func HelmDeploymentStatus(c *gin.Context) {
 	// todo error handling - design it, refine it, refactor it
 
 	name := c.Param("name")
+	banzaiUtils.LogInfof("HelmDeploymentStatus", "Retrieving status for deployment: %s", name)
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -942,3 +945,49 @@ func InitHelmOnCluster(c *gin.Context) {
...
+		cloud.SetResponseBodyJson(c, http.StatusOK, gin.H{
+			cloud.JsonKeyStatus:  http.StatusOK,
+			cloud.JsonKeyMessage: "Deployment status: " + status,
+		})
+	}
+
+}
+
+func helmDeploymentStatusErrorResponse(c *gin.Context, err error) {
+	if err != nil {
```
stoader: Write the error to the log as well.

PR #163, opened 2018-01-25
Add initial endpoint listing

matyix:


main.go
```diff
@@ -266,6 +270,73 @@ func CreateDeployment(c *gin.Context) {
...
+	// --- [ Get K8S Config ] --- //
+	kubeConfig, err := cloud.GetK8SConfig(cloudCluster, c)
+	if err != nil {
+		return
+	}
+	apiconfig, _ := clientcmd.Load(kubeConfig)
+	clientConfig := clientcmd.NewDefaultClientConfig(*apiconfig, &clientcmd.ConfigOverrides{})
+	config, err := clientConfig.ClientConfig()
+	if err != nil {
+		banzaiUtils.LogError(banzaiConstants.TagKubernetes, "Could not create kubernetes client from config.")
```
stoader: Log the content of the config that caused the error.

main.go
```diff
@@ -266,6 +270,73 @@ func CreateDeployment(c *gin.Context) {
...
+		})
+		return
+	}
+	serviceList, err := client.CoreV1().Services("").List(meta_v1.ListOptions{})
+	for _, service := range serviceList.Items {
+		banzaiUtils.LogDebugf(banzaiConstants.TagKubernetes, "Service: %#v", service.Status)
+		if len(service.Status.LoadBalancer.Ingress) > 0 {
+			endpointList = append(endpointList, &EndpointItem{
+				Name: service.Name,
+				Host: service.Status.LoadBalancer.Ingress[0].Hostname,
```
stoader: Does this work AKS as well? Apparently on AKS the Azure LB created for LoadBalancer type Kubernetes services has a (static) IP instead contrary to AWS.

baluchicken: No it does not works on AKS because there the IP is used instead of Hostname

baluchicken:
Rebased and squashed into one commit.

PR #167, opened 2018-01-25
Delete statestore fix in case of Azure

colin014:


main.go
```diff
@@ -485,7 +485,12 @@ func DeleteCluster(c *gin.Context) {
 		}
 	}
 	if cloud.DeleteCluster(cl, c) {
-		// cluster delete success, delete from db
+		// cluster delete success
+
+		// delete sate store
```
baluchicken: typo:) state store

PR #168, opened 2018-01-25
fix status endpoint config retrieval

sancyx:


pods/kubernetes.go
```diff
@@ -147,23 +148,26 @@ func CreatePodMap(client *kubernetes.Clientset, namespace string) (map[string][]
...
 }
 
-func ListPodsForCluster(cluster *cluster.Cluster) (ClusterStatusResponse, error) {
+func ListPodsForCluster(cluster *banzaiSimpleTypes.ClusterSimple) (ClusterStatusResponse, error) {
 	var (
 		config *rest.Config
 		err    error
 	)
-	kubeConfig, err := cloud.GetConfig(cluster, "")
+	kubeConfig, err := cloud.GetKubeConfigPath(fmt.Sprintf("./statestore/%s/", cluster.Name))
```
baluchicken: This is not kubeConfig any more, this is a kubeConfigPath

baluchicken: I prefer to use cloud.GetK8SConfig() instead of GetKubeConfigPath(), it returns a kubeconfig and not the path and use this code snippet to generate the config.
`if len(kubeConfig) > 0 {
		apiconfig, _ :=clientcmd.Load(kubeConfig)
		clientConfig := clientcmd.NewDefaultClientConfig(*apiconfig, &clientcmd.ConfigOverrides{})
		config, err = clientConfig.ClientConfig()
	} else {
		banzaiUtils.LogInfo(banzaiConstants.TagKubernetes, "Use K8S InCluster Config.")
		config, err = rest.InClusterConfig()
	}`

PR #227, opened 2018-03-04
set permission for object storage

sancyx:
grant read/write for object storage


sancyx:
If set the some service account other then default, it will not work, you need to set the scope as well.

sancyx:
It seems setting scopes still is required even if service account is set.

PR #229, opened 2018-03-05
Add service account in case of gke

colin014:


baluchicken:
Plz wait until we test this.

PR #241, opened 2018-03-07
Minor fixes

colin014:


cluster/gke.go
```diff
@@ -950,14 +950,14 @@ func (g *GKECluster) AddDefaultsToUpdate(r *components.UpdateClusterRequest) {
 
 	// ---- [ Master check ] ---- //
 	if r.GoogleMaster == nil {
-		log.Info(constants.TagValidateCreateCluster, "'master' field is empty. Load it from stored data.")
+		log.Info("'master' field is empty. Load it from stored data.")
```
stoader: Is this a warning message? if so we should log it as warning.

baluchicken: The log is not initialised properly on this function pls initialise it.

PR #275, opened 2018-03-13
Fix update cluster method in case of GKE

colin014:


cluster/gke.go
```diff
@@ -1012,6 +1025,17 @@ func (g *GKECluster) AddDefaultsToUpdate(r *components.UpdateClusterRequest) {
 		log.Warn("Node count set to default value: ", def)
 		r.UpdateClusterGoogle.GoogleNode.Count = def
 	}
+
+	// ---- [ Node Version check] ---- //
+	if len(r.UpdateClusterGoogle.GoogleNode.Version) == 0 {
+		r.UpdateClusterGoogle.GoogleNode.Version = g.modelCluster.Google.NodeVersion
```
baluchicken: Don't we need warn logging here? as above?

cluster/gke.go
```diff
@@ -1012,6 +1025,17 @@ func (g *GKECluster) AddDefaultsToUpdate(r *components.UpdateClusterRequest) {
...
+	}
+
+	// ---- [ Master Version check] ---- //
+	if len(r.UpdateClusterGoogle.GoogleMaster.Version) == 0 {
+		r.UpdateClusterGoogle.GoogleMaster.Version = g.modelCluster.Google.MasterVersion
```
baluchicken: Don't we need warn logging here? as above?

PR #287, opened 2018-03-19
Update default gke master/node version

colin014:
`1.8.7-gke.1` master version is no longer supported by Google. Current default cluster version is `1.8.8-gke.0`.


baluchicken:
LGTM thanks @colin014 :)

Issue #314, opened 2018-03-23
Before delete a secret, check there's a running cluster with that secret


colin014:

colin014:
fixed in: #320

PR #320, opened 2018-03-26
Check cluster(s) before delete a secret

colin014:
#314

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -60,7 +60,7 @@ func GetCommonClusterFromRequest(c *gin.Context) (cluster.CommonCluster, bool) {
 	}
 
 	if len(modelCluster) == 0 {
-		log.Errorf("Empty cluster list")
+		log.Error("Empty cluster list")
 		c.JSON(http.StatusNotFound, components.ErrorResponse{
 			Code:    http.StatusNotFound,
 			Message: "Cluster not found",
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -58,7 +58,18 @@ func GetCommonClusterFromRequest(c *gin.Context) (cluster.CommonCluster, bool) {
 		})
 		return nil, false
 	}
-	commonCLuster, err := cluster.GetCommonClusterFromModel(modelCluster)
+
+	if len(modelCluster) == 0 {
+		log.Errorf("Empty cluster list")
+		c.JSON(http.StatusNotFound, components.ErrorResponse{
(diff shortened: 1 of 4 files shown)


api/cluster.go
```diff
@@ -58,7 +58,18 @@ func GetCommonClusterFromRequest(c *gin.Context) (cluster.CommonCluster, bool) {
 		})
 		return nil, false
 	}
-	commonCLuster, err := cluster.GetCommonClusterFromModel(modelCluster)
+
+	if len(modelCluster) == 0 {
+		log.Errorf("Empty cluster list")
```
baluchicken: log.Error

api/secrets.go
```diff
@@ -184,3 +195,25 @@ func IsValidSecretType(secretType string) error {
...
+// checkClustersBeforeDelete returns error if there's a running cluster that created with the given secret
+func checkClustersBeforeDelete(orgId, secretId string) error {
+
+	filter := map[string]interface{}{
+		"organization_id": orgId,
+		"secret_id":       secretId,
+	}
+
+	if modelCluster, err := model.QueryCluster(filter); err != nil {
+		return nil
```
baluchicken: Handle error properly here.

api/secrets.go
```diff
@@ -184,3 +195,25 @@ func IsValidSecretType(secretType string) error {
...
+	} else {
+		for _, mc := range modelCluster {
+			if commonCluster, err := cluster.GetCommonClusterFromModel(&mc); err == nil {
+				if _, err := commonCluster.GetStatus(); err == nil {
+					return errors.New(fmt.Sprintf("There's a running cluster with this secret: %s[%d]", mc.Name, mc.ID))
```
baluchicken: Use fmt.Errorf instead of errors.new and fmt.Sprintf

secret/store.go
```diff
@@ -157,6 +157,11 @@ func (ss *secretStore) Get(organizationID string, secretID string) (*SecretsItem
 	if err != nil {
 		return nil, err
 	}
+
+	if secret == nil {
+		return nil, errors.New(fmt.Sprintf("There's no secret with this id: %s", secretID))
```
baluchicken: Use fmt.Errors

api/secrets.go
colin014: When an error occours here that means no cluster found the database with the given params and the secret can be deleted so the method sould be return nil error here. But i add an info log before return

PR #323, opened 2018-03-26
Introduce access Token type, and Token API

bonifaido:


main.go
```diff
@@ -138,7 +138,10 @@ func main() {
 			orgs.GET("/:orgid/allowed/secrets/:type", api.ListAllowedSecretTypes)
 		}
 		//v1.GET("/clusters/gke/:projectid/:zone/serverconf", cluster.GetGkeServerConfig) // todo think about it and move
-		v1.GET("/token", auth.GenerateToken)
+		v1.GET("/token", auth.GenerateToken) // Should be POST /tokens
```
matyix: Maybe be more specific that this is generating a token (like `/generateToken`) or change to POST

auth/authn.go
```diff
@@ -205,6 +208,59 @@ func GenerateToken(c *gin.Context) {
 	}
 }
 
+func GetTokens(c *gin.Context) {
```
baluchicken: Add docs

auth/tokenstore.go
```diff
@@ -3,71 +3,105 @@ package auth
...
+
+// NewToken Creates a new Token instance initialized ID and Name and CreatedAt fields
+func NewToken(id, name string) *Token {
+	createdAt := time.Now()
+	return &Token{ID: id, Name: name, CreatedAt: createdAt}
+}
+
+func parseToken(secret *vaultapi.Secret) (*Token, error) {
+	if secret == nil {
+		return nil, nil
```
baluchicken: Shouldn't we return some error here?

auth/user.go
```diff
@@ -67,6 +67,10 @@ type Organization struct {
 	Clusters  []model.ClusterModel `gorm:"foreignkey:organization_id" json:"clusters,omitempty"`
 }
 
+func (user *User) IDString() string {
```
baluchicken: Docs

main.go
bonifaido: Once the UI can handle token generation I will change it to POST.

bonifaido:
Rebased.

PR #333, opened 2018-03-28
Clean statestore after cluster delete

colin014:


cluster/gke.go
```diff
@@ -818,7 +819,7 @@ func storeConfig(c *kubernetesCluster, name string) ([]byte, error) {
 		token = c.ServiceAccountToken
 	}
 
-	configFile := fmt.Sprintf("./statestore/%s/config", name)
+	configFile := fmt.Sprintf("%s/%s/config", viper.GetString("statestore.path"), name)
```
stoader: I'd factor out into a function the returns the state store path of a cluster as the `fmt.Sprintf("%s/%s", viper.GetString("statestore.path")` part is repeated in multiple places

Issue #339, opened 2018-04-03
Supported cluster providers


colin014:
To List available clusters we need an endpoint to describe this.
This can be a read only endpoint.
The response should include:
- [x] Cloud name 
- [x] Key to cloud type
- [x] Regexp (name) to create cluster
- [x] Locations
- [x] Machine types
- [x] Kubernetes versions

PR #351, opened 2018-04-09
Supported providers

colin014:
#339


cluster/gke.go
```diff
@@ -10,13 +10,13 @@ import (
...
 	"github.com/banzaicloud/pipeline/secret"
 	"github.com/banzaicloud/pipeline/utils"
-	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/json"
 	"github.com/go-errors/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2/google"
 	gke "google.golang.org/api/container/v1"
+	"google.golang.org/api/compute/v1"
```
baluchicken: this one is wrong here, you can check out that v1 package is already imported in https://github.com/banzaicloud/pipeline/blob/53b3b1a82368291c2eef994c64145f1d32e6e2e7/cluster/gke.go#L23

cluster/aks.go
```diff
@@ -365,3 +365,49 @@ func (c *AKSCluster) DeleteFromDatabase() error {
...
+func GetMachineTypes(orgId uint, secretId, location string) (response map[string]components.MachineType, err error) {
+	client, err := getAKSClient(orgId, secretId)
+	if err != nil {
+		return nil, err
+	}
+
+	response = make(map[string]components.MachineType)
+	response[location], err = azureClient.GetVmSizes(client, location)
+
+	return
```
baluchicken: This function does not return anything

cluster/aws.go
```diff
@@ -783,3 +785,54 @@ func getKubicornLogLevel() int {
 		return 4
 	}
 }
+
+// ListRegions lists supported regions
+func ListRegions() ([]*ec2.Region, error) {
+	// Load session from shared config
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
```
baluchicken: Maybe this can be factored out getting the session and creating a new EC2 client ListAMI and ListRegion does the same thing

cluster/supported/aks.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/banzaicloud/banzai-types/constants"
+	"github.com/banzaicloud/banzai-types/components"
+)
+
+type AzureInfo struct {
+	BaseFields
+}
+
+// GetType returns cloud type
+func (a *AzureInfo) GetType() string {
```
baluchicken: I don't really understand why are these methods? it never uses the AzureInfo

cluster/supported/base.go
```diff
@@ -0,0 +1,128 @@
...
+				SecretId: r.SecretId,
+			},
+		}, nil
+
+	default:
+		return nil, constants.ErrorNotSupportedCloudType
+	}
+}
+
+func ProcessFilter(p CloudInfoProvider, r *components.CloudInfoRequest) (*components.GetCloudInfoResponse, error) {
```
baluchicken: Doc is missing

cluster/supported/base.go
```diff
@@ -0,0 +1,128 @@
...
+		NameRegexp: p.GetNameRegexp(),
+	}
+	if r != nil && r.Filter != nil {
+		for _, field := range r.Filter.Fields {
+			switch field {
+
+			case constants.KeyWorldLocation:
+				if l, err := p.GetLocations(); err != nil {
+					return nil, err
+				} else {
```
baluchicken: we don't need else here, because the if returns

utils/utils.go
```diff
@@ -75,3 +75,7 @@ func ConvertJson2Map(js []byte) (map[string]string, error) {
 	err := json.Unmarshal(js, &result)
 	return result, err
 }
+
+func S(s string) *string {
```
baluchicken: why are we using string pointers instead of string? I did not see any place where we are comparing to nil. Remove it.

cluster/aks.go
colin014: This is a "naked" return :)

cluster/supported/aks.go
colin014: The 3 current supported clouds are Google (GoogleInfo), Azure (AzureInfo) and Amazon (AmazonInfo) which queried the supported locations, machine types and Kubernetes versions. These 3 structs helps in that. Here the creation: https://github.com/banzaicloud/pipeline/blob/master/cluster/supported/base.go#L44

Issue #372, opened 2018-04-11
NodePool support for AKS


matyix:
Similar to this GKE feature we added - [node pool support](https://github.com/banzaicloud/pipeline/pull/357) we need it for AKS as well (agent pools).

>On AWs we use multi ASG, no need to change there *yet*

PR #385, opened 2018-04-13
Add nodepool support in case of azure

colin014:
#372

### DIFF
--- a/cluster/aks.go
+++ b/cluster/aks.go
@@ -388,7 +388,7 @@ func (c *AKSCluster) AddDefaultsToUpdate(r *components.UpdateClusterRequest) {
 		r.Azure = &banzaiAzureTypes.UpdateClusterAzure{}
 	}
 
-	if r.Azure.NodePools == nil && len(r.Azure.NodePools) == 0 {
+	if len(r.Azure.NodePools) == 0 {
 		storedPools := c.modelCluster.Azure.NodePools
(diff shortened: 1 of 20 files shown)


cluster/aks.go
```diff
@@ -299,41 +383,45 @@ func CreateAKSClusterFromModel(clusterModel *model.ClusterModel) (*AKSCluster, e
...
-		r.UpdateClusterAzure = &banzaiAzureTypes.UpdateClusterAzure{}
+		r.Azure = &banzaiAzureTypes.UpdateClusterAzure{}
 	}
 
-	// ---- [ Node check ] ---- //
-	if r.UpdateAzureNode == nil {
-		log.Info("'node' field is empty. Load it from stored data.")
-		r.UpdateAzureNode = &banzaiAzureTypes.UpdateAzureNode{
-			AgentCount: c.modelCluster.Azure.AgentCount,
+	if r.Azure.NodePools == nil && len(r.Azure.NodePools) == 0 {
```
baluchicken: Do we really need both checks? Isn't len(x) enough?

colin014: `NodePools` field is a map and can be nil. If we just check len(x) then (if `r.Azure.NodePools` is nil) this can cause nil pointer exception

colin014: My mistake :) `len(nil) == 0` is true

Issue #393, opened 2018-04-16
Heterogeneous cluster nodes support for AKS cluster profiles


colin014:

matyix:
closed by #396

Issue #358, opened 2018-04-10
Heterogeneous cluster nodes support for GKE cluster profiles


stoader:
- [x] Create GKE Profile
- [x] Update GKE Profile
- [x] Get GKE Profile

colin014:
closed by #396

PR #396, opened 2018-04-16
Update profiles

colin014:
#358 #393

### DIFF
--- a/model/defaults/aks.go
+++ b/model/defaults/aks.go
@@ -12,16 +12,64 @@ import (
 // AKSProfile describes an Azure cluster profile
 type AKSProfile struct {
 	DefaultModel
-	Location          string `gorm:"default:'eastus'"`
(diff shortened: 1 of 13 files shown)


model/defaults/aks.go
```diff
@@ -12,18 +12,66 @@ import (
...
+	Location          string                `gorm:"default:'eastus'"`
+	KubernetesVersion string                `gorm:"default:'1.9.2'"`
+	NodePools         []*AKSNodePoolProfile `gorm:"foreignkey:Name"`
+}
+
+// AKSNodePoolProfile describes an Azure cluster profile's nodepools
+type AKSNodePoolProfile struct {
+	ID               uint   `gorm:"primary_key"`
+	Count            int    `gorm:"default:1"`
+	NodeInstanceType string `gorm:"default:'Standard_D2_v2'"`
```
matyix: Change this to D4, usually D2 is not enough for the workload types we run

PR #400, opened 2018-04-16
add multi asg

sancyx:
Temporarly removed AMI validation, because takes half minute to get all images.


cluster/aws.go
```diff
@@ -230,50 +246,48 @@ func getStateStoreForCluster(clusterType *model.ClusterModel) (stateStore state.
...
-			Type:       kcluster.NetworkTypePublic,
-			CIDR:       "10.0.0.0/16",
-			InternetGW: &kcluster.InternetGW{},
-		},
-		Values: &kcluster.Values{
-			ItemMap: map[string]string{
-				"INJECTEDTOKEN": kubeadm.GetRandomToken(),
-			},
+func GetMasterServerPool(cs *model.ClusterModel, nodeServerPool []*kcluster.ServerPool, uuidSuffix string) *kcluster.ServerPool {
+	var ingressRules = make([]*kcluster.IngressRule, 2+len(nodeServerPool))
```
baluchicken: Why don't you use the append function here for the slice? It will remove the 2+ ...

cluster/aws.go
```diff
@@ -30,6 +29,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
baluchicken: Doc is missing from all public functions

cluster/aws.go
```diff
@@ -454,12 +534,9 @@ func (c *AWSCluster) UpdateCluster(request *components.UpdateClusterRequest) err
 		SecretId:         c.modelCluster.SecretId,
 		Status:           c.modelCluster.Status,
 		Amazon: model.AmazonClusterModel{
-			NodeSpotPrice:      c.modelCluster.Amazon.NodeSpotPrice,
-			NodeMinCount:       request.Amazon.MinCount,
-			NodeMaxCount:       request.Amazon.MaxCount,
-			NodeImage:          c.modelCluster.Amazon.NodeImage,
 			MasterInstanceType: c.modelCluster.Amazon.MasterInstanceType,
 			MasterImage:        c.modelCluster.Amazon.MasterImage,
+			NodePools:          updatedNodePools,
```
baluchicken: This variable is just initialised but holds no value, is this right?:)

cluster/aws.go
```diff
@@ -782,9 +859,11 @@ func getKubicornLogLevel() int {
 }
 
 // ListRegions lists supported regions
-func ListRegions() ([]*ec2.Region, error) {
+func ListRegions(region string) ([]*ec2.Region, error) {
 
-	svc := newEC2Client(nil)
+	svc := newEC2Client(&aws.Config{
+		Region: &region,
```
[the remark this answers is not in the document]
sancyx: I got the following error here: `could not find region configuration` without specifying the region.

cluster/aws.go
sancyx: Update part will be refactored in other patch

cluster/aws.go
sancyx: fixed

cluster/aws.go
sancyx: reset to private

Issue #412, opened 2018-04-19
Refactor GetClusterStatusResponse


colin014:
Remove `google.NodePool` from `GetClusterStatusResponse` and introduce `StatusNodePool` with fields:

- [ ] Count
- [ ] InsatnceType
- [ ] ServiceAccount
- [ ] SpotPrice
- [ ] MinCount
- [ ] MaxCount
- [ ] Image

PR #419, opened 2018-04-20
Remove unnecessary NodeInstanceType fields and refactor statusnodepool

colin014:
#412 #413 #415


cluster/aws.go
```diff
@@ -486,13 +484,24 @@ func GetKubicornProfile(cs *model.ClusterModel) *kcluster.Cluster {
 func (c *AWSCluster) GetStatus() (*components.GetClusterStatusResponse, error) {
 	log.Info("Start get cluster status (amazon)")
 
+	nodePools := make(map[string]*components.NodePoolStatus)
+	for _, np := range c.modelCluster.Amazon.NodePools {
```
baluchicken: In case of azure nil np was checked, is that required here also?

cluster/gke.go
```diff
@@ -267,15 +259,22 @@ func (g *GKECluster) GetStatus() (*components.GetClusterStatusResponse, error) {
 	log := logger.WithFields(logrus.Fields{"action": constants.TagGetClusterStatus})
 	log.Info("Create cluster status response")
 
-	nodePools, _ := createNodePoolsRequestDataFromNodePoolModel(g.modelCluster.Google.NodePools)
+	nodePools := make(map[string]*components.NodePoolStatus)
+	for _, np := range g.modelCluster.Google.NodePools {
+		nodePools[np.Name] = &components.NodePoolStatus{
```
baluchicken: nil check is not required here like you did in case of azure?

cluster/supported/aws.go
```diff
@@ -10,6 +10,8 @@ type AmazonInfo struct {
 	BaseFields
 }
 
+var defaultRegion = "eu-west-1"
```
baluchicken: this can be a const

cluster/aws.go
colin014: yes, i fixed it

Issue #438, opened 2018-04-24
Tiller install fails when kubernetes version > 1.8


baluchicken:
The relevant warn maybe the following:
```
INFO[11021] Waiting 0/30                                  tag=RetryHelmInstall
INFO[11021] Setting up helm directories.                  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore//azclusterbaluchicken259/helm'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/repository'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/repository/cache'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/repository/local'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/plugins'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/starters'  tag=EnsureHelmDirectories
INFO[11021] Creating '/Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/cache/archive'  tag=EnsureHelmDirectories
INFO[11021] Setting up default helm repos.                tag=EnsureDefaultRepositories
INFO[11021] Creating /Users/baluchicken/go/src/github.com/banzaicloud/pipeline/statestore/azclusterbaluchicken259/helm/repository/repositories.yaml  tag=EnsureDefaultRepositories
INFO[11021] Adding stable repo with URL: https://kubernetes-charts.storage.googleapis.com  tag=InitHelmRepositories
INFO[11023] Adding banzaicloud-stable repo with URL: http://kubernetes-charts.banzaicloud.com  tag=InitHelmRepositories
INFO[11023] Initializing helm client succeeded, happy helming!  tag=InstallHelmClient
[...]

PR #440, opened 2018-04-24
Fix tiller install in case of kubernetes version >1.8

baluchicken:
#438


baluchicken:
After further discussion we support only kubernetes version 1.8 and above.

PR #444, opened 2018-04-25
Validate fields after cluser persist

colin014:
#442

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -199,19 +199,6 @@ func CreateCluster(c *gin.Context) {
 	}
 	log.Info("Secret validation passed")
 
-	log.Info("Validate creation fields")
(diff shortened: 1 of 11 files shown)


api/cluster.go
```diff
@@ -228,12 +215,22 @@ func CreateCluster(c *gin.Context) {
 		ResourceID: commonCluster.GetID(),
 	})
 
-	go postCreateCluster(commonCluster)
+	go postCreateCluster(commonCluster, &createClusterRequest)
```
stoader: If we run the validation in a separate go routine than how the validation error is returned back to the caller?

baluchicken:
After further discussion LGTM:) can you please rebase:)

colin014:
Please dont merge yet. I add a new status message field.

Issue #439, opened 2018-04-24
Add new secret type called repo


baluchicken:
Create a new secret type API which stores secret with  path org/repo in Vault

PR #447, opened 2018-04-25
Add new secrettype repo and allow to list and create them

baluchicken:
#439


baluchicken:
Please let @tarokkk to merge this, because it may contain api changes which breaks the UI.

PR #466, opened 2018-05-03
Fix filtered listEndpoints API

baluchicken:


bonifaido:
There is test fail: TestPendingLoadBalancer

Issue #514, opened 2018-05-30
Secret handling reconciliation


bonifaido:
- [x] Attachability: all secret types should attachable to a repository (via PATCH).
- [x] Keys: secret keys should be available in via the API, only the values should be hidden.
- [x] Types: repo secret type should be deprecated, instead there should be a generic secret type which can hold any number of values, but without a schema.
- [x] Tagging: repo and category tags should be added (repo:banzaicloud/nodejs, category:cloud)
- [ ] Patching: secrets can't be updated right now.
- [ ] Swagger: update it accordingly

PR #519, opened 2018-05-31
Secret handling reconciliation 

bonifaido:
Fixes: #514

### DIFF
--- a/api/secrets.go
+++ b/api/secrets.go
@@ -45,7 +45,6 @@ func AddSecrets(c *gin.Context) {
 	}
 
 	log.Info("Binding request succeeded")
-	log.Debugf("%#v", createSecretRequest)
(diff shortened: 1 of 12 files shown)


api/secrets.go
```diff
@@ -73,9 +72,69 @@ func AddSecrets(c *gin.Context) {
...
+		Name: createSecretRequest.Name,
+		Type: createSecretRequest.Type,
+		ID:   secretID,
+	})
+}
+
+// UpdateSecrets update the given secret to vault
+func UpdateSecrets(c *gin.Context) {
+
+	log := logger.WithFields(logrus.Fields{"tag": "Create Secrets"})
```
baluchicken: I think wrong tag name it should be update secrets

PR #529, opened 2018-06-01
Ssh-key generator with Bank Vaults support

### LINKED ISSUE #527: Store K8S config in vault
- [x] Save config after cluster created
- [x] Load config from vault

ahma:
Fixes #527

### DIFF
--- a/cluster/aws.go
+++ b/cluster/aws.go
@@ -13,10 +13,10 @@ import (
 	"github.com/banzaicloud/pipeline/config"
 	"github.com/banzaicloud/pipeline/model"
 	"github.com/banzaicloud/pipeline/secret"
+	pipelineSsh "github.com/banzaicloud/pipeline/ssh"
 	"github.com/banzaicloud/pipeline/utils"
 	kcluster "github.com/kubicorn/kubicorn/apis/cluster"
 	"github.com/kubicorn/kubicorn/pkg"
-	"github.com/kubicorn/kubicorn/pkg/initapi"
 	"github.com/kubicorn/kubicorn/pkg/kubeadm"
 	kubicornLogger "github.com/kubicorn/kubicorn/pkg/logger"
 	"github.com/kubicorn/kubicorn/pkg/uuid"
@@ -185,20 +185,14 @@ func (c *AWSCluster) CreateCluster() error {
 
 	//TODO check if this should be private
 	c.kubicornCluster = GetKubicornProfile(c.modelCluster)
-	sshKeyPath := viper.GetString("cloud.keypath")
-	//TODO move to the profile section
-	if sshKeyPath != "" {
-		log.Debug("Overwriting default SSH key path to:", sshKeyPath)
-		c.kubicornCluster.SSH.PublicKeyPath = sshKeyPath
-	}
 
-	log.Info("Init cluster")
-	newCluster, err := initapi.InitCluster(c.kubicornCluster)
+	sshSecretID, sshKey, err := pipelineSsh.KeyAdd(c.GetModel().OrganizationId, c.GetModel().ID)
+	c.GetModel().Amazon.SshSecretID = sshSecretID
 
-	if err != nil {
-		return err
-	}
+	c.kubicornCluster.SSH.PublicKeyData = []byte(sshKey.PublicKeyData)
+	c.kubicornCluster.SSH.PublicKeyFingerprint = sshKey.PublicKeyFingerprint
 
+	newCluster := c.kubicornCluster
 	log.Info("Get reconciler")
 	reconciler, err := pkg.GetReconciler(newCluster, &runtimeParam)
 
@@ -591,6 +585,7 @@ func (c *AWSCluster) UpdateCluster(request *components.UpdateClusterRequest) err
 			MasterInstanceType: c.modelCluster.Amazon.MasterInstanceType,
 			MasterImage:        c.modelCluster.Amazon.MasterImage,
 			NodePools:          updatedNodePools,
+			SshSecretID:        c.modelCluster.Amazon.SshSecretID,
 		},
 	}
 
@@ -821,7 +816,7 @@ func (c *AWSCluster) GetK8sConfig() ([]byte, error) {
 		err = errors.Wrap(err, "error getting kubicorn cluster")
 		return nil, err
 	}
-	kubeConfig, err := DownloadK8sConfig(kubicornCluster)
+	kubeConfig, err := DownloadK8sConfig(kubicornCluster, fmt.Sprint(c.GetModel().OrganizationId), c.GetModel().ID)
 	if err != nil {
 		err = errors.Wrap(err, "error downloading kubernetes config")
 		return nil, err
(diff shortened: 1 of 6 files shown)


secret/store_test.go
```diff
@@ -95,6 +97,14 @@ const (
 	gkeClientCert   = "client_x509_cert_url"
 )
 
+const (
```
bonifaido: I think you can reuse the constants from the package, no need to repeat them here.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+
+var logger *logrus.Logger
+var log *logrus.Entry
+
+// Simple init for logging
+func init() {
+	logger = config.Logger()
+}
+
+type Key struct {
```
bonifaido: Please add comment to the type.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+	User                 string `json:"user,omitempty"`
+	Identifier           string `json:"identifier,omitempty"`
+	PublicKeyData        string `json:"publicKeyData,omitempty"`
+	PublicKeyFingerprint string `json:"publicKeyFingerprint,omitempty"`
+	PrivateKeyData       string `json:"PrivateKeyData,omitempty"`
+}
+
+// Get SSH Key from Bank Vaults
+func KeyGet(organizationID string, ClusterID uint) (sshKey Key, err error) {
```
bonifaido: Please start the comment with `KeyGet` otherwise `go lint` cries.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+	db := model.GetDB()
+	awsProperties := &model.AmazonClusterModel{ClusterModelId: ClusterID}
+	if err := db.First(&awsProperties).Error; err != nil {
+		log.Errorf("Get ssh key failed reason: %s", err.Error())
+		return sshKey, err
+	}
+
+	vaultContent, err := secret.Store.Get(organizationID, awsProperties.SshSecretID)
+	if err != nil {
+		log.Debugf("organizationID: %q, SshSecretID: %q", organizationID, awsProperties.SshSecretID)
```
bonifaido: I think one `Errorf` log with all the details should be sufficient here, I would merge Debugf and Errorf here.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+		log.Errorf("Error during store: %s", err.Error())
+		return "", err
+	}
+
+	log.Info("Store SSH Key stored.")
+	return secretID, nil
+}
+
+// KeyGenerator for Generate new SSH Key
+func KeyGenerator() (Key, error) {
```
bonifaido: I would change this method to return `*Key` and return `nil, err` in case of an error.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+	sshKey.Identifier = vaultContent.Values[secret.Identifier]
+	sshKey.PublicKeyData = vaultContent.Values[secret.PublicKeyData]
+	sshKey.PublicKeyFingerprint = vaultContent.Values[secret.PublicKeyFingerprint]
+	sshKey.PrivateKeyData = vaultContent.Values[secret.PrivateKeyData]
+	log.Debug("Get SSH Key Done.")
+	return sshKey, nil
+}
+
+// KeyAdd for Generate and store SSH key
+func KeyAdd(organizationID uint, clusterID uint) (secretID string, sshKey Key, err error) {
```
bonifaido: I would change this method to return `*Key` and return `secretID, nil, err` in case of an error.

cluster/aws.go
```diff
@@ -185,20 +185,14 @@ func (c *AWSCluster) CreateCluster() error {
...
-	sshKeyPath := viper.GetString("cloud.keypath")
-	//TODO move to the profile section
-	if sshKeyPath != "" {
-		log.Debug("Overwriting default SSH key path to:", sshKeyPath)
-		c.kubicornCluster.SSH.PublicKeyPath = sshKeyPath
-	}
 
-	log.Info("Init cluster")
-	newCluster, err := initapi.InitCluster(c.kubicornCluster)
+	sshSecretID, sshKey, err := pipelineSsh.KeyAdd(c.GetModel().OrganizationId, c.GetModel().ID)
```
bonifaido: Please handle this `error`.

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+
+	sshKey, err = KeyGenerator()
+	if err != nil {
+		log.Errorf("KeyGenerator failed reason: %s", err.Error())
+		return secretID, sshKey, err
+	}
+
+	db := model.GetDB()
+	var clusterName model.ClusterModel
+	db.Model(&model.ClusterModel{ID: clusterID}).First(&clusterName)
```
stoader: Not sure if this the proper way to query cluster record from db.

I would do: 
```
cluster := model.ClusterModel{ID: clusterID}
if err = db.First(&cluster).Error; err != nil {
  log.Errorf("Cluster with id=% not found: %s", clusterId, err.Error())
  return "", nil, err
}
```

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/banzaicloud/pipeline/config"
+	"github.com/banzaicloud/pipeline/model"
+	"github.com/banzaicloud/pipeline/secret"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh"
+	"strings"
+)
+
+var logger *logrus.Logger
+var log *logrus.Entry
```
baluchicken: I think this one is unused

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+func KeyGet(organizationID string, ClusterID uint) (sshKey Key, err error) {
+	log := logger.WithFields(logrus.Fields{"tag": "KeyGet"})
+	log.Info("Get SSH key")
+	if organizationID == "" || ClusterID == 0 {
+		log.Debugf("KeyGet organizationID: %q ClusterID: %q", organizationID, ClusterID)
+		return sshKey, fmt.Errorf("parameter missing")
+	}
+	db := model.GetDB()
+	awsProperties := &model.AmazonClusterModel{ClusterModelId: ClusterID}
+	if err := db.First(&awsProperties).Error; err != nil {
```
baluchicken: No need to use '&'awsProperties because you already declared it as a pointer

baluchicken: Sadly the compiler accepts literally anything  for method First because go lacks generics they had to use interface{} as a parameter type.

ssh/ssh.go
baluchicken: I would create a struct from secretID string, sshKey Key and return that instead of three things

ssh/ssh.go
```diff
@@ -0,0 +1,153 @@
...
+		secret.PublicKeyFingerprint: key.PublicKeyFingerprint,
+		secret.PrivateKeyData:       key.PrivateKeyData,
+	}
+
+	if err := secret.Store.Store(fmt.Sprint(organizationID), secretID, &createSecretRequest); err != nil {
+		log.Errorf("Error during store: %s", err.Error())
+		return "", err
+	}
+
+	log.Info("Store SSH Key stored.")
```
baluchicken: I would say SSH Key stored

PR #538, opened 2018-06-04
Store Kubernetes config in vault

colin014:
Fixes: #527

### DIFF
--- a/api/secrets.go
+++ b/api/secrets.go
@@ -3,7 +3,6 @@ package api
 import (
 	"net/http"
 
-	"encoding/base64"
 	"fmt"
 
 	"github.com/banzaicloud/banzai-types/components"
@@ -11,6 +10,7 @@ import (
 	"github.com/banzaicloud/pipeline/cluster"
 	"github.com/banzaicloud/pipeline/model"
 	"github.com/banzaicloud/pipeline/secret"
+	"github.com/banzaicloud/pipeline/utils"
 	"github.com/gin-gonic/gin"
 	"github.com/go-errors/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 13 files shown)


api/secrets.go
```diff
@@ -290,10 +297,12 @@ func checkClustersBeforeDelete(orgId, secretId string) error {
 	return nil
 }
 
-// encodeStringToBase64 first checks if the string is encoded if yes returns it if no than encodes it.
-func encodeStringToBase64(s string) string {
-	if _, err := base64.StdEncoding.DecodeString(s); err != nil {
-		return base64.StdEncoding.EncodeToString([]byte(s))
+func searchForbiddenTags(orgId, secretId string) error {
```
stoader: add comment for this function

cluster/kubernetes.go
```diff
@@ -202,3 +202,18 @@ func (b *KubeCluster) ValidateCreationFields(r *components.CreateClusterRequest)
 func (b *KubeCluster) GetSecretWithValidation() (*secret.SecretsItemResponse, error) {
 	return b.commonSecret.get(b)
 }
+
+// SaveConfigSecretId saves the config secret id in database
+func (b *KubeCluster) SaveConfigSecretId(configSecretId string) error {
```
stoader: Apparently the `SaveConfigSecretId`, `GetConfigSecretId` is replicated for each kubernetes types. If possible this should be factored or use composition to avoid copy pasting the same code to multiple places

colin014: Unfortunately, this means a big refactor. I did an issue for this: #539

cluster/aws.go
```diff
@@ -1090,7 +1081,7 @@ func (c *AWSCluster) GetClusterDetails() (*components.ClusterDetailsResponse, er
 	log := logger.WithFields(logrus.Fields{"tag": "GetClusterDetails"})
 	log.Info("Start getting cluster details")
 
-	c.GetK8sConfig()
+	c.DownloadK8sConfig()
```
baluchicken: This is wrong here because https://github.com/banzaicloud/pipeline/pull/538/files#diff-8e624ff87723b97e43c3758eff0c292dL828 is factored out so this method does nothing here.

PR #540, opened 2018-06-05
Delete security group for ELB before cluster delete

colin014:


cluster/aws.go
```diff
@@ -1194,3 +1212,145 @@ func (c *AWSCluster) GetConfigSecretId() string {
...
+
+	return output.SecurityGroups, nil
+}
+
+// deleteSecurityGroup deletes a security group by group params
+func (c *AWSCluster) deleteSecurityGroup(svc *ec2.EC2, group *ec2.SecurityGroup) error {
+
+	log.Infof("Delete security group [%s]", *group.GroupId)
+
+	alive := true
```
baluchicken: instead of this for { break}

PR #546, opened 2018-06-05
Gke Cluster destroy timeout

### LINKED ISSUE #507: GKE Cluster termination resource cleanup
Some resource is retained because the cluster termination process is too quick. We need to wait until the Kubernetes destroys all the resources which it had created.

ahma:
Fixes #507


bonifaido:
LGTM

PR #548, opened 2018-06-06
Support Namespaces for Deployments (#541)

### LINKED ISSUE #541: Support Namespaces for Deployments
Should add a new field to like namespace to DeploymentRequest
```
type DeploymentRequest struct {
	deploymentName        string
	deploymentReleaseName string
	reuseValues           bool
        namespace       string
	values                []byte
	kubeConfig            []byte
	organizationName      string
}
```

baluchicken:
Fixes #541

### DIFF
--- a/api/helm.go
+++ b/api/helm.go
@@ -59,7 +59,7 @@ func CreateDeployment(c *gin.Context) {
 		return
 	}
 	release, err := helm.CreateDeployment(parsedRequest.deploymentName,
-		helm.DefaultNamespace,
+		parsedRequest.namespace,
 		parsedRequest.deploymentReleaseName,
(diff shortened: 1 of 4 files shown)


api/helm.go
```diff
@@ -353,6 +354,12 @@ func parseCreateUpdateDeploymentRequest(c *gin.Context) (*parsedDeploymentReques
 	pdr.deploymentReleaseName = deployment.ReleaseName
 	pdr.reuseValues = deployment.ReUseValues
 
+	if deployment.Namespace == "" {
```
stoader: This logic should reside inside `helm.CreateDeployment` method.

Issue #528, opened 2018-06-01
Check cloud credentials


colin014:
Add API endpoint to check:
- [x] AWS credentials
- [x] AKS credentials
- [x] GKE credentials

PR #551, opened 2018-06-06
Verify cloud credentials in secret creation

colin014:
Fixes: #528


utils/constants.go
```diff
@@ -0,0 +1,102 @@
+package utils
```
stoader: I'd create a `constants` package for this within pipeline instead of `utils`

auth/cloud/aws.go
```diff
@@ -0,0 +1,82 @@
...
+var log *logrus.Entry
+
+// Simple init for logging
+func init() {
+	logger = config.Logger()
+	log = logger.WithFields(logrus.Fields{"action": "Cluster"})
+}
+
+// awsVerify for validation AWS credentials
+type awsVerify struct {
```
baluchicken: Why we need this struct? You are using it in `CreateAWSSecret` as a return type but inside `CreateEC2Client` you are not using it, also in case of `CreateAWSCredentials` you are returning `credentials *credentials.Credentials` instead of awsVerify:)

auth/cloud/common.go
```diff
@@ -0,0 +1,23 @@
...
+import "github.com/banzaicloud/banzai-types/constants"
+
+// Verifier validates cloud credentials
+type Verifier interface {
+	VerifySecret() error
+}
+
+// NewVerifier create new instance which implements `Verifier` interface
+func NewVerifier(cloudType string, values map[string]string) Verifier {
+	switch cloudType {
```
baluchicken: Just a minor use `default:` to return nil:) in the end

auth/cloud/gke.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/gin-gonic/gin/json"
+	"golang.org/x/oauth2/google"
+	"golang.org/x/oauth2/jwt"
+	gkeCompute "google.golang.org/api/compute/v1"
+	gke "google.golang.org/api/container/v1"
+	"net/http"
+)
+
+// gkeVerify for validation GKE credentials
+type gkeVerify struct {
```
baluchicken: Same as in case of Amazon:)

auth/cloud/common.go
colin014: thx 👍 done :)

PR #572, opened 2018-06-11
Add missing filter to check load balancer resources

colin014:


cluster/gke.go
```diff
@@ -405,7 +408,32 @@ func checkTargetPools(csv *gkeCompute.Service, project, regionName string) error
...
+					}
+				}
+			}
+		}
+	}
+
+	return filteredPools
+}
+
+func getInstancePrefix(project, zone, clusterName string) string {
```
stoader: Add comment

cluster/gke.go
```diff
@@ -405,7 +408,32 @@ func checkTargetPools(csv *gkeCompute.Service, project, regionName string) error
...
+
+// findTargetPoolsByInstances returns all target pools which created by Kubernetes
+func findTargetPoolsByInstances(pools []*gkeCompute.TargetPool, project, zone, clusterName string) []*gkeCompute.TargetPool {
+
+	instancePrefix := getInstancePrefix(project, zone, clusterName)
+
+	var filteredPools []*gkeCompute.TargetPool
+	for _, p := range pools {
+		if p != nil {
+			if strings.Contains(p.Description, kubernetesIO) {
```
stoader: Isn't there any better way to check if a TargetPool belongs to a gke cluster? Relying on description and prefix is not a robust solution.

cluster/gke.go
```diff
@@ -405,7 +408,33 @@ func checkTargetPools(csv *gkeCompute.Service, project, regionName string) error
...
+			}
+		}
+	}
+
+	return filteredPools
+}
+
+// getInstancePrefix returns URL prefix to the given cluster's virtual machine instance
+func getInstancePrefix(project, zone, clusterName string) string {
+	return fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/gke-%s", project, zone, clusterName)
```
stoader: You could use `targetPrefix` constant here instead of `gke-` literal.

cluster/gke.go
colin014: Unfortunately the target pools only contains these fields:
```
    {
        "creationTimestamp": "2018-06-11T12:55:27.938-07:00",
        "description": "{\"kubernetes.io/service-name\":\"default/pipeline-traefik\"}",
        "healthChecks": [
            "https://www.googleapis.com/compute/v1/projects/colin-pipeline/global/httpHealthChecks/k8s-cc78b40d79f68729-node"
        ],
        "id": "7769113520174755136",
        "instances": [
            "https://www.googleapis.com/compute/v1/projects/colin-pipeline/zones/us-central1-a/instances/gke-gkecluster-colin-369-pool1-a1c22fcc-qvhc"
        ],
        "kind": "compute#targetPool",
        "name": "a4fb01dd26db111e8b50e42010a80024",
        "region": "https://www.googleapis.com/compute/v1/projects/colin-pipeline/regions/us-central1",
        "selfLink": "https://www.googleapis.com/compute/v1/projects/colin-pipeline/regions/us-central1/targetPools/a4fb01dd26db111e8b50e42010a80024",
        "sessionAffinity": "NONE"
    }
```

PR #573, opened 2018-06-11
[WIP] Add applications

### LINKED ISSUE #481: Deploy Application API Endpoint
- [ ] Aplication Create
- [ ] Application List
- [ ] Application Details
- [ ] Tests
- [ ] OpenAPI docs

tarokkk:
- [x] Aplication Create
- [x] Application List
- [x] Application Details
- [x] Application Delete
- [ ] Tests
- [ ] OpenAPI docs
Fixes #481

### DIFF
--- a/api/applications.go
+++ b/api/applications.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"github.com/banzaicloud/banzai-types/components"
 	ctype "github.com/banzaicloud/banzai-types/components/catalog"
 	"github.com/banzaicloud/banzai-types/constants"
@@ -9,9 +10,11 @@ import (
 	"github.com/banzaicloud/pipeline/cluster"
 	"github.com/banzaicloud/pipeline/model"
 	"github.com/gin-gonic/gin"
+	"github.com/jinzhu/gorm"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"net/http"
+	"strconv"
 )
 
 // ApplicationDetailsResponse for API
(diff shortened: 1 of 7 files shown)


api/applications.go
```diff
@@ -37,15 +39,103 @@ type ApplicationListResponse struct {
...
+	if err != nil {
+		log.Errorf("Error getting application: %s", err.Error())
+		c.JSON(http.StatusBadRequest, components.ErrorResponse{
+			Code:    http.StatusBadRequest,
+			Message: "Error getting application",
+			Error:   err.Error(),
+		})
+		return
+	}
+	app.Deployments = deployments
```
stoader: Why do we read deployments first before knowing if the application exists at all?

The flow rather should be, shouldn't it:

db.Model(organization).Related(app)

db.Model(app).Related(&app.Deployments, "Deployments")

api/applications.go
```diff
@@ -37,15 +39,103 @@ type ApplicationListResponse struct {
...
+
+	// Get cluster for application
+	filter := make(map[string]interface{})
+	filter["organization_id"] = organization
+	filter["id"] = app.ClusterID
+	commonCluster, ok := GetCommonClusterFromFilter(c, filter)
```
stoader: Not clear why we need to use `GetCommonClusterFromFilter` here.
There is already a FK constraint on `Application.ClusterId` field, so we should be able to retrieve the Cluster that the app belongs to based on `app.ClusterID` only. Once we have the cluster db record we can create the common cluster from it.



I'd add a foreign key constraint to `Application.ClusterID` and retrieve the related cluster based on the FK form db.  The FK would also

api/applications.go
```diff
@@ -37,15 +39,103 @@ type ApplicationListResponse struct {
...
+			Message: fmt.Sprintf("Invalid id=%q", idParam),
+			Error:   err.Error(),
+		})
+		return
+	}
+	db := model.GetDB()
+	application := &model.Application{
+		ID: uint(id),
+	}
+	var deployments []*model.Deployment
```
stoader: This section is similar to what's being used in other methods. Please factor out into separate function.

model/application.go
```diff
@@ -1,53 +1,86 @@
...
+	ID             uint          `json:"id" gorm:"primary_key"`
+	CreatedAt      time.Time     `json:"createdAt"`
+	UpdatedAt      time.Time     `json:"updatedAt"`
+	DeletedAt      *time.Time    `json:"-" sql:"index"`
+	Name           string        `json:"name"`
+	CatalogName    string        `json:"catalogName"`
+	CatalogVersion string        `json:"catalogVersion"`
+	Description    string        `json:"description"`
+	Icon           string        `json:"icon"`
+	OrganizationId uint          `json:"organizationId"`
```
stoader: I think `OrganizationId` is redundant here as the application belongs to a cluster and the cluster to an organisation.

api/applications.go
```diff
@@ -119,7 +209,7 @@ func CreateApplication(c *gin.Context) {
 			return
 		}
 	}
-	am := model.ApplicationModel{
+	am := model.Application{
```
baluchicken: I would put the Application into Pending State here. As I see it is never on Pending state.

model/application.go
tarokkk: This is not the case. Someone can start a Cluster and Another can deploy Applications. We can rename this to Owner if its more feasable

Issue #553, opened 2018-06-07
Unique and Queryable secret names


tarokkk:
**Is your feature request related to a problem? Please describe.**
We should be able to query secrets by name

**Describe the solution you'd like**
Make secret name unique inside organisations. Provide API endpoint to query by name.

**Additional context**
This feature helps CI/CD systems to specify the needed secret instead of providing all available secrets inside an organisation.
Please add Secret metadata also to the secrets:
- createdAt `time.Time`

PR #575, opened 2018-06-12
Support Vault KV version 2 and unique secret names

bonifaido:
Fixes: #553


secret/store.go
```diff
@@ -79,9 +86,8 @@ func newVaultSecretStore() *secretStore {
 	return &secretStore{client: client, logical: logical}
 }
 
-// GenerateSecretID uuid for new secrets
-func GenerateSecretID() string {
-	return uuid.NewV4().String()
+func generateSecretID(request *CreateSecretRequest) string {
+	return base64.StdEncoding.EncodeToString([]byte(request.Name))
```
baluchicken: Maybe a good Hash function (Sha-1, Sha-256) is a better choice:)

stoader: sha-256 should be good enough

bonifaido: So true!

bonifaido:
I will add also Postman and Swagger updates.

bonifaido:
Ready for review.

Issue #586, opened 2018-06-14
Use common json key for node instance type


colin014:
There are several json keys to node instance type (`nodeInstanceType`, `instance_type` etc.)

Use common key to this: `instanceType`

PR #587, opened 2018-06-14
Use common instanceType json key

colin014:
Fixes: #586


bonifaido:
Was swagger updated previously?

colin014:
I don't know how but it fells behind. I changed the OpenAPI yaml

PR #595, opened 2018-06-16
upgrade / delete / install autoscaler chart when cluster node pools a…

sancyx:
…re updated on Amazon & Azure

### DIFF
--- a/cluster/autoscale.go
+++ b/cluster/autoscale.go
@@ -1,16 +1,26 @@
 package cluster
 
 import (
+	banzaiConst "github.com/banzaicloud/banzai-types/constants"
+	"github.com/banzaicloud/pipeline/auth"
 	"github.com/banzaicloud/pipeline/constants"
 	"github.com/banzaicloud/pipeline/helm"
+	"github.com/ghodss/yaml"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const cloudProviderAzure = "azure"
 const cloudProviderAws = "aws"
-const autoSalerChart = "banzaicloud-stable/cluster-autoscaler"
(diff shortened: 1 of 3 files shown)


cluster/autoscale.go
```diff
@@ -146,3 +156,101 @@ func createAutoscalingForAzure(cluster CommonCluster, groups []nodeGroup) *autos
...
+
+	var nodeGroups []nodeGroup
+
+	switch cluster.GetType() {
+	case banzaiConst.Amazon:
+		nodeGroups = getAmazonNodeGroups(cluster)
+	case banzaiConst.Azure:
+		nodeGroups = getAzureNodeGroups(cluster)
+	default:
+		return nil
```
stoader: I'd error out with a not supported message here.

cluster/autoscale.go
```diff
@@ -146,3 +156,101 @@ func createAutoscalingForAzure(cluster CommonCluster, groups []nodeGroup) *autos
...
+
+func deployAutoscalerChart(cluster CommonCluster, nodeGroups []nodeGroup, kubeConfig []byte, action deploymentAction) error {
+	var values *autoscalingInfo
+	switch cluster.GetType() {
+	case banzaiConst.Amazon:
+		values = createAutoscalingForAmazon(cluster, nodeGroups)
+	case banzaiConst.Azure:
+		values = createAutoscalingForAzure(cluster, nodeGroups)
+	default:
+		return nil
```
stoader: Return with a "not supported" error

cluster/autoscale.go
sancyx: Yes but calling with other providers is not really an error, this like more like we deploy autoscaler only for Amazon & Azure. I could make this check before calling DeployClusterAutoscaler, but then I had to check this all the places this function is invoked.

PR #600, opened 2018-06-18
Application patch

colin014:


cluster/hooks.go
```diff
@@ -56,6 +75,24 @@ func pollingKubernetesConfig(cluster CommonCluster) ([]byte, error) {
 	return kubeConfig, err
 }
 
+// WaitingForTillerComeUp waits until till to come up
+func WaitingForTillerComeUp(kubeConfig []byte) error {
```
stoader: There is already a "wait for tiller" implementation in `install/helm.go:RetryHelmInstall` function. There should be only one "wait for tiller" functionality.

colin014: This function only waits for the tiller, doesn't install helm.

Issue #581, opened 2018-06-13
Extend spotguide.yaml with secrets options


tarokkk:
**Is your feature request related to a problem? Please describe.**
To properly manage applications we should be able to save general secrets (like passwords) to our vault instance. For better user experience we should provide automatically generated default secrets for Spotguides.

**Describe the solution you'd like to see**
Example spotguide section:
```
secrets:
  - name: mysql_credential
     password:
       username: root
       password: [randAlphaNum,randAlpha,randNumeric,randAscii],<length>

  - name: nginx_credential
     htaccess:
       username: root
       password: [randAlphaNum,randAlpha,randNumeric,randAscii],<length>

  - name: mysql_tls
     tls:
       host: ???
       expiration: <days>
```
**Additional context**
This simple snippet is an example how-to define custom secrets. The secret is stored in Vault, tagged with the application id. The secret should be presented on the target Kubernetes cluster. The user after can view/download the relevant part of the secret (like client.crt, client.key).

**Secret types**
Secret types defines the way of the secret Kubernetes secret layout:

- `password`: Kubernetes secret with `username` and `password` field
- `tls`: Kubernetes secret with the following fields:
  - CA.crt
  - CA.key
  - Server.crt
  - Server.key
  - Client.crt
  - Client.key

bonifaido:
Work is being done in https://github.com/banzaicloud/pipeline/tree/secret-tls-support

PR #602, opened 2018-06-18
TLS Secret Type support with automatic cert generation if requested

bonifaido:
This feature is required for the spotguide Secret support described in #581


secret/store.go
```diff
@@ -132,6 +133,22 @@ func (ss *secretStore) Store(organizationID string, value *CreateSecretRequest)
 
 	log.Debugln("Storing secret:", path)
 
+	// If we are not storing a full TLS secret instead of it's a request to generate one
+	if value.Type == constants.TLSSecretType && len(value.Values) <= 2 {
+		validity := value.Values[constants.TLSValidity]
+		if validity == "" {
+			validity = "8760h" // 1 year
```
stoader: Can we make the default value configurable?

bonifaido: Yes, of course, I will update it.

bonifaido:
Fixed now.

PR #631, opened 2018-06-25
Generating OpenAPI Go client

bonifaido:
This PR tries to create a working Go client from the existing Swagger description.
Some workarounds had to be applied to the Swagger YAML because of a current issue in the generator: https://github.com/OpenAPITools/openapi-generator/issues/393
An example had been added to the client/examples directory.
Fixes: #612


Makefile
```diff
@@ -87,9 +87,16 @@ ifndef MISSPELL_CMD
 	go get -u github.com/client9/misspell/cmd/misspell
 endif
 
-clean_vendor:
+clean-vendor:
```
colin014: Please, change here also:
https://github.com/banzaicloud/pipeline/blob/master/scripts/symlink-check.sh#L5

bonifaido: Thanks! Fixed.

client/api_helm.go
```diff
@@ -0,0 +1,757 @@
...
+package client
+
+import (
+	"context"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"fmt"
+	"github.com/antihax/optional"
```
colin014: Please add this package to dependencies and vendor

bonifaido: I have fixed this as well now, and have added the check for other vendoring issues! Thanks for catching this.

PR #717, opened 2018-07-16
Posthook fixes

bonifaido:
- Make StoreKubeConfig posthook function idempotent
- Log posthook states as status message during cluster creation

### DIFF
--- a/cluster/hooks.go
+++ b/cluster/hooks.go
@@ -39,6 +39,7 @@ func RunPostHooks(postHooks []PostFunctioner, createdCluster CommonCluster) erro
 			err = createdCluster.UpdateStatus(pkgCluster.Creating, statusMsg)
 			if err != nil {
 				log.Errorf("Error during posthook status update in db [%s]: %s", postHook, err.Error())
+				return err
 			}
 		}
 	}
--- a/cluster/hooks.go
+++ b/cluster/hooks.go
@@ -23,16 +23,22 @@ import (
 )
 
 //RunPostHooks calls posthook functions with created cluster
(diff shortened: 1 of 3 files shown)


cluster/hooks.go
```diff
@@ -23,17 +23,23 @@ import (
...
-				i.Error(createdCluster, err)
+				log.Errorf("Error during posthook function[%s]: %s", postHook, err.Error())
+				postHook.Error(createdCluster, err)
 				return err
 			}
+
+			statusMsg := fmt.Sprintf("Posthook function finished: %s", postHook)
+			err = createdCluster.UpdateStatus(pkgCluster.Creating, statusMsg)
+			if err != nil {
+				log.Errorf("Error during posthook status update in db [%s]: %s", postHook, err.Error())
```
baluchicken: Shouldn't we return here with the error? Otherwise we will have inconsistent state in the DB

secret/store.go
```diff
@@ -27,6 +27,9 @@ var log *logrus.Logger
 // Store object that wraps up vault logical store
 var Store *secretStore
 
+// ErrSecretNotExists denotes 'Not Found' errors for secrets
+var ErrSecretNotExists = fmt.Errorf("There's no secret with this ID")
```
baluchicken: This can be a const

cluster/hooks.go
bonifaido: Right, I will return the err. Thanks!

secret/store.go
baluchicken: My bad sorry this is a struct:)

bonifaido: 👍

PR #736, opened 2018-07-18
repo:status GitHub scope is redundant

bonifaido:
repo scope already covers it


bonifaido:
Yes you are right, fixed.

Issue #740, opened 2018-07-19
Wrong deployments.version type


petivagyok16:
**Describe the bug**
[PUT] `/api/v1/orgs/{orgId}/clusters/{id}/deployments/{name}`
throws the following error:

```
error:"json: cannot unmarshal number into Go struct field CreateUpdateDeploymentRequest.version of type string"
message:"Error during parsing request!"
```
**Expected behavior**
version is number format everywhere else

PR #741, opened 2018-07-19
CreateUpdateDeployment's version field should be an int32

bonifaido:
Fixes: #740


docs/openapi/pipeline.yaml
```diff
@@ -1549,9 +1549,41 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateDeploymentRequest'
         responses:
           '501':
             description: "Not implemented"
+      responses:
```
colin014: Please fix the yaml, currently it is invalid:
```
Schema error at paths['/api/v1/orgs/{orgId}/clusters/{id}/deployments/{name}']
should NOT have additional properties
additionalProperty: requestBody, responses
Jump to line 1471

Schema error at paths['/api/v1/orgs/{orgId}/clusters/{id}/deployments/{name}'].put
should have required property 'responses'
missingProperty: responses
Jump to line 1525
```

PR #756, opened 2018-07-22
Grafana admin password generator

ahma:
- Grafana admin password generator
- CreateOrUpdate secret function

### DIFF
--- a/cluster/hooks.go
+++ b/cluster/hooks.go
@@ -108,7 +108,7 @@ func InstallMonitoring(input interface{}) error {
 		return errors.Errorf("Wrong parameter type: %T", cluster)
 	}
 
-	const grafanaAdminUsername = "admin"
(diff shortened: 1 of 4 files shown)


cluster/hooks.go
```diff
@@ -106,7 +107,34 @@ func InstallMonitoring(input interface{}) error {
 	if !ok {
 		return errors.Errorf("Wrong parameter type: %T", cluster)
 	}
-	return installDeployment(cluster, helm.DefaultNamespace, pkgHelm.BanzaiRepository+"/pipeline-cluster-monitor", "pipeline-monitoring", nil, "InstallMonitoring")
+
+	const grafanaAdminUsername = "admin"
```
stoader: Shouldn't rather this come from config?

ahma: Now you can configure the grafana admin username.

matyix: Should be externalized to viper/config

PR #764, opened 2018-07-26
External-DNS chart support

ahma:


cluster/hooks.go
```diff
@@ -433,23 +433,42 @@ func RegisterDomainPostHook(input interface{}) error {
...
 		return err
 	}
 
-	log.Info("route53 secret successfully installed into cluster")
+	secretID := secret.GenerateSecretIDFromName(secretSources[0].Name)
 
-	return nil
+	route53Secret, err := secret.Store.Get(orgId, secretID)
+	if err != nil {
+		log.Errorf("Failed to install route53 get secret : %s", err.Error())
```
stoader: Fix the log message.

stoader: You could add a new method `GetByName(orgId, name)` to `Store` that expects a name and internally takes care of generating the secret id and invoking  `Get(orgId, secretID)`

PR #769, opened 2018-07-30
Add TLS related requirements to IntallLogging posthook

baluchicken:


cluster/hooks.go
```diff
@@ -149,7 +150,28 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
...
+			},
+		}
+		_, err := secret.Store.Store(cluster.GetOrganizationId(), req)
+		if err != nil {
+			return errors.Errorf("Failed generate TLS secrets to logging operator")
+		}
+		_, err = InstallOrUpdateSecrets(cluster,
+			&pkgSecret.ListSecretsQuery{
+				Type: secretTypes.TLSSecretType,
+				Tag:  "logging-operator",
```
stoader: `logging-operator` should be passed in via a variable or const to avoid code duplication.

cluster/hooks.go
```diff
@@ -149,7 +150,28 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
 	if err != nil {
 		return err
 	}
-
+	if loggingParam.TLSEnabled {
+		req := &secret.CreateSecretRequest{
+			Name: "tls-for-logging-operator",
+			Type: secretTypes.TLSSecretType,
+			Tags: []string{"logging-operator"},
```
stoader: `logging-operator` should be passed in via a variable or const to avoid code duplication.

cluster/hooks.go
```diff
@@ -149,7 +150,28 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
...
+			Values: map[string]string{
+				secretTypes.TLSHosts: "fluent",
```
stoader: Probably it's better this value to come from config.

PR #771, opened 2018-07-31
Grafana deployment hook refactor

ahma:


cluster/hooks.go
```diff
@@ -135,8 +135,18 @@ func InstallMonitoring(input interface{}) error {
...
-	return installDeployment(cluster, grafanaNamespace, pkgHelm.BanzaiRepository+"/pipeline-cluster-monitor", "pipeline-monitoring", []byte(grafanaValues), "InstallMonitoring")
+	grafanaValues := map[string]map[string]string{
+		"grafana": {
+			"adminUser":     grafanaAdminUsername,
+			"adminPassword": grafanaAdminPass,
+		},
+	}
+	grafanaValuesJson, err := json.Marshal(grafanaValues)
+	if err != nil {
+		log.Errorf("Json Convert Failed : %s", err.Error())
```
baluchicken: This one is a fatal error here, because if this fails we are going to get a cryptic error message inside installDeployment. I think we should return here with an error

PR #779, opened 2018-08-05
Finish uncompleted Route53 tasks on startup in background

stoader:
This implementation solely handles the finishing of incomplete tasks running against Route53 in case of a Pipeline restart.

It's separate work to make already created clusters aware of the completion of unfinished Route53 tasks.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -31,7 +31,7 @@ const (
 
 	// Route53MaintenanceWndMinute configuration key for the maintenance window for Route53.
 	// This is the maintenance window before the next AWS Route53 pricing period starts
-	Route53MaintenanceWndMinute = "route53.mainthelm.banzaiRepositoryURenanceWindowMinute"
+	Route53MaintenanceWndMinute = "route53.maintenanceWindowMinute"
 
 	//PipelineMonitorNamespace pipeline infra namespace key
 	PipelineMonitorNamespace = "infra.namespace"
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -7,10 +7,9 @@ import (
 	"strings"
 	"time"
 
-	"path/filepath"
-
 	"github.com/gin-contrib/cors"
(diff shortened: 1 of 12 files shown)


config/configuration.go
```diff
@@ -29,14 +28,17 @@ const (
 
 	// Route53MaintenanceWndMinute configuration key for the maintenance window for Route53.
 	// This is the maintenance window before the next AWS Route53 pricing period starts
-	Route53MaintenanceWndMinute = "route53.maintenanceWindowMinute"
+	Route53MaintenanceWndMinute = "route53.mainthelm.banzaiRepositoryURenanceWindowMinute"
```
bonifaido: I think this configuration flag has a bit confusing name.

dns/route53/iamUser.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/banzaicloud/pipeline/auth"
+	"github.com/sirupsen/logrus"
+)
+
+// createIAMUser creates a Amazon IAM user with the given name and with no login access to console
+// Returns the created IAM user in case of success
+func (dns *awsRoute53) createIAMUser(userName *string) (*iam.User, error) {
```
bonifaido: The IAM user and AccessKey parts should be later on extracted to a common package because the EKS integration uses the same functionality, I don't think it should be done in this PR.
I will open a separate issue for that.

stoader: I agree.

config/configuration.go
stoader: Fixed

dns/route53/iamUser.go
matyix: #784

PR #798, opened 2018-08-07
Simplify global logger setup

sagikazarmark:
This PR tries to simplify or at least make the global logger setup consistent across the repository.

It adds:

- a log.go file with the same setup to each package where the logger is needed (which could become a standard)
- uses logrus.FieldLogger interface instead of *logrus.Logger (where possible) (so that the parent logger itself can be a *logrus.Entry with some base context)

This move also makes it easier to change how we do logging globally if we decide to, but it's heavily opinionated, so feel free to reject if you think it's too big change.

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"fmt"
+	"math"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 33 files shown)


bonifaido:
I haven't checked, but is the runtime [formatter](https://github.com/banzaicloud/logrus-runtime-formatter) working correctly with the `FieldLogger` logger?

sagikazarmark:
`FieldLogger` is just an interface which defines the common methods of `*Logger` and `*Entry`. Underneath it can either be a Logger or an Entry instance, it doesn't matter.

bonifaido:
👍 Ok, I will check out this branch and test it.

bonifaido:
Can we wrap the logger initialization in a [`Once`](https://golang.org/pkg/sync/#Once) as I don't think it is thread-safe currently (it wasn't done in this change, it is an older issue).

sagikazarmark:
Sure, nice idea!

sagikazarmark:
Done

PR #807, opened 2018-08-09
Introduce distribution field

colin014:


colin014:
@waynz0r You're right, sorry. I rename for the original package name

PR #809, opened 2018-08-09
Introduce GetSecretByVaultID and install logging related helm chart in…

baluchicken:
… posthook

### DIFF
--- a/cluster/secrets.go
+++ b/cluster/secrets.go
@@ -157,3 +157,52 @@ func InstallOrUpdateSecretsByK8SConfig(k8sConfig []byte, orgID uint, query *secr
 
 	return secretSources, nil
 }
+
+// InstallSecretWithVaultID installs a secret which determined by the vaultID to the given namespace
+func InstallSecretWithVaultID(cc CommonCluster, secretID, namespace string) (*secretTypes.K8SSourceMeta, error) {
+	k8sConfig, err := cc.GetK8sConfig()
+	if err != nil {
+		log.Errorf("Error during getting config: %s", err.Error())
+		return nil, err
+	}
+	return InstallSecretWithVaultIDByK8SConfig(k8sConfig, cc.GetOrganizationId(), secretID, namespace)
+}
+
+// InstallSecretWithVaultIDByK8SConfig is the same as InstallSecretWithVaultID but use this if you already have a K8S config at hand.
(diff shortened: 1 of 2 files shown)


cluster/secrets.go
```diff
@@ -157,3 +157,52 @@ func InstallOrUpdateSecretsByK8SConfig(k8sConfig []byte, orgID uint, query *secr
 
 	return secretSources, nil
 }
+
+// InstallSecretWithVaultID installs a secret which determined by the vaultID to the given namespace
+func InstallSecretWithVaultID(cc CommonCluster, secretID, namespace string) (*secretTypes.K8SSourceMeta, error) {
+	k8sConfig, err := cc.GetK8sConfig()
+	if err != nil {
+		log.Errorf("Error during getting config: %s", err.Error())
```
sagikazarmark: If the error is returned, probably the caller should log it.

sagikazarmark: You can wrap the error if you want:

`fmt.Errorf("error during getting config: %s", err.Error())`

(Also, error messages should be lower cased/should not start with capital letters as per common golang standard)

cluster/secrets.go
```diff
@@ -157,3 +157,52 @@ func InstallOrUpdateSecretsByK8SConfig(k8sConfig []byte, orgID uint, query *secr
...
+	}
+	return InstallSecretWithVaultIDByK8SConfig(k8sConfig, cc.GetOrganizationId(), secretID, namespace)
+}
+
+// InstallSecretWithVaultIDByK8SConfig is the same as InstallSecretWithVaultID but use this if you already have a K8S config at hand.
+func InstallSecretWithVaultIDByK8SConfig(k8sConfig []byte, orgID uint, secretID, namespace string) (*secretTypes.K8SSourceMeta, error) {
+
+	clusterClient, err := helm.GetK8sConnection(k8sConfig)
+	if err != nil {
+		log.Errorf("Error during building k8s client: %s", err.Error())
```
sagikazarmark: Same here

cluster/secrets.go
```diff
@@ -157,3 +157,52 @@ func InstallOrUpdateSecretsByK8SConfig(k8sConfig []byte, orgID uint, query *secr
...
+		return nil, err
+	}
+
+	resolvedSecret, err := secret.Store.Get(orgID, secretID)
+	if err != nil {
+		log.Errorf("Error during getting secrets with ID %s: %s", secretID, err.Error())
```
sagikazarmark: Same here

PR #817, opened 2018-08-09
added missing table to the list of tables to be created. Made logged …

lpuskas:
…and creating tables consistent

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,10 +92,17 @@ func main() {
 		&objectstore.ManagedGoogleBucket{},
 		&route53model.Route53Domain{}}
 
-	logger.Infoln("Create table(s):", tables)
+	var tableNames string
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -103,9 +90,19 @@ func main() {
...
-	).Error; err != nil {
+		&route53model.Route53Domain{}}
+
+	var tableNames string
+	for _, table := range tables {
+		tableNames += fmt.Sprintf(" %s", db.NewScope(table).TableName())
+	}
+
+	logger.WithFields(logrus.Fields{
+		"table_names": tableNames,
```
sagikazarmark: `strings.TrimSpace` would remove the leading space in the field

lpuskas: fixed, thx

PR #818, opened 2018-08-09
fixed the dockerfile - added the migrate.go file to the build command

lpuskas:
The build command in the dockerfile failed ...


sagikazarmark:
Sorry, my bad. Do we actually need to specify the files? We want to build the whole package, don't we?

Issue #790, opened 2018-08-07
AKS resource group endpoints


colin014:
- [ ] Add a new API endpoint to listing AKS resource groups
- [ ] Add a new API endpoint to creating a new AKS resource group

PR #819, opened 2018-08-09
Add resourcegroup endpoints

colin014:
Fixes: #790


api/resourcegroup.go
```diff
@@ -0,0 +1,102 @@
...
+	pkgCommon "github.com/banzaicloud/pipeline/pkg/common"
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
+
+// GetResourceGroups lists resource groups by secret
+func GetResourceGroups(c *gin.Context) {
+
+	orgID := auth.GetCurrentOrganization(c.Request).ID
+	secretId := c.Query("secretId")
```
stoader: Shouldn't this be rather a header param then query param? IMHO query params better fit passing in filters.

api/resourcegroup.go
```diff
@@ -0,0 +1,102 @@
...
+		Name: request.Name,
+	})
+
+}
+
+// DeleteResourceGroups deletes resource group by name
+func DeleteResourceGroups(c *gin.Context) {
+
+	orgID := auth.GetCurrentOrganization(c.Request).ID
+	secretId := c.Query("secretId")
```
stoader: Shouldn't this be rather a header param then query param? IMHO query params better fit passing in filters.

api/resourcegroup.go
colin014: sure, fixed

api/resourcegroup.go
colin014: sure, fix this

PR #820, opened 2018-08-09
Revert bucket property renaming, relocate bucket create messages

sagikazarmark:


sagikazarmark:
@colin014 not sure if other things related to the API changed regarding the new property names (docs, postman, etc). Since you did the change in the first place: can you verify that the changes in this PR are correct and enough?

colin014:
The first change i did earlier was a regrettable variable rename which changed all json key also (`amazon` -> `ec2` so on). This not touched the Postman collection and OpenAPI with client. So thanks for the catch @sagikazarmark  👍

sagikazarmark:
Cool, thanks for clarifying.

PR #829, opened 2018-08-10
Rename accs to acsk

bonifaido:


model/cluster.go
```diff
@@ -78,7 +78,7 @@ type ACCSNodePoolModel struct {
 }
 
 // AlibabaClusterModel describes the Alibaba Cloud CS cluster model
```
colin014: Wrong comment

bonifaido:
Thanks @colin014, fixed! :)

PR #855, opened 2018-08-13
Move oracle object storage bucket management to provider package

waynz0r:
Refactor oracle related db table migrations


pkg/providers/oracle/migrate_tables.go
```diff
@@ -0,0 +1,37 @@
+package oracle
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/banzaicloud/pipeline/pkg/providers/oracle/model"
+	objectstoreModel "github.com/banzaicloud/pipeline/pkg/providers/oracle/model/objectstore"
```
sagikazarmark: `objectstore` looks enough here

pkg/providers/oracle/model/objectstore/objectstore.go
```diff
@@ -1,25 +1,27 @@
...
 
 import (
-	pipelineAuth "github.com/banzaicloud/pipeline/auth"
+	"github.com/banzaicloud/pipeline/auth"
 )
 
 // TableName constants
 const (
-	ManagedOracleBucketTableName = "managed_oracle_buckets"
+	BucketsTableName = "oracle_buckets"
```
sagikazarmark: Not sure if this is necessary...the table name is not used at any other place

pkg/providers/oracle/model/objectstore/objectstore.go
```diff
@@ -1,25 +1,27 @@
...
-// ManagedOracleBucket is the schema for the DB
-type ManagedOracleBucket struct {
-	ID            uint                      `gorm:"primary_key"`
-	Organization  pipelineAuth.Organization `gorm:"foreignkey:OrgID"`
-	OrgID         uint                      `gorm:"index;not null"`
-	CompartmentID string
-	Name          string `gorm:"unique_index:bucketNameLocation"`
-	Location      string `gorm:"unique_index:bucketNameLocation"`
+// ObjectStoreBucket is the schema for the DB
+type ObjectStoreBucket struct {
```
sagikazarmark: Either follow the `ObjectStoreModel` or rename all models. I'm in favor of the first option.

sagikazarmark: I would probably move this model to the `github.com/banzaicloud/pipeline/pkg/providers/oracle` package as I did with other models.

pkg/providers/oracle/objectstore/objectstore.go
```diff
@@ -0,0 +1,347 @@
+package objectstore
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/banzaicloud/pipeline/auth"
+	"github.com/banzaicloud/pipeline/pkg/objectstore"
+	model "github.com/banzaicloud/pipeline/pkg/providers/oracle/model/objectstore"
```
sagikazarmark: Not sure if this alias is necessary, once files are moved to the proper packages.

pkg/providers/oracle/objectstore/objectstore.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// WithRegion updates the region.
+func (o *ObjectStore) WithRegion(region string) error {
+	o.location = region
+	return nil
+}
+
+// MarkManagedBucket marks buckets exists in the database to 'managed'
+func (o *ObjectStore) MarkManagedBuckets(buckets []*objectstore.BucketInfo, compartmentID string) error {
```
sagikazarmark: Why is this method exported?

pkg/providers/oracle/objectstore/objectstore.go
```diff
@@ -0,0 +1,347 @@
...
+		if mBuckets[key] == key {
+			bucketInfo.Managed = true
+		}
+	}
+
+	return nil
+}
+
+// GetBucketsFromDB gives back object store buckets from DB
+func (o *ObjectStore) GetBucketsFromDB() ([]model.ObjectStoreBucket, error) {
```
sagikazarmark: Why is this method exported?

pkg/providers/oracle/objectstore/objectstore.go
```diff
@@ -0,0 +1,347 @@
...
+		logger.Errorf("Retrieving managed buckets failed: %s", err.Error())
+		return nil, err
+	}
+
+	return buckets, nil
+}
+
+// GetBucketFromDB looks up the managed bucket record in the database based on the specified searchCriteria
+// If no db record is found than returns with bucketNotFoundError
+func (o *ObjectStore) GetBucketFromDB(searchCriteria *model.ObjectStoreBucket, managedBucket *model.ObjectStoreBucket) error {
```
sagikazarmark: Why is this method exported?

sagikazarmark:
I had a few comments, mostly about being consistent with other providers, but other than that: cool 👍

pkg/providers/oracle/migrate_tables.go
waynz0r: I used `objectstoreModel` just to be clear on which package is that. In the current structure the `objectstore` model should be in the `oracle/model` package anyway, but putting it there would cause cyclic imports right now.

pkg/providers/oracle/model/objectstore/objectstore.go
waynz0r: Even so, I think it is useful to have the used constants on top, but it should be unexported.

pkg/providers/oracle/model/objectstore/objectstore.go
waynz0r: Currently the model is in a separate package, the `ObjectStoreModel` name seemed redundant.

Also `ObjectStoreModel` isn't as specific as it should be. We store bucket info there, so it should be `ObjectStoreBucket` or `ObjectStoreBucketModel` if it is not in a separate package.

pkg/providers/oracle/objectstore/objectstore.go
waynz0r: In the current structure the `objectstore` model should be in the `oracle/model` package anyway, but putting it there would cause cyclic imports right now.

pkg/providers/oracle/objectstore/objectstore.go
waynz0r: You're right, it should be unexported.

pkg/providers/oracle/objectstore/objectstore.go
waynz0r: You're right, it should be unexported.

pkg/providers/oracle/objectstore/objectstore.go
waynz0r: You're right, it should be unexported.

PR #856, opened 2018-08-13
Use url if it possible

ahma:


api/custom.go
```diff
@@ -163,15 +163,16 @@ func getLoadBalancersWithIngressPaths(serviceList *v1.ServiceList, ingressList *
 
 	for _, service := range serviceList.Items {
 		var endpointURLs []*pkgHelm.EndPointURLs
-		log.Debugf("Service: %#v", service.Status)
+		log.Debugf("Service Name: %s in Namespace: %s", service.Name, service.Namespace)
```
sagikazarmark: Use `log.Fields`here

PR #859, opened 2018-08-14
EKS config as yaml instead of json

colin014:


colin014:
The yaml config format works fine: 
```
$ kubectl get nodes
NAME                                           STATUS    ROLES     AGE       VERSION
ip-192-168-102-43.us-west-2.compute.internal   Ready     <none>    1h        v1.10.3
```

PR #860, opened 2018-08-14
deploy metrics-server on Amazon & Azure

sancyx:


cluster/hooks.go
```diff
@@ -414,7 +414,25 @@ func InstallHorizontalPodAutoscalerPostHook(input interface{}) error {
...
+	var valuesOverride []byte
+  // install metricsServer for Amazon & Azure
+	switch cluster.GetCloud() {
+	case pkgCluster.Amazon, pkgCluster.Azure:
+		values := &map[string]map[string]string{
+			"metricsServer": {
+				"enabled": "true",
+			},
+		}
+		yamlValues, err := yaml.Marshal(*values)
```
sagikazarmark: Dereferencing is not required (once the pointer above is removed)

cluster/hooks.go
```diff
@@ -414,7 +414,25 @@ func InstallHorizontalPodAutoscalerPostHook(input interface{}) error {
...
+		if err != nil {
+			log.Errorf("Error during values marshal: %s", err.Error())
```
sagikazarmark: Whoever handles the error should also log it, this way the error is logged twice.

PR #869, opened 2018-08-15
Remove GetModel from CommonCluster

sagikazarmark:
This PR stops leaking the internal model to the outside world.


cluster/aks.go
```diff
@@ -793,3 +798,23 @@ func DeleteResourceGroup(orgId uint, secretId, rgName string) error {
 
 	return azureClient.DeleteResourceGroup(client, rgName)
 }
+
+// GetAKSNodePools returns AKS node pools from a common cluster.
+func GetAKSNodePools(cluster CommonCluster) ([]*model.AKSNodePoolModel, error) {
+	akscluster, ok := cluster.(*AKSCluster)
+	if !ok {
+		return nil, errors.New("invalid cluster instance")
```
colin014: This error is used different places, i think this should be moved to a common place and using as a variable.

sagikazarmark: Good point, although that way we lose the attached stack trace, but I hope this will only be an intermediate solution, until we restructure all clusters.

Issue #788, opened 2018-08-07
Delete GKE resources


colin014:
Currently we only waiting for resources(firewall rules, target pools and forwarding rules) to be deleted.

If we run out some quota the Kubernetes service will not created well. In this case we need to delete these resources.

```
INFO[2018-08-07T10:15:02+02:00] get firewall[k8s-2a7a7a03bf254ac7-node-http-hc] in project[banzai-platform-203712]  function=isFirewallDeleted
WARN[2018-08-07T10:15:02+02:00] firewall[k8s-2a7a7a03bf254ac7-node-http-hc] is still alive  function=checkFirewalls
INFO[2018-08-07T10:15:07+02:00] get firewall[k8s-2a7a7a03bf254ac7-node-http-hc] in project[banzai-platform-203712]  function=isFirewallDeleted
WARN[2018-08-07T10:15:07+02:00] firewall[k8s-2a7a7a03bf254ac7-node-http-hc] is still alive  function=checkFirewalls
INFO[2018-08-07T10:15:12+02:00] get firewall[k8s-2a7a7a03bf254ac7-node-http-hc] in project[banzai-platform-203712]  function=isFirewallDeleted
WARN[2018-08-07T10:15:13+02:00] firewall[k8s-2a7a7a03bf254ac7-node-http-hc] is still alive  function=checkFirewalls
INFO[2018-08-07T10:15:18+02:00] get firewall[k8s-2a7a7a03bf254ac7-node-http-hc] in project[banzai-platform-203712]  function=isFirewallDeleted
WARN[2018-08-07T10:15:18+02:00] firewall[k8s-2a7a7a03bf254ac7-node-http-hc] is still alive  function=checkFirewalls
```

PR #875, opened 2018-08-15
Add force delete in case of GKE resources

colin014:
Fixes: #788


cluster/gke_operations.go
```diff
@@ -0,0 +1,5 @@
+package cluster
+
+type OperationInfoer interface {
```
sagikazarmark: This should be unexported, operation constructors should return struct pointer types instead.

PR #895, opened 2018-08-21
Add config option to set CORS allowed origins as regular expression

tothandras:


config/configuration.go
```diff
@@ -137,19 +138,30 @@ func init() {
...
 	config := cors.DefaultConfig()
-	cors.DefaultConfig()
 	config.AllowAllOrigins = viper.GetBool("cors.AllowAllOrigins")
 	if !config.AllowAllOrigins {
+		allowOriginsRegexp := viper.GetString("cors.AllowOriginsRegexp")
+		if allowOriginsRegexp != "" {
+			originsRegexp, err := regexp.Compile(allowOriginsRegexp)
+			if err == nil {
+				config.AllowOriginFunc = func(origin string) bool {
+					return originsRegexp.Match([]byte(origin))
```
orymate: I would highlight somewhere in the code that the regexp is matched as substring pattern, which is not that usual for similar config options.

PR #916, opened 2018-08-23
Add default image to OKE cluster in case of create/update

colin014:


pkg/providers/oracle/cluster/defaults.go
```diff
@@ -0,0 +1,5 @@
+package cluster
+
+const (
+	defaultImage = "Oracle-Linux-7.4" // todo needs to be refactor out in change where defaults came from config
```
waynz0r: Oracle-Linux-7.5 seems available, although I did not try it.

PR #934, opened 2018-08-27
Fix cors.AllowOrigin* config parsing

orymate:
* Fix regression from #895: AllowOrigins option became unreachable
* Use whole string matching instead of substring matching for AllowOriginsRegexp as it is not intuitive and exposes a potential security risk

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -155,7 +155,7 @@ func GetCORS() cors.Config {
 	if !config.AllowAllOrigins {
 		allowOriginsRegexp := viper.GetString("cors.AllowOriginsRegexp")
 		if allowOriginsRegexp != "" {
-			originsRegexp, err := regexp.Compile(allowOriginsRegexp)
+			originsRegexp, err := regexp.Compile(fmt.Sprintf("^(%s)$", allowOriginsRegexp))
 			if err == nil {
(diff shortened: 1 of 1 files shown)


config/configuration.go
```diff
@@ -155,15 +155,15 @@ func GetCORS() cors.Config {
...
 		allowOriginsRegexp := viper.GetString("cors.AllowOriginsRegexp")
 		if allowOriginsRegexp != "" {
-			originsRegexp, err := regexp.Compile(allowOriginsRegexp)
+			originsRegexp, err := regexp.Compile(fmt.Sprintf("^(%s)$", allowOriginsRegexp))
 			if err == nil {
 				config.AllowOriginFunc = func(origin string) bool {
 					return originsRegexp.Match([]byte(origin))
 				}
 			}
+		} else if len(viper.GetStringSlice("cors.AllowOrigins")) > 0 {
```
sagikazarmark: This would look nicer as
```
} else if allowOrigins := viper.GetStringSlice("cors.AllowOrigins"); len(allowOrigins) > 0 {
    config.AllowOrigins = allowOrigins
```

orymate: agree, fixed

PR #939, opened 2018-08-28
Spotguides first API draft

bonifaido:


api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+
+// GetSpotguide get detailed information about a spotguide
+func GetSpotguide(c *gin.Context) {
+	spotguideName := strings.TrimPrefix(c.Param("name"), "/")
+	spotguideDetails, err := spotguide.GetSpotguide(spotguideName)
+	if err != nil {
+		if gorm.IsRecordNotFoundError(err) {
+			c.JSON(http.StatusNotFound, pkgCommon.ErrorResponse{
+				Code:    http.StatusNotFound,
+				Message: "Spotguide not found",
```
baluchicken: Just a minor thing these error messages should start with lowercase letter

api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+			})
+			return
+		}
+		log.Errorf("Error getting spotguide details: %s", err.Error())
+		c.JSON(http.StatusInternalServerError, pkgCommon.ErrorResponse{
+			Code:    http.StatusInternalServerError,
+			Message: "Error getting spotguide details",
```
baluchicken: same here

api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// GetSpotguides lists all available spotguides
+func GetSpotguides(c *gin.Context) {
+	spotguides, err := spotguide.GetSpotguides()
+	if err != nil {
+		log.Errorf("Error listing spotguides: %s", err.Error())
+		c.JSON(http.StatusInternalServerError, pkgCommon.ErrorResponse{
+			Code:    http.StatusInternalServerError,
+			Message: "Error listing spotguides",
```
baluchicken: same here

api/spotguide.go
baluchicken: See @sagikazarmark earlier refactor about #940

api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+	c.Status(http.StatusAccepted)
+}
+
+// LaunchSpotguide creates a spotguide workflow, all secrets, repositories.
+func LaunchSpotguide(c *gin.Context) {
+	var launchRequest spotguide.LaunchRequest
+	if err := c.BindJSON(&launchRequest); err != nil {
+		c.JSON(http.StatusBadRequest, pkgCommon.ErrorResponse{
+			Code:    http.StatusBadRequest,
+			Message: "Error parsing request",
```
baluchicken: same here

api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	org := auth.GetCurrentOrganization(c.Request)
+	user := auth.GetCurrentUser(c.Request)
+
+	err := spotguide.LaunchSpotguide(&launchRequest, org.ID, user.ID)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, pkgCommon.ErrorResponse{
+			Code:    http.StatusInternalServerError,
+			Message: "Error creating spotguide",
```
baluchicken: same here

auth/user.go
```diff
@@ -174,6 +204,14 @@ func (bus BanzaiUserStorer) Save(schema *auth.Schema, context *auth.Context) (us
 
 	AddDefaultRoleForUser(currentUser.ID)
 
+	// Save the Github token to Vault
+	token := bauth.NewToken(GithubTokenID, "Github access token")
+	token.Value = githubExtraInfo.Token
+	err = TokenStore.Store(fmt.Sprint(currentUser.ID), token)
+	if err != nil {
+		return "", "", fmt.Errorf("Failed to store Github access token: %s", err.Error())
```
baluchicken: same here

spotguide/spotguide.go
```diff
@@ -0,0 +1,323 @@
...
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"time"
+
+	"github.com/banzaicloud/pipeline/auth"
+	"github.com/banzaicloud/pipeline/config"
+	"github.com/banzaicloud/pipeline/secret"
+	yaml "github.com/ghodss/yaml"
```
baluchicken: Don't need to call this as yaml because the package already called that way

spotguide/spotguide.go
```diff
@@ -0,0 +1,323 @@
...
+	err := db.Where("name = ?", name).Find(&spotguide).Error
+	return &spotguide, err
+}
+
+// curl -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" -v http://localhost:9090/api/v1/orgs/1/spotguides -d '{"repoName":"spotguide-test", "repoOrganization":"banzaicloud"}'
+func LaunchSpotguide(request *LaunchRequest, orgID, userID uint) error {
+
+	err := createGithubRepo(request, userID)
+	if err != nil {
+		log.Errorln("Failed to create GitHub repository", err.Error())
```
baluchicken: consider using the pkg/ errors package here. we can call errors.cause before sending back bad requests to the customer.

api/spotguide.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/banzaicloud/pipeline/auth"
+	pkgCommon "github.com/banzaicloud/pipeline/pkg/common"
+	"github.com/banzaicloud/pipeline/spotguide"
+	"github.com/gin-gonic/gin"
+	"github.com/jinzhu/gorm"
+)
+
+// GetSpotguide get detailed information about a spotguide
+func GetSpotguide(c *gin.Context) {
+	spotguideName := strings.TrimPrefix(c.Param("name"), "/")
```
stoader: Why is `name` path param optional? If no spot guide `name` is passed in than the consequent `spotguide.GetSpotguide(spotguideName)` call will return not found as spotguideName is an empty string.

main.go
```diff
@@ -171,22 +182,17 @@ func main() {
...
-			orgs.GET("/:orgid/applications/:id", api.ApplicationDetails)
-			orgs.DELETE("/:orgid/applications/:id", api.DeleteApplications)
-
-			orgs.GET("/:orgid/catalogs", api.GetCatalogs)
-			orgs.PUT("/:orgid/catalogs/update", api.UpdateCatalogs)
-			orgs.GET("/:orgid/catalogs/:name", api.CatalogDetails)
+			orgs.GET("/:orgid/spotguides", api.GetSpotguides)
+			orgs.PUT("/:orgid/spotguides", api.SyncSpotguides)
+			orgs.POST("/:orgid/spotguides", api.LaunchSpotguide)
+			orgs.GET("/:orgid/spotguides/*name", api.GetSpotguide)
```
stoader: Why is `name` parameter optional?

I'd add `HEAD` as well to for verifying if a spotguide exists or not.

api/spotguide.go
bonifaido: `name` is not optional, I just had to use `*` because otherwise, I can't capture the spotguide's name which can contain `/` (because it is the name of the github repo it is originating from). But the `if == ""` check is missing you are right!

PR #942, opened 2018-08-28
OKE: enable RBAC and set proper admin rights after cluster creation

waynz0r:


cluster/oke.go
```diff
@@ -79,7 +85,17 @@ func (o *OKECluster) CreateCluster() error {
...
 
-	return cm.ManageOKECluster(&o.modelCluster.OKE)
+	err = cm.ManageOKECluster(&o.modelCluster.OKE)
+	if err != nil {
+		return errors.Wrap(err, "error creating cluster")
+	}
+
+	err = o.setClusterAdminRights("cluster-creator-admin-right")
+	if err != nil {
+		return errors.Wrap(err, "error get/create clusterrolebinding")
```
sagikazarmark: Errors in the `setClusterAdminRights` already Wrap original errors. `errors.Wrap` does two things: attaches stack trace and an additional message to the error. By calling Wrap again, the original stack trace gets lost.

For subsequent error wrapping use `errors.WithMessage` instead which attaches an additional message only.

cluster/oke.go
```diff
@@ -536,5 +552,56 @@ func (o *OKECluster) ListNodeNames() (nodeNames pkgCommon.NodeNames, err error)
 
 // RbacEnabled returns true if rbac enabled on the cluster
 func (o *OKECluster) RbacEnabled() bool {
-	return o.modelCluster.RbacEnabled
+	return true
+}
+
+// setClusterAdminRights creates a cluster role binding which gives admin
+// rights to the user ocid specified in the secret used to create the cluster
+func (o *OKECluster) setClusterAdminRights(name string) (err error) {
```
sagikazarmark: No need to define a name `err` in advance as you don't use it anywhere (the first `err :=` redeclares the variable and you return a value everywhere)

cluster/oke.go
```diff
@@ -79,7 +85,17 @@ func (o *OKECluster) CreateCluster() error {
 		return err
 	}
 
-	return cm.ManageOKECluster(&o.modelCluster.OKE)
+	err = cm.ManageOKECluster(&o.modelCluster.OKE)
+	if err != nil {
+		return errors.Wrap(err, "error creating cluster")
```
sagikazarmark: The error wrapping comment might be relevant here, too.

cluster/oke.go
```diff
@@ -536,5 +552,56 @@ func (o *OKECluster) ListNodeNames() (nodeNames pkgCommon.NodeNames, err error)
...
+		return errors.Wrap(err, "error getting k8s client")
+	}
+
+	secret, err := o.GetSecretWithValidation()
+	if err != nil {
+		return errors.Wrap(err, "error getting secret")
+	}
+
+	if secret.Values["user_ocid"] == "" {
+		return errors.Errorf("empty user OCID")
```
sagikazarmark: You can use errors.New since there is no formatting.

cluster/oke.go
```diff
@@ -536,5 +552,56 @@ func (o *OKECluster) ListNodeNames() (nodeNames pkgCommon.NodeNames, err error)
...
+	if err != nil {
+		return errors.Wrap(err, "error getting k8s client")
+	}
+
+	secret, err := o.GetSecretWithValidation()
+	if err != nil {
+		return errors.Wrap(err, "error getting secret")
+	}
+
+	if secret.Values["user_ocid"] == "" {
```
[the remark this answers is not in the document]
waynz0r: Good catch, we actually have one already, coming right up! Thanks @stoader!

PR #960, opened 2018-08-31
Introduce setup privileges posthook function

colin014:
To avoid this error:
```
time="2018-08-31T06:48:12Z" level=error msg="failed to create cluster: error get/create clusterrolebinding: error getting k8s config: There's no secret with this ID" cluster=banzaicloudbdy4 correlation-id=4204c97e-8f48-4e1d-9349-f1fc9897875e function=func1 organization=2 user=14
```

- Add `NeedAdminRights` and `GetKubernetesUserID` to common interface.
- Set admin rights in posthook function


cluster/hooks.go
```diff
@@ -443,6 +445,21 @@ func InstallKubernetesDashboardPostHook(input interface{}) error {
 			return err
 		}
 
+		// set admin rights (if needed)
```
waynz0r: We should probably move this to its own hook and call it right after `StoreKubeConfig`. The admin rights could also be necessary elsewhere.

cluster/hooks.go
```diff
@@ -443,6 +445,21 @@ func InstallKubernetesDashboardPostHook(input interface{}) error {
...
+		if cluster.NeedAdminRights() {
+
+			userId, err := cluster.GetKubernetesUserID()
+			if err != nil {
+				return err
+			}
+
+			if err := setAdminRights(client, userId); err != nil {
```
stoader: This is confusing as `cluster.GetKubernetesUserID()` returns a user id while `setAdminRights` expects `userName`.

PR #967, opened 2018-08-31
Add logging switch based on type GCS

tarokkk:


cluster/hooks.go
```diff
@@ -246,19 +240,48 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
...
-		},
-	}
-	marshaledValues, err := yaml.Marshal(loggingValues)
+
+	// Determine the type of output plugin
+	logSecret, err := secret.Store.Get(cluster.GetOrganizationId(), loggingParam.SecretId)
 	if err != nil {
 		return err
 	}
+	switch logSecret.Type {
```
stoader: Is this secret needed by logging operator at startup? if so shouldn't this be installed into the cluster before the operator deployed?

cluster/hooks.go
```diff
@@ -246,19 +240,48 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
...
+	case pkgCluster.Amazon:
+		installedSecretValues, err := InstallSecretWithVaultID(cluster, loggingParam.SecretId, loggingParam.GenTLSForLogging.Namespace)
```
stoader: The `InstallSecretWithVaultID` is not idempotent thus the hook may fail in case of a re-run. Please consider using/addin a version of `InstallSecretWithVaultID` (see as an example `InstallOrUpdateSecrets`)

cluster/hooks.go
tarokkk: This needed for the output plugin installed separatly

cluster/hooks.go
```diff
@@ -246,19 +240,48 @@ func InstallLogging(input interface{}, param pkgCluster.PostHookParam) error {
...
+		if err != nil {
+			return err
+		}
+		loggingValues := map[string]interface{}{
+			"bucketName": loggingParam.BucketName,
```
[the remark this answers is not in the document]
tarokkk: Agreed after GCS and Azure have a final values layout I will transform these maps to structs

PR #970, opened 2018-09-02
Upgrade helm (2.10) and update dependencies

tarokkk:


Gopkg.lock
```diff
@@ -1680,13 +2015,15 @@
...
     "util/retry",
-    "util/workqueue"
+    "util/workqueue",
   ]
+  pruneopts = "NUT"
   revision = "23781f4d6632d88e869066eaebb743857aa1ef9b"
   version = "v7.0.0"
 
 [[projects]]
   branch = "release-2.9"
```
sagikazarmark: It looks like helm version didn't change.

tarokkk: It's my bad I mixed up something at the dep update gona fix

baluchicken:
If you want to upgrade the tiller version also you need to change this and the config example
https://github.com/banzaicloud/pipeline/blob/ad9a26aa8bd38f8de2de926a08d8d0fe9e0004fb/config/configuration.go#L70 to use 2.10

PR #974, opened 2018-09-03
LaunchSpotguide should use an existing Drone token

bonifaido:


auth/user.go
```diff
@@ -135,8 +134,20 @@ func GetCurrentUserFromDB(req *http.Request) (*User, error) {
 	return nil, errors.New("error fetching user from db")
 }
 
+func newDroneClient(apiToken string) drone.Client {
+	droneURL := viper.GetString("drone.url")
```
baluchicken: please also add this parameter to the config.toml.example

bonifaido: Added, thanks!

PR #983, opened 2018-09-04
Update CI and local dev env to Vault 0.11.0

bonifaido:


bonifaido:
Please wait, I will upgrade the client lib to 0.11 as well.

PR #984, opened 2018-09-04
Add location info to bucket list endpoint in case of Amazon

### LINKED ISSUE #973: S3 bucket region filter
We need to filter this call GET path=“/pipeline/api/v1/orgs/2/buckets?cloudType=amazon&location=us-west-2 result with the region param.

baluchicken:
Fixes #973 
It will return the list of the buckets enhanced by the location.
```
    {   "name": "your-mysql-backup",
        "managed": false,
        "location": "us-west-1"
    },
    {
        "name": "zeppelin-sample-notebooks",
        "managed": false,
        "location": "us-west-2"
    },
```

### DIFF
--- a/pkg/providers/amazon/objectstore/objectstore.go
+++ b/pkg/providers/amazon/objectstore/objectstore.go
@@ -79,16 +79,25 @@ func (s *objectStore) ListBuckets() ([]string, error) {
 	return bucketList, nil
 }
 
-// CheckBucket checks the status of the given bucket.
-func (s *objectStore) CheckBucket(bucketName string) error {
-	// Check if the bucket's region matches the current region
-	actualRegion, err := s3manager.GetBucketRegionWithClient(context.Background(), s.client, bucketName)
+// GetRegion gets the region of the given bucket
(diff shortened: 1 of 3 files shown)


pkg/providers/amazon/objectstore/objectstore.go
```diff
@@ -79,16 +79,24 @@ func (s *objectStore) ListBuckets() ([]string, error) {
 	return bucketList, nil
 }
 
-// CheckBucket checks the status of the given bucket.
-func (s *objectStore) CheckBucket(bucketName string) error {
-	// Check if the bucket's region matches the current region
-	actualRegion, err := s3manager.GetBucketRegionWithClient(context.Background(), s.client, bucketName)
+func (s *objectStore) GetRegion(bucketName string) (string, error) {
```
stoader: Add comment

baluchicken:
Just to be clear about #973 It requests a filter for ListBucket but we are using the  location query parameter for Amazon client connection reasons. I know it is a little bit odd, but all the other providers ListBucket calls are returning all Buckets so I don't want to change that behaviour instead I enhanced the list with location/region details, like we do in case of Oracle.

PR #988, opened 2018-09-05
Add test for getregion and create amazon specific interface instead o…

baluchicken:
…f extending the common one

### DIFF
--- a/internal/providers/amazon/objectstore.go
+++ b/internal/providers/amazon/objectstore.go
@@ -23,9 +23,14 @@ type bucketNotFoundError struct{}
 func (bucketNotFoundError) Error() string  { return "bucket not found" }
 func (bucketNotFoundError) NotFound() bool { return true }
 
+type amazonObjectStore interface {
+	commonObjectstore.ObjectStore
+	GetRegion(bucket string) (string, error)
+}
+
 // objectStore stores all required parameters for bucket creation.
 type objectStore struct {
-	objectStore commonObjectstore.ObjectStore
+	objectStore amazonObjectStore
 
 	region string
 	secret *secret.SecretItemResponse
--- a/pkg/objectstore/objectstore.go
+++ b/pkg/objectstore/objectstore.go
@@ -13,7 +13,4 @@ type ObjectStore interface {
 
 	// DeleteBucket removes a bucket from the object store.
 	DeleteBucket(string) error
-
-	// GetRegion returns the region for a given bucket
-	GetRegion(string) (string, error)
 }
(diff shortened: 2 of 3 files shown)


internal/providers/amazon/objectstore.go
```diff
@@ -23,9 +23,14 @@ type bucketNotFoundError struct{}
 func (bucketNotFoundError) Error() string  { return "bucket not found" }
 func (bucketNotFoundError) NotFound() bool { return true }
 
+type amazonObjectStore interface {
+	commonObjectstore.ObjectStore
+	GetRegion(bucket string) (string, error)
```
stoader: This should be private if not used by other packages, only inside the amazon objectstore.

bonifaido: I think no interface should have private methods.

pkg/objectstore/objectstore.go
```diff
@@ -13,7 +13,9 @@ type ObjectStore interface {
 
 	// DeleteBucket removes a bucket from the object store.
 	DeleteBucket(string) error
+}
 
-	// GetRegion returns the region for a given bucket
+type ObjectStoreWithRegion interface {
```
sagikazarmark: Since (for now) we only need this in case of amazon, this contract could be owned by our amazon object store wrapper. (lacking a better name: the one in the internal directory 😄 ) It could also be unexported, since it's only used there.

pkg/providers/amazon/objectstore/objectstore_test.go
```diff
@@ -80,6 +80,40 @@ func TestObjectStore_CreateBucket(t *testing.T) {
...
+
+	input := &s3.CreateBucketInput{
+		Bucket: aws.String(bucketName),
+	}
+
+	_, err := client.CreateBucket(input)
+	if err != nil {
+		t.Fatal("could not create test bucket: ", err.Error())
+	}
+	region, err := s.GetRegion(bucketName)
```
sagikazarmark: New line before (readability)

pkg/providers/amazon/objectstore/objectstore_test.go
```diff
@@ -80,6 +80,40 @@ func TestObjectStore_CreateBucket(t *testing.T) {
...
+	if err != nil {
+		t.Error("testing bucket region failed: ", err.Error())
+	} else {
+		if strings.TrimSpace(os.Getenv("AWS_REGION")) != region {
+			t.Error("test bucket region does not match")
+		}
+	}
+	del := &s3.DeleteBucketInput{
```
sagikazarmark: New line before (readability)

PR #994, opened 2018-09-06
Count available nodes when checking if the cluster is ready in case o…

baluchicken:
…f alibaba

### DIFF
--- a/cluster/acsk.go
+++ b/cluster/acsk.go
@@ -317,7 +317,7 @@ func (c *ACSKCluster) CreateCluster() error {
 
 	// wait for cluster created
 	log.Info("Waiting for cluster...")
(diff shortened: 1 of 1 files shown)


cluster/acsk.go
```diff
@@ -454,7 +454,13 @@ func waitForClusterState(client *cs.Client, clusterID string) (*alibabaDescribeC
 
 		switch r.State {
 		case AlibabaClusterStateRunning:
-			return r, nil
+			for _, v := range r.Outputs {
+				if v.OutputKey == "NodeInstanceIDs" {
+					if len(v.OutputValue.([]interface{})) == numNodes {
```
colin014: I think we should check tha cast result:)

colin014:
LGTM, thx @baluchicken

PR #995, opened 2018-09-06
EKS: update AMI's & nodegroup cf template

sancyx:


pkg/cluster/eks/defaults.go
```diff
@@ -7,8 +7,8 @@ const (
 
 // DefaultImages in each supported location in EC2 (from https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html)
 var DefaultImages = map[string]string{
-	UsEast1: "ami-0fef2bff3c2e2da93",
-	UsWest2: "ami-0ea01e1d1dea65b5c",
+	UsEast1: "ami-0b2ae3c6bda8b5c06",
+	UsWest2: "ami-08cab282f9979fc7a",
```
colin014: Please update these in postman collection and OpenAPI

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -356,42 +388,19 @@ Resources:
...
-          ]
-
+          !Sub |
+            #!/bin/bash
+            set -o xtrace
+            /etc/eks/bootstrap.sh ${ClusterName} ${BootstrapArguments}
+            /opt/aws/bin/cfn-signal --exit-code $? \
+                     --stack  ${AWS::StackName} \
+                     --resource NodeGroup  \
+                     --region ${AWS::Region}
```
stoader: For adding node pool label to node we need to pass `--node-labels pipeline-nodepool-name=", { Ref: NodeGroupName }`

pkg/cluster/eks/defaults.go
```diff
@@ -7,8 +7,8 @@ const (
 
 // DefaultImages in each supported location in EC2 (from https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html)
 var DefaultImages = map[string]string{
-	UsEast1: "ami-0fef2bff3c2e2da93",
-	UsWest2: "ami-0ea01e1d1dea65b5c",
+	UsEast1: "ami-0b2ae3c6bda8b5c06",
```
stoader: Please add `ami` for GPU (p2, p3) instances.

pkg/cluster/eks/defaults.go
sancyx: I'm not sure we had to update example Ami id's in OpenAPi & Postman, since it's only an example and there are different AMI for different zones, plus it will be more complicated in case of GPU instrances. I think we should make a description elsewhere about correct AMI's.

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: set as Bootstrap param to nodegroup template

pkg/cluster/eks/defaults.go
sancyx: I think we should handle GPU related images in a separate PR.

pkg/cluster/eks/defaults.go
stoader: Then remove the example and update the field's description such as point the user to the documentation where the AMIs are listed: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html

templates/eks/amazon-eks-nodepool-cf.yaml
stoader: Apparently `MaxPodsPerNode` section was removed in latest CF template so we need to remove it our side as well.

pkg/cluster/eks/defaults.go
matyix: Since we update this anyway, can we add the `new` [Irelend])https://aws.amazon.com/about-aws/whats-new/2018/09/amazon-eks-available-in-ireland/) region as well.

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: removed from template it's not used anyway in latest template

pkg/cluster/eks/defaults.go
sancyx: eu-west-1 image has been added

PR #1005, opened 2018-09-07
Create/UpgradeDeployment: fix invalid json response

orymate:
If the cluster was not found, both parseCreateUpdateDeploymentRequest()
and Create/UpgradeDeployment() wrote a full json response to the context
resulting in two json top-level dicts sent.

needs testing!

### DIFF
--- a/api/helm.go
+++ b/api/helm.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/banzaicloud/pipeline/auth"
+	"github.com/banzaicloud/pipeline/cluster"
 	"github.com/banzaicloud/pipeline/helm"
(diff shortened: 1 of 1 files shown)


orymate:
Tested.
```
Before:
% curl -H "Authorization: Bearer $LOCAL_TOKEN"   -H "Content-Type: application/json" -XPOST http://localhost:9090/api/v1/orgs/2/clusters/0/deployments  [d37ca242*]
{"code":404,"message":"cluster not found","error":"could not get cluster from database: cluster not found"}{"code":400,"message":"Error during parsing request!","error":"Get cluster failed!"}%

After:
{"code":404,"message":"cluster not found","error":"could not get cluster from database: cluster not found"}%
```

PR #1010, opened 2018-09-07
Client header

sagikazarmark:


client/api/openapi.yaml
```diff
@@ -6423,6 +6423,25 @@ components:
             $ref: '#/components/schemas/NodePoolsAmazon'
           type: object
       type: object
+    CreateEKSProperties_eks_master:
```
stoader: Please remove the `master` section from create EKS cluster openapi spec as for EKS this is not valid. This seems to be an oversight from a previous commit.

sagikazarmark: Will be done in a separate PR

Issue #1012, opened 2018-09-10
Add new secret endpoint


colin014:
Add `/clusters/:clusterID/secrets`. Similar like `/endpoints`, list secrets which belongs to cluster (and deployment)

PR #1016, opened 2018-09-10
List secrets which belongs to cluster

colin014:
Fixes: #1012


colin014:
Good catch @bonifaido thx, i fixed that

PR #1018, opened 2018-09-10
Fix EKS create cluster properties documentation

sagikazarmark:


docs/openapi/pipeline.yaml
```diff
@@ -3566,18 +3566,12 @@ components:
         eks:
           type: object
           required:
+            - version
```
colin014: `version` is not required in case of EKS

sagikazarmark:
Fixed

PR #1023, opened 2018-09-11
Delete all Kubernetes resources

### LINKED ISSUE #808: K8S proper resource cleanup at cluster delete
Before deleting the cluster currently, we do a `helm delete --all` like call to delete all deployments. This is deleting all backing resources the helm deployments have created, however if someone have created a resource with plain kubectl calls, for example, `kubectl expose service --type LoadBalancer ...` that LoadBalancer will never get cleaned up. So before deleting a cluster we should call a `kubectl delete services --all --all-namespaces` like call, probably for volumes as well (?).

tarokkk:
Fixes #808

### DIFF
--- a/cluster/manager_delete.go
+++ b/cluster/manager_delete.go
@@ -32,7 +32,7 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
 	return nil
 }
 
-func deleteAllResource(kubeConfig []byte) error {
+func deleteAllResource(kubeConfig []byte, logger *logrus.Entry) error {
 	client, err := helm.GetK8sConnection(kubeConfig)
 	if err != nil {
 		return err
@@ -53,9 +53,12 @@ func deleteAllResource(kubeConfig []byte) error {
 	options := metav1.NewDeleteOptions(0)
 
 	for _, service := range services {
-		err := service.Delete("", options)
-		if err != nil {
-			return err
+		for i := 0; i < 3; i++ {
+			err := service.Delete("", options)
+			if err != nil {
+				logger.Debug("deleting resources %T attempt %d/%d failed", service, i, 3)
+				time.Sleep(1)
+			}
 		}
 	}
 	return nil
@@ -93,24 +96,20 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 	if !(force && c == nil) {
 		// delete deployments
 		for i := 0; i < 3; i++ {
-			err = deleteAllResource(c)
+			err = helm.DeleteAllDeployment(c)
 			// TODO we could check to the Authorization IAM error explicit
 			if err != nil {
-				logger.Errorf("deleting resources failed: %s", err.Error())
+				logger.Errorf("deleting deployments attempt %d/%d failed: %s", i, 3, err.Error())
 				time.Sleep(1)
 			} else {
 				break
 			}
 		}
-		for i := 0; i < 3; i++ {
-			err = helm.DeleteAllDeployment(c)
-			// TODO we could check to the Authorization IAM error explicit
-			if err != nil {
-				logger.Errorf("deleting deployments failed: %s", err.Error())
-				time.Sleep(1)
-			} else {
-				break
-			}
+		err = deleteAllResource(c, logger)
+		if err != nil && !force {
+			return emperror.Wrap(err, "deleting resources failed")
+		} else {
+			logger.Errorf("deleting resources failed: %s", err.Error())
 		}
 	} else {
 		logger.Info("skipping deployment deletion without kubeconfig")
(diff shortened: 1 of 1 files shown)


cluster/manager_delete.go
```diff
@@ -61,7 +91,7 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 
 	if !(force && c == nil) {
 		// delete deployments
-		err = helm.DeleteAllDeployment(c)
+		err = deleteAllResource(c, logger)
```
stoader: Add retry logic to handle intermittent issues like unauthorized in case of EKS

cluster/manager_delete.go
```diff
@@ -30,6 +31,35 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
...
 
+func deleteAllResource(kubeConfig []byte, logger *logrus.Entry) error {
+	client, err := helm.GetK8sConnection(kubeConfig)
+	if err != nil {
+		return err
+	}
+	// Delete all resources, log errors but ignore them
+	err = client.CoreV1().Services("").Delete("", metav1.NewDeleteOptions(0))
+	if err != nil {
+		logger.Info(err.Error())
```
sagikazarmark: Returning a wrapped error would be better here, so that the caller can log the error if it wants to.

cluster/manager_delete.go
sagikazarmark: Retry logic should probably part of the function itself.

Also, with a bit of code compressing you can make retrying easier:

```go
type resourceDeleter interface {
	Delete(name string, options metav1.DeleteOptions)
}

services := []resourceDeleter{
	client.CoreV1().Services(""),
	client.AppsV1().Deployments(""),
	// ...
}

options := metav1.NewDeleteOptions(0)

for service := range services {
	err := service.Delete(options)
	if err != nil {
		// retry?
	}
}
```

Not sure if this is ultimately better.

cluster/manager_delete.go
```diff
@@ -61,7 +91,7 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 
 	if !(force && c == nil) {
 		// delete deployments
-		err = helm.DeleteAllDeployment(c)
```
sagikazarmark: Why is helm deployment deletion completely removed? It could contain created resources which are not being deleted by the new code. (Note that deployment != deployment in this case, also the correct terminology is release in case of helm....which (I know) is a confusing name)

tarokkk: True that I'll put it back

cluster/manager_delete.go
```diff
@@ -30,6 +32,38 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
...
+	if err != nil {
+		return err
+	}
+
+	type resourceDeleter interface {
+		Delete(name string, options *metav1.DeleteOptions) error
+	}
+
+	services := []resourceDeleter{
+		client.CoreV1().Services(""),
```
baluchicken: These should be metav1.NamespaceAll

cluster/manager_delete.go
```diff
@@ -30,6 +32,38 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
...
+		client.AppsV1().DaemonSets(""),
+		client.AppsV1().StatefulSets(""),
+		client.AppsV1().ReplicaSets(""),
+	}
+
+	options := metav1.NewDeleteOptions(0)
+
+	for _, service := range services {
+		for i := 0; i < 3; i++ {
+			err := service.Delete("", options)
```
baluchicken: same here

cluster/manager_delete.go
```diff
@@ -30,6 +32,38 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
...
+			if err != nil {
+				logger.Debug("deleting resources %T attempt %d/%d failed", service, i, 3)
```
baluchicken: It is Debugf

cluster/manager_delete.go
```diff
@@ -30,6 +32,38 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
...
+		client.AppsV1().Deployments(""),
+		client.AppsV1().DaemonSets(""),
+		client.AppsV1().StatefulSets(""),
+		client.AppsV1().ReplicaSets(""),
+	}
+
+	options := metav1.NewDeleteOptions(0)
+
+	for _, service := range services {
+		for i := 0; i < 3; i++ {
```
baluchicken: Use const instead of 3 because 3 is used in a debug message

cluster/manager_delete.go
tarokkk: This is not a namespace it is the resource name

cluster/manager_delete.go
tarokkk: Fixed

cluster/manager_delete.go
baluchicken: Sorry you are right:)

PR #1027, opened 2018-09-12
HPA: install Metrics Server on Oracle & Alibaba, fix annotation setup

sancyx:


api/hpa.go
```diff
@@ -24,6 +24,12 @@ func PutHpaResource(c *gin.Context) {
 	kubeConfig, ok := GetK8sConfig(c)
 	if !ok {
 		log.Errorf("could not get the k8s config")
+		httpStatusCode := http.StatusInternalServerError
+		c.JSON(httpStatusCode, pkgCommmon.ErrorResponse{
```
colin014: Please dont send JSON response here, the `GetK8sConfig` already did it.

api/hpa.go
```diff
@@ -73,6 +90,12 @@ func DeleteHpaResource(c *gin.Context) {
...
```
colin014: Same as above

api/hpa.go
```diff
@@ -107,6 +130,12 @@ func GetHpaResource(c *gin.Context) {
...
```
colin014: Same as above

api/hpa.go
```diff
@@ -253,8 +282,10 @@ func DeleteDeploymentAutoscalingInfo(kubeConfig []byte, scaleTarget string) erro
 	// get doesn't work with v12.NamespaceAll only if you specify the namespace exactly
 	// deployment, err := client.AppsV1().Deployments(v12.NamespaceAll).Get(request.Name, v12.GetOptions{})
 	scaleTargetFound := false
-
-	deploymentList, err := client.AppsV1().Deployments(v12.NamespaceAll).List(v12.ListOptions{})
+	listOptions := v12.ListOptions{
+		FieldSelector: fmt.Sprintf("metadata.name=%v", scaleTarget),
+	}
+	deploymentList, err := client.AppsV1().Deployments(v12.NamespaceAll).List(listOptions)
```
colin014: Please handle the error here

api/hpa.go
```diff
@@ -265,7 +296,7 @@ func DeleteDeploymentAutoscalingInfo(kubeConfig []byte, scaleTarget string) erro
 	}
 
 	// find statefulset & update hpa annotations
-	statefulSetList, err := client.AppsV1().StatefulSets(v12.NamespaceAll).List(v12.ListOptions{})
+	statefulSetList, err := client.AppsV1().StatefulSets(v12.NamespaceAll).List(listOptions)
```
colin014: Please handle the error

api/hpa.go
```diff
@@ -291,25 +322,29 @@ func SetDeploymentAutoscalingInfo(kubeConfig []byte, request hpa.DeploymentScali
...
 	// get doesn't work with v12.NamespaceAll only if you specify the namespace exactly
-	// deployment, err := client.AppsV1().Deployments(v12.NamespaceAll).Get(request.Name, v12.GetOptions{})
+	//deployment, err := client.AppsV1().Deployments(v12.NamespaceAll).Get(request.Name, v12.GetOptions{})
 	scaleTargetFound := false
-
-	deploymentList, err := client.AppsV1().Deployments(v12.NamespaceAll).List(v12.ListOptions{})
+	listOptions := v12.ListOptions{
+		FieldSelector: fmt.Sprintf("metadata.name=%v", request.ScaleTarget),
+	}
+	deploymentList, err := client.AppsV1().Deployments(v12.NamespaceAll).List(listOptions)
```
colin014: Please handle the error

api/hpa.go
```diff
@@ -291,25 +322,29 @@ func SetDeploymentAutoscalingInfo(kubeConfig []byte, request hpa.DeploymentScali
...
 			log.Infof("set annotations on deployment: %v", dep.Name)
+			dep.Annotations = removeHpaAnnotations(dep.Annotations)
 			setupHpaAnnotations(request, dep.Annotations)
 			_, err = client.AppsV1().Deployments(dep.Namespace).Update(&dep)
 		}
 	}
 
 	// find statefulset & update hpa annotations
-	statefulSetList, err := client.AppsV1().StatefulSets(v12.NamespaceAll).List(v12.ListOptions{})
+	statefulSetList, err := client.AppsV1().StatefulSets(v12.NamespaceAll).List(listOptions)
```
colin014: Please handle the error

api/hpa.go
```diff
@@ -39,7 +45,18 @@ func PutHpaResource(c *gin.Context) {
 		})
 		return
 	}
-	log.Info("Parse deployment succeeded")
+
+	err = scalingRequest.Validate()
+	if err != nil {
+		err := errors.Wrap(err, "Error parsing request:")
```
baluchicken: @sagikazarmark can you please take a look? I don't think we need to wrap the error here.

api/hpa.go
baluchicken: I just wondering how does this went through our lint mechanism, it suppose to fail with something "error does not handled properly" cc @sagikazarmark

pkg/hpa/hpa.go
```diff
@@ -45,6 +53,74 @@ type DeploymentScalingRequest struct {
...
+func (rm ResourceMetric) validateResourceMetric() error {
+	switch rm.TargetAverageValueType {
+	case PercentageValueType:
+		int64Value, err := strconv.ParseInt(rm.TargetAverageValue, 10, 32)
+		if err != nil {
+			return errors.New("invalid percentage value specified")
+		}
+		targetValue := int32(int64Value)
+		if targetValue <= 0 || targetValue > 100 {
+			return errors.New(fmt.Sprintf("invalid percentage value specified: %v (Percentage value shoud be between [1,99]", targetValue))
```
baluchicken: This can be fmt.Errorf, that way you can eliminate the usage of fmt.Sprintf

pkg/hpa/hpa.go
```diff
@@ -45,6 +53,74 @@ type DeploymentScalingRequest struct {
...
+
+		}
+	case QuantityValueType:
+		_, err := resource.ParseQuantity(rm.TargetAverageValue)
+		if err != nil {
+			return errors.New(fmt.Sprintf("invalid resource metric value: %v (%v)", rm.TargetAverageValue, err.Error()))
```
baluchicken: same here

pkg/hpa/hpa.go
```diff
@@ -45,6 +53,74 @@ type DeploymentScalingRequest struct {
...
+		}
+	}
+	return nil
+}
+
+func (rm CustomMetric) validateCustomMetric() error {
+	if rm.Type != "pod" {
+		return errors.New(fmt.Sprintf("invalid custom metric type specified: %v", rm.Type))
```
baluchicken: same here

pkg/hpa/hpa.go
```diff
@@ -45,6 +53,74 @@ type DeploymentScalingRequest struct {
...
+	}
+
+	_, err := resource.ParseQuantity(rm.TargetAverageValue)
+	if err != nil {
+		return errors.New(fmt.Sprintf("invalid custom metric value: %v (%v)", rm.TargetAverageValue, err.Error()))
```
baluchicken: same here

api/hpa.go
sagikazarmark: I suppose it did, because the err variable is actually used later. That's why `ineffassign` lint didn't kick in. The `errcheck` lint is explicitly disabled. 😄

api/hpa.go
sagikazarmark: The scaling request internally uses the errors package from the standard library, so errors emitted by it do not have any stack trace, which might make sense in case of validation errors actually, so if that method only emits validation errors (and nothing like connection or timeout errors) then I think it's OK to use the errors package from the stdlib as validation errors are rather business errors than application errors.

For the same reason I agree with @baluchicken that annotating the error with a stack trace here doesn't add too much, since the error gets "handled" just a few lines later. If you want to add an extra message to the error, you can do so using `errors.WithMessage` (also, the extra colon at the end of the message is not necessary, error messages in Go are started with lower case by convention).

api/hpa.go
```diff
@@ -39,7 +45,18 @@ func PutHpaResource(c *gin.Context) {
...
+		log.Error(err.Error())
```
sagikazarmark: I know it's not how usually done in pipeline, but this is not actually an error from the application point of view, since it's only a validation failure of the data provided by the user. As such, it should logged as info or debug information, normally validation failure doesn't concern us (imagine an attack where the attacker deliberately sends invalid payloads just to flood us with error messages.)

api/hpa.go
sancyx: fixed

api/hpa.go
```diff
@@ -68,20 +78,25 @@ func DeleteHpaResource(c *gin.Context) {
 		})
 		return
 	}
-	log.Infof("getting hpa details for scaleTarget: [%s]", scaleTarget)
+	log.Debugf("getting hpa details for scaleTarget: [%s]", scaleTarget)
 
 	kubeConfig, ok := GetK8sConfig(c)
 	if !ok {
 		log.Errorf("could not get the k8s config")
+		c.JSON(http.StatusInternalServerError, pkgCommmon.ErrorResponse{
```
colin014: Please dont send JSON response here, `GetK8sConfig` already did it

PR #1028, opened 2018-09-12
fix logging at delete

sancyx:


cluster/manager_delete.go
```diff
@@ -108,10 +108,14 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 			}
 		}
 		err = deleteAllResource(c, logger)
-		if err != nil && !force {
-			return emperror.Wrap(err, "deleting resources failed")
+		if err != nil {
```
baluchicken: It seems the same behaviour to me.

PR #1040, opened 2018-09-17
Update EKS AMIs to latest

stoader:


sagikazarmark:
Merged too soon 😄 

I would prefer if this kind of default would come from code (eg. in a before create hook if the value is not provided). Setting database level defaults requires a schema change and this kind of default value belongs to the application logic anyway.

PR #1047, opened 2018-09-18
Create and update flows should survive restarts

sagikazarmark:


waynz0r:
We should put this into the interface too.

sagikazarmark:
Already done :)

sagikazarmark:
As per our latest discussion, this won't be part of the release.

Issue #1000, opened 2018-09-07
Define API endpoints for Security scans


tarokkk:
To use and visualise  the security sate of a cluster we need Pipeline endpoints to connect the resource
- [x] Define API endpoint to represent Cluster, Helm deployment, Pod with security scan statuses
- [x] Define API for policy management

tarokkk:
OpenAPI definitioön in #1057

PR #1057, opened 2018-09-21
Security scan OpenApi3 definition

tarokkk:


tarokkk:
Fixes #1000

docs/openapi/pipeline.yaml
```diff
@@ -6402,3 +6733,303 @@ components:
...
+        resource:
+          type: string
+          example: 'banzaicloud'
+        image:
+          example: ['nginx:latest', 'busybox:latest']
+          type: array
+          items:
+            type: string
+        result:
+          example: 'Warning level CVE found'
```
pbalogh-sa: Result field is a list, because listed images can have different scan result

docs/openapi/pipeline.yaml
```diff
@@ -1480,6 +1719,98 @@ paths:
...
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/ReleaseWhiteList'
+      post:
+        security:
+          - bearerAuth: []
+        tags:
+          - whitelist
+        summary: List Whitelisted deployments
```
pbalogh-sa: Add whitelist instead of listing

tarokkk: Fixed ty

PR #1076, opened 2018-09-24
Add drain mode middleware

sagikazarmark:
This is a temporary solution for pipeline deployments maintenance mode. It will probably become obsolete when cluster management workflows become durable.

Usage:

Enable drain mode: `curl -X POST http://localhost:9090/-/drain`

Disable drain mode: `curl -X DELETE http://localhost:9090/-/drain`

Note: The middleware disables operations that might change state (POST, PUT, PATCH, DELETE), but it might also break features which use these HTTP methods, but don't change state (eg. authentication).

### DIFF
--- a/internal/platform/gin/drain.go
+++ b/internal/platform/gin/drain.go
@@ -16,6 +16,7 @@ package gin
 
 import (
 	"net/http"
+	"strings"
 	"sync"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


sagikazarmark:
@stoader done

internal/platform/gin/drain.go
```diff
@@ -0,0 +1,109 @@
...
+
+func isWriteOperation(c *gin.Context) bool {
+	return c.Request.Method == http.MethodPost ||
+		c.Request.Method == http.MethodPut ||
+		c.Request.Method == http.MethodPatch ||
+		c.Request.Method == http.MethodDelete
+}
+
+func isException(c *gin.Context) bool {
+	if c.Request.URL.Path == "/api/v1/tokens" {
```
bonifaido: I'm not sure if the basePath should be added here or not. I guess yes.

PR #1090, opened 2018-09-26
Heptio/ARK support

waynz0r:


api/ark/backupservice/backupservice.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/banzaicloud/pipeline/api/ark/common"
+	"github.com/banzaicloud/pipeline/config"
+)
+
+// AddRoutes adds ARK backups related API routes
+func AddRoutes(group *gin.RouterGroup) {
+
+	group.Use(common.ARKMiddleware(config.DB(), common.Log))
+	group.GET("/status", Status)
```
colin014: I think this should be a HEAD endpoint

bonifaido:
Depends on #1084

api/ark/backups/get.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/banzaicloud/pipeline/api/ark/common"
+	"github.com/banzaicloud/pipeline/internal/platform/gin/correlationid"
+)
+
+// Get gets an ARK backup
+func Get(c *gin.Context) {
+	backupName := c.Param("name")
+
+	logger := correlationid.Logger(common.Log, c).WithField("backup", backupName)
```
colin014: Are you sure we want to gettig backups by `name` instead of `id`?

config/configuration.go
```diff
@@ -131,6 +131,13 @@ func init() {
 	viper.SetDefault(OKEWaitAttemptsForNodepoolActive, 60)
 	viper.SetDefault(OKESleepSecondsForNodepoolActive, 30)
 
+	viper.SetDefault("ark.name", "ark")
+	viper.SetDefault("ark.namespace", "pipeline-infra")
+	viper.SetDefault("ark.chart", "stable/ark")
+	viper.SetDefault("ark.image", "waynz0r/ark")
```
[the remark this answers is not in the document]
waynz0r: Sure. We need to modify the stable helm chart anyway to support mixed deployment, where the cluster is at one provider and the bucket is at another.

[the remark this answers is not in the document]
waynz0r: In the current state of the PR only GCP is supported, so mixed deployment is not possible anyways.

internal/ark/api/backups.go
```diff
@@ -0,0 +1,170 @@
...
+// DeleteBackupResponse describes a delete backup response
+type DeleteBackupResponse struct {
+	Name   string `json:"name"`
+	Status int    `json:"status"`
+}
+
+// CreateBackupRequest descibes a create backup request
+type CreateBackupRequest struct {
+	Name    string          `json:"name" binding:"required"`
+	TTL     metav1.Duration `json:"ttl" binding:"required"`
```
colin014: Are you sure this is required? I think this should be optional

api/ark/backups/get.go
waynz0r: You're right. I'll change it to `id` for the backup/bucket/restore related requests.

I would probably stick with the `name` for the schedules. They are only exists within the cluster and the name is unique there.

internal/ark/api/backups.go
waynz0r: I think it is an important field and although there is a default for it, it maybe better if the caller must set it explicitly.

api/ark/backups/get.go
colin014: OK, thx

waynz0r:
FYI, i'll push some commits later today.

api/ark/common/middleware.go
```diff
@@ -0,0 +1,107 @@
...
+
+		svc := ark.NewARKService(org, cluster, db, logger)
+		c.Request = setVariableToContext(c.Request, arkServiceName, svc)
+
+		c.Next()
+	}
+}
+
+// GetCurrentClusterModel return the current cluster model
+func GetCurrentClusterModel(req *http.Request) *model.ClusterModel {
```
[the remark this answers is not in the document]
waynz0r: This is dead code, I'll remove it.

api/ark/common/middleware.go
```diff
@@ -0,0 +1,107 @@
...
+func ARKMiddleware(db *gorm.DB, logger logrus.FieldLogger) gin.HandlerFunc {
+
+	return func(c *gin.Context) {
+		logger = logger.WithField("middleware", "ARK")
+
+		clusters := intCluster.NewClusters(db)
+		clusterID, ok := ginutils.UintParam(c, "id")
+		if !ok {
+			logger.Error("invalid ID parameter")
+			c.Abort()
```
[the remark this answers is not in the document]
waynz0r: UintParams sends a reply, but it doesn't abort the further processing the request.

config/config.toml.example
```diff
@@ -107,3 +107,11 @@ resourceDeleteSleepSeconds = 5
 [oke]
 waitAttemptsForNodepoolActive = 60
 sleepSecondsForNodepoolActive = 30
+
+[ark]
+name = "ark"
+namespace = "pipeline-infra"
+chart = "stable/ark"
+image = "waynz0r/ark"
```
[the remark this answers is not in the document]
waynz0r: 👍

PR #1098, opened 2018-09-26
[WIP] Add WhiteList api

### LINKED ISSUE #1095: Implement Pipeline Whitelist API

tarokkk:
fixes #1095

### DIFF
--- a/cluster/whitelist.go
+++ b/cluster/whitelist.go
@@ -1,0 +1,58 @@
+package cluster
+
+import (
+	"github.com/banzaicloud/pipeline/helm"
+	"github.com/banzaicloud/pipeline/pkg/security"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+)
+
+const GroupName = "security.banzaicloud.com"
+const GroupVersion = "v1alpha1"
+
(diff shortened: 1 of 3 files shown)


cluster/whitelist.go
```diff
@@ -0,0 +1,58 @@
...
+
+func GetWhitelists(cc CommonCluster) (*security.WhiteList, error) {
+	result := &security.WhiteList{}
+	// Get kubernetes configuration
+	kubeConfig, err := cc.GetK8sConfig()
+	if err != nil {
+		return nil, err
+	}
+	// Get customresource Client
+	clientCfg, err := helm.GetK8sClientConfig(kubeConfig)
```
stoader: The code for instantiating a client for manipulating CRD could be used in other places as well, thus I think it worth moving it to a common place.

cluster/whitelist.go
```diff
@@ -0,0 +1,58 @@
...
+	clientCfg.APIPath = "/apis"
+	clientCfg.NegotiatedSerializer = serializer.DirectCodecFactory{CodecFactory: scheme.Codecs}
+	clientCfg.UserAgent = rest.DefaultKubernetesUserAgent()
+
+	client, err := rest.RESTClientFor(clientCfg)
+	if err != nil {
+		return nil, err
+	}
+	opts := &metav1.ListOptions{}
+	err = client.Get().Namespace("").Resource("whitelists").VersionedParams(opts, scheme.ParameterCodec).Do().Into(result)
```
stoader: We should add support for namespaces

pbalogh-sa: The CRD is clusterwide at this moment.

stoader: Does the "" stand for all namespaces or  for default?

baluchicken: Yes it stands for all namespace, please use metav1.AllNamespaces instead.

PR #1102, opened 2018-09-27
Small HPA fixes

colin014:
Fix invalid json response and send not found message if needed

### DIFF
--- a/api/hpa.go
+++ b/api/hpa.go
@@ -67,7 +67,7 @@ func PutHpaResource(c *gin.Context) {
 
 	err = setDeploymentAutoscalingInfo(kubeConfig, *scalingRequest)
 	if err != nil {
-		err := errors.Wrap(err, "Error during request processing:")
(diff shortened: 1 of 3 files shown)


api/hpa.go
```diff
@@ -118,11 +118,6 @@ func DeleteHpaResource(c *gin.Context) {
 func GetHpaResource(c *gin.Context) {
 	scaleTarget, ok := ginutils.RequiredQuery(c, "scaleTarget")
 	if !ok {
-		c.JSON(http.StatusBadRequest, pkgCommmon.ErrorResponse{
```
bonifaido: Why this block got deleted?

bonifaido: Sorry I missed that, probably it should have a name like `RequiredQueryOrError` just to make it more understandable.

colin014: You're right. I will rename this function

bonifaido: OrAbort is probably more Gin-ish! :)

PR #1113, opened 2018-10-01
fix bucket api

orymate:


orymate:
could someone approve the new commit ec2ad37 as well? thx

PR #1127, opened 2018-10-01
openapi: fix field name in bucket creation response

orymate:


api/bucket.go
```diff
@@ -194,7 +194,7 @@ func CreateBucket(c *gin.Context) {
 	logger.Debug("bucket creation started")
 
 	c.JSON(http.StatusAccepted, CreateBucketResponse{
-		Name: createBucketRequest.Name,
```
bonifaido: Shouldn't be CreateBucketRequest.Name also renamed?

orymate: Then using Name universally would be better. But this breaks the api, and it has already a few uses

Issue #1130, opened 2018-10-02
spotguide: organization level separation of spotguide catalog


bonifaido:
This is separated on the API level already but not on the DB level.

PR #1138, opened 2018-10-03
Add org level spotguide catalog support

bonifaido:
This PR also introduces the idea of general eventing as well to the project to support more loose coupling between modules and support extraction of services to separate processes in the future.

Fixes: #1130

### DIFF
--- a/auth/events.go
+++ b/auth/events.go
@@ -1,0 +1,36 @@
+// Copyright © 2018 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+// OrganizationRegisteredTopic is the name of the topic where organization registration events are published.
+const OrganizationRegisteredTopic = "organization_registered"
+
+// authEvents is responsible for dispatching domain events throughout the system.
+// It does not express any infrastructural detail (like pubsub).
+type authEvents interface {
+	OrganizationRegistered(organizationID uint)
+}
+
+type eventBus interface {
+	Publish(topic string, args ...interface{})
+}
+
+type ebAuthEvents struct {
+	eb eventBus
+}
+
+func (e ebAuthEvents) OrganizationRegistered(organizationID uint) {
+	e.eb.Publish(OrganizationRegisteredTopic, organizationID)
+}
--- a/auth/user.go
+++ b/auth/user.go
@@ -188,6 +188,7 @@ type BanzaiUserStorer struct {
 	auth.UserStorer
 	signingKeyBase32 string // Drone uses base32 Hash
 	droneDB          *gorm.DB
+	events           authEvents
 }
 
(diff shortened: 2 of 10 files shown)


auth/events.go
```diff
@@ -0,0 +1,35 @@
...
+
+package auth
+
+import "github.com/banzaicloud/pipeline/internal/events"
+
+const OrganizationRegisteredTopic = "organization_registered"
+
+// authEvents is responsible for dispatching domain events throughout the system.
+// It does not express any infrastructural detail (like pubsub).
+type authEventPublisher interface {
```
sagikazarmark: authEvents seems to be a better name from a domain point of view. You don't need to care about different event responsibilities here other than publishing.

Either way, the interface name should match the name in the comment.

auth/events.go
```diff
@@ -0,0 +1,35 @@
...
+	OrganizationRegistered(organizationID uint)
+}
+
+type ebAuthEventPublisher struct {
+	eb events.EventBus
```
sagikazarmark: I don't see the need for a common, global interface. Go's implicit interface implementation makes it quite easy. A private interface here is less costly when building a mental model about the code than the extra dependency that it creates.

auth/events.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (e ebAuthEventPublisher) OrganizationRegistered(organizationID uint) {
+	e.eb.Publish(OrganizationRegisteredTopic, organizationID)
+}
+
+var authEventer authEventPublisher = ebAuthEventPublisher{eb: events.Bus}
```
sagikazarmark: I would really like these global instances go away. The consuming code is inside a struct, so injection could work there perfectly.

auth/user.go
```diff
@@ -245,6 +245,10 @@ func (bus BanzaiUserStorer) Save(schema *auth.Schema, context *auth.Context) (us
 		orgids = append(orgids, githubOrgIDs...)
 		AddOrgRoles(orgids...)
 		AddOrgRoleForUser(currentUser.ID, orgids...)
+
+		for _, orgID := range orgids {
+			authEventer.OrganizationRegistered(orgID)
```
sagikazarmark: Maybe just call this events?

internal/events/events.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package events
+
+import (
+	evbus "github.com/asaskevich/EventBus"
+)
+
+// eventBus interface separates the actual event handling implementation.
+type EventBus interface {
```
sagikazarmark: See my comment about the global interface.

internal/events/events.go
```diff
@@ -0,0 +1,29 @@
...
+	Publish(topic string, args ...interface{})
+	Subscribe(topic string, fn interface{}) error
+	SubscribeAsync(topic string, fn interface{}, transaction bool) error
+}
+
+// Bus is the global EventBus dispatcher object
+var Bus EventBus = evbus.New()
```
sagikazarmark: Do we actually need a global instance? Can't we just inject it where we needed? If we can't, can we move this near other global instances, like database and logging, etc?

spotguide/spotguide.go
```diff
@@ -51,6 +51,11 @@ const DeployApplicationStep = "deploy_application"
 
 var ctx = context.Background()
 
+func init() {
```
sagikazarmark: Better solution would be calling this somewhere in main, at least indirectly, but I can live with it as a temporary solution if it would require a huge amount of work now. Even then a big, fat TODO warning that it should be refactored at some point.

internal/events/events.go
bonifaido: Yes, the publisher and the subscriber have to use the same instance.

bonifaido:
Updated according to review comments.

PR #1140, opened 2018-10-03
Install secret

### LINKED ISSUE #1121: Add API endpoint for installing a single secret
We need to be able to install secrets in custom formats. The current secret installation call installs secrets matching a query which means it could install multiple secrets in one batch. IMHO that makes it non-ideal choice for custom secret format, hence the new API endpoint.

The API should support the following use cases:

### Remap keys of a secret and organize them in a structure
```json
PUT /api/v1/:orgid/clusters/:id/secrets/:secretId?
{
    "spec": {
        "docker.json": {
            "secretKeyMap": {
                "docker_username": "username",
                "docker_password": "password"
            }
        },
    },
}
```

### Remap keys of a secret
```json
PUT /api/v1/:orgid/clusters/:id/secrets/:secretId?
{
    "spec": {
        "tls.crt": {
            "secretKey": "client.crt"
        },
        "tls.key": {
            "secretKey": "client.key"
        },
    },
}
```

### Output the whole secret under a certain key (in JSON format)

```json
PUT /api/v1/:orgid/clusters/:id/secrets/:secretId?
{
    "spec": {
        "google.json": {},
    },
}
```

## Topics to clear

- [ ] Do we like the secretKey/secretKeyMap names? (Alternatives: `ref|value|spec|template|secretRef|valueRef|refValue|secretMap|sourceMap|sourceRef`)
- [ ] Is it okay to put the secret ID in the URL. The API is already quite far from being a REST API, this call is definitely not one, but IMO it would be nice to preserve some REST characteristics.

sagikazarmark:
Fixes  #1121


api/cluster_secret.go
```diff
@@ -0,0 +1,103 @@
...
+	Source    string            `json:"source,omitempty"`
+	SourceMap map[string]string `json:"sourceMap,omitempty"`
+}
+
+type InstallSecretResponse struct {
+	Name     string `json:"name"`
+	Sourcing string `json:"sourcing"`
+}
+
+// InstallSecretsoCluster installs a particular secret to a cluster's namespace.
```
stoader: Small typo

api/cluster_secret.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	secretName := c.Param("secretName")
+
+	secretSource, err := cluster.InstallSecret(commonCluster, secretName, secretRequest)
+
+	if err != nil {
+		errorHandler.Handle(emperror.With(
+			emperror.Wrap(err, "failed to install secret into cluster"),
+			"cluster-id", commonCluster.GetID(),
```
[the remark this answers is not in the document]
sagikazarmark: That works too, there was no agreed standard before.

sagikazarmark:
> Why do we have both openapi.yaml and pipeline.yaml. These look the be the same.

The first one is generated by the code generator.

client/api/openapi.yaml
```diff
@@ -1074,6 +1074,75 @@ paths:
...
+        name: secret
+        required: true
+        schema:
+          type: string
+        style: simple
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/InstallSecretsRequest'
```
orymate: `InstallSecretRequest` (singular)

sagikazarmark: Thanks, see #1158

cluster/secrets.go
```diff
@@ -84,37 +90,26 @@ func InstallSecretsByK8SConfig(k8sConfig []byte, orgID uint, query *secretTypes.
...
+		newK8sSecret, err := intSecret.CreateKubeSecret(kubeSecretRequest)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to create k8s secret")
 		}
 
 		if create {
-			_, err = clusterClient.CoreV1().Secrets(namespace).Create(k8sSecret)
+			newK8sSecret.ObjectMeta.Namespace = namespace
+
+			_, err = clusterClient.CoreV1().Secrets(namespace).Create(&k8sSecret)
```
orymate: this should be `Create(newK8sSecret)`, shouldn't it?

orymate: fix in #1166

PR #1161, opened 2018-10-08
fixed GetUsers

prekoa:


api/user.go
```diff
@@ -58,9 +58,16 @@ func GetUsers(c *gin.Context) {
 		})
 	} else if id == 0 {
 		c.JSON(http.StatusOK, users)
-	} else if len(users) == 1 {
-		c.JSON(http.StatusOK, users[0])
-	} else if len(users) > 1 {
+	}
+	var user []auth.User
```
stoader: Why do we need this extra filtering here?  The `err = db.Model(organization).Related(&users, "Users")` above should filter by user id if `id` is provided.

prekoa: thx, fixed

api/user.go
```diff
@@ -47,7 +47,11 @@ func GetUsers(c *gin.Context) {
 
 	var users []auth.User
 	db := config.DB()
-	err = db.Model(organization).Related(&users, "Users").Error
+	if id == 0 {
```
bonifaido: You can write this: `err = db.Model(organization).Where(&auth.User{ID: uint(id)}).Related(&users, "Users").Error` instead of the if block. When userId is 0 the Where clause gets ignored by GORM.

prekoa: thx

bonifaido:
Thx @prekoa !

PR #1165, opened 2018-10-09
#1094 - Implement Pipeline Policy API

sancyx:


internal/security/anchore.go
```diff
@@ -0,0 +1,237 @@
...
+		err := db.createAnchoreUser(anchorUserName, userPassword)
+		if err != nil {
+			return true, errors.WrapPrefix(err, "Error creating Anchor user", 0)
+		}
+	} else {
+		logger.Infof("Anchore user %v found", anchorUserName)
+		userPassword = user.Password
+	}
+
+	secretRequest := secret.CreateSecretRequest{
```
bonifaido: If you submit a CreateSecretRequest with type password, with an empty password field, the secret engine will generate a password for you.

PR #1175, opened 2018-10-10
openapi: fix InstallSecret response type

orymate:


orymate:
I changed the wrong response type. Could you reapprove, @baluchicken or @bonifaido ?

PR #1181, opened 2018-10-11
Implement list images by cluster and deployment API

pbalogh-sa:
fixes: #1174


api/images.go
```diff
@@ -0,0 +1,163 @@
...
+	imageList := make([]apiclient.ClusterImage, 0)
+	for _, pod := range podList {
+		images := getPodImages(pod)
+		imageList = append(imageList, images...)
+	}
+	removeDuplicatedImages(&imageList)
+	return imageList, nil
+}
+
+func getPodImages(pod v1.Pod) []apiclient.ClusterImage {
```
stoader: IMHO it's better to return []*apiclient.ClusterImage here.

api/images.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		images = append(images, image)
+	}
+	return images
+}
+
+func removeDuplicatedImages(images *[]apiclient.ClusterImage) {
+	found := make(map[string]bool)
+	j := 0
+	for i, image := range *images {
```
stoader: The loop will make a copy of each item when iterating through 'images'. Use '[]*apiclient.ClusterImage ' to avoid that.

Issue #1092, opened 2018-09-26
Google project support


colin014:
- [X] Add new API endpoint to list projects
- [X] Make `project` field in Create secret request to optional
- [X] Add `project` field to Create cluster request, and save it.
- [X] Load project from database (everywhere) instead of Vault secret, but if it's empty load from secret. If it's still empty throw an error

PR #1184, opened 2018-10-11
endpoint and API for listing projects for a given secret on google

lpuskas:
- new endpoint to retrieve google projects visible for a given secret
- extended the openapi spec with the new call
- generated client
- made project field optional when creating google secrets

relates to #1092


api/projects.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/banzaicloud/pipeline/internal/platform/gin/utils"
+	"github.com/banzaicloud/pipeline/pkg/providers"
+	"github.com/banzaicloud/pipeline/secret/verify"
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+type ListProjectsResponse struct {
+	Projects []*cloudresourcemanager.Project
```
bonifaido: Shouldn't be this tagged with `json` tags?

client/api/openapi.yaml
```diff
@@ -8289,6 +8337,41 @@ components:
         vulnerabilities:
           $ref: '#/components/schemas/VulnerabilityList'
       type: object
+    ListProjectsResponse:
+      description: wraps a list of project details information
+      example:
+        Projects:
```
bonifaido: should be `projects` after the struct tagging.

api/projects.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil
+	}); err != nil {
+		log.WithError(err).Error("could not retrieve projects")
+		ginutils.ReplyWithErrorResponse(c, errorResponseFrom(err))
+	}
+
+}
+
+// HttpClient builds a http client with the service account available through the secret and organization
+func (sc *servicesContext) HttpClient() (*http.Client, error) {
```
stoader: Since this is not used from other packages probably shouldn't be public.

api/projects.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	log.Debugf("retrieving projects for orgId: [%d], secretId [%s]",
+		organization.ID, secretID)
+	servicesCtx := newServicesCtx(organization.ID, secretID)
+
+	cli, err := servicesCtx.HttpClient()
+	if err != nil {
+		log.WithError(err).Error("could not build http client")
+		ginutils.ReplyWithErrorResponse(c, errorResponseFrom(err))
+	}
```
colin014: I think a return is missing here

api/projects.go
```diff
@@ -0,0 +1,113 @@
...
+
+	projectsSvc, err := servicesCtx.ProjectsService(cli)
+	if err != nil {
+		log.WithError(err).Error("could not build projects service")
+		ginutils.ReplyWithErrorResponse(c, errorResponseFrom(err))
```
colin014: Same as above

PR #1185, opened 2018-10-12
ARK config and logging enhancements

waynz0r:
* Separate log level
 * Configuration option to turn syncs on/off
 * Make sync intervals configurable


config/config.toml.example
```diff
@@ -122,7 +122,12 @@ sleepSecondsForNodepoolActive = 30
...
-chart_version = "1.2.1"
+chartVersion = "1.2.1"
 image = "banzaicloud/ark"
-imagetag = "v0.9.6"
-pullpolicy = "IfNotPresent"
+imageTag = "v0.9.6"
+pullPolicy = "IfNotPresent"
+syncEnabled = true
+logLevel = "info"
+bucketsSyncInterval = "10m"
```
sagikazarmark: Singular form sounds better IMHO

internal/ark/sync/run.go
```diff
@@ -34,7 +34,26 @@ func RunSyncServices(
 	db *gorm.DB,
 	logger logrus.FieldLogger,
 	errorHandler emperror.Handler,
+	bucketsSyncInterval, restoresSyncInterval, backupsSyncInterval time.Duration,
 ) {
+	if bucketsSyncInterval.Seconds() < 1 {
+		logger.WithField("interval-seconds", bucketsSyncInterval.Seconds()).Error("invalid buckets sync interval")
```
sagikazarmark: Do we need the "-seconds" suffix?

PR #1189, opened 2018-10-12
bucket information decorated with secret id and name if requested

lpuskas:
- extended list managed buckets response with secret information (id, name)
- to include the secret name the get request should call the endpoint with the query param `include=secret` ({{url}}/api/v1/orgs/:orgId/buckets?include=secret)
- the former behaviour was left intact
- openapi spec updated + client regenerated


api/bucket.go
```diff
@@ -123,6 +123,17 @@ func ListBuckets(c *gin.Context) {
...
 func ListManagedBuckets(c *gin.Context) {
+
+	type secretData struct {
+		SecretId   string `json:"secretId"`
+		SecretName string `json:"secretName"`
+	}
+	type bucketWithSecretName struct {
+		*objectstore.BucketInfo
+		SecretRef  string      `json:"secretId,-,omitempty"`
+		SecretInfo *secretData `json:"secretInfo"`
```
stoader: The `secretId` in `secretData` struct seems to be redundant as that field already available through `BucketInfo`.

Can't we just do
```
type bucketWithSecretName struct {
  *objectstore.BucketInfo
  SecretName string `json:"secretName,omitempty"`
}
```

lpuskas: i have slightly changed this - these json binding tricks are only for returning nicer jsons to callers (json response is enhanced with data that isn;t required in the "api" or "domain" structs

api/bucket.go
```diff
@@ -134,6 +153,13 @@ func ListManagedBuckets(c *gin.Context) {
 		pkgProviders.Oracle,
 	}
 
+	const (
+		fieldsQueryKey = "includes"
```
tothandras: `includes` -> `include`?

lpuskas: corrected, thx

PR #1196, opened 2018-10-14
Goland and Helm local dev support

ahma:


Dockerfile.dev
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.10-alpine
```
sagikazarmark: This should be `golang:1.11-alpine`

Dockerfile.dev
```diff
@@ -0,0 +1,13 @@
...
+RUN apk add --update --no-cache ca-certificates git
+RUN go get github.com/kubernetes-sigs/aws-iam-authenticator/cmd/aws-iam-authenticator
+RUN go get github.com/derekparker/delve/cmd/dlv
+
+FROM alpine:3.6
```
sagikazarmark: The latest alpine version is 3.7

Dockerfile.dev
```diff
@@ -0,0 +1,13 @@
...
+RUN apk add --no-cache ca-certificates
```
sagikazarmark: As far as I know, Delve requires `libc6-compat` on alpine to work properly. Did this image work without it?

Dockerfile.dev
```diff
@@ -0,0 +1,13 @@
...
+COPY --from=0 /go/bin/aws-iam-authenticator /usr/bin/
+COPY --from=0 /go/bin/dlv /
+ADD ./pipeline-debug /
+ADD ./views /views/
+
+ENTRYPOINT ["/dlv", "--listen=:40000", "--headless=true", "--api-version=2", "--log", "exec", "/pipeline-debug"]
```
sagikazarmark: Missing new line from the end of file

Makefile
```diff
@@ -31,6 +31,15 @@ build: ## Builds binary package
 	@go version | grep -q -E "go${GOLANG_VERSION_REGEX} " || (echo "Required Go version is ${GOLANG_VERSION}\nInstalled: `go version`" && exit 1)
 	go build -v -ldflags "-X main.Version=${VERSION} -X main.GitRev=${GITREV} -X main.BuildDate=${BUILD_DATE}" .
 
+.PHONY: debug
+debug: ## Builds binary package
```
sagikazarmark: I think it would be better to reuse the same build target and modify it's behaviour using variables.

Here is an example for that:

https://github.com/sagikazarmark/modern-go-application/blob/master/Makefile

Fixing this can be part of a subsequent PR though

scripts/debug_builder_proxy.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -e
```
sagikazarmark: Some newlines would be nice for better readability

PR #1199, opened 2018-10-14
Allow passing a custom secret name to the install secret call

sagikazarmark:
Note that the name in the URL is the cluster secret name (since this is the resource we edit in this call), and the optional parameter is the pipeline secret name. When it is not provided, it falls back to the cluster secret name.


client/api/openapi.yaml
```diff
@@ -1099,7 +1099,7 @@ paths:
           format: int32
           type: integer
         style: simple
-      - description: Secret name
+      - description: Secret name as it will be seen in the cluster
         explode: false
         in: path
         name: secret
```
bonifaido: `secret` should be renamed to `secretName`

orymate:
as we discussed: please rename the new value to sourceSecretName to resolve the collision between the two secretNames

PR #1210, opened 2018-10-16
List releases for images

tarokkk:


api/security.go
```diff
@@ -381,3 +383,60 @@ func DeletePolicy(c *gin.Context) {
...
+	}
+	var releaseList []string
+	for k := range releaseMap {
+		releaseList = append(releaseList, k)
+	}
+
+}
+
+func getImageDigest(imageID string) string {
+	image := strings.Split(imageID, "@")
```
sancyx: Probably we should make some validation here on imageId, with regexp or something.

tarokkk: added basic validation

PR #1212, opened 2018-10-16
remove helm retry (eks unauthorized error fixed)

prekoa:


helm/kubernetes.go
```diff
@@ -34,13 +33,16 @@ var tillerTunnel *kube.Tunnel
...
 func GetK8sConnection(kubeConfig []byte) (*kubernetes.Clientset, error) {
+	var client *kubernetes.Clientset
 	config, err := GetK8sClientConfig(kubeConfig)
 	if err != nil {
 		return nil, fmt.Errorf("create kubernetes config failed: %v", err)
 	}
-	client, err := kubernetes.NewForConfig(config)
-	if err != nil {
-		return nil, fmt.Errorf("create kubernetes connection failed: %v", err)
+	for i := 0; i < 2; i++ {
```
baluchicken: please write a comment why we need to get the client for two times without an error.

cluster/hooks.go
```diff
@@ -32,7 +32,9 @@ import (
 	"github.com/banzaicloud/pipeline/internal/security"
 	pkgCluster "github.com/banzaicloud/pipeline/pkg/cluster"
 	pkgCommon "github.com/banzaicloud/pipeline/pkg/common"
+	helm2 "github.com/banzaicloud/pipeline/pkg/helm"
```
baluchicken: this import is present as pkgHelm so it is unnecessary to import as helm2 again

cluster/hooks.go
```diff
@@ -32,7 +32,9 @@ import (
...
 	pkgHelm "github.com/banzaicloud/pipeline/pkg/helm"
```
baluchicken: maybe we should rename this to helm2 because in every other place this import called that way

prekoa: thx, fixed

PR #1213, opened 2018-10-16
Disable schema automigration

sagikazarmark:
This PR disables automatic migration of database schema by default.

To enable it locally add the following line to your config's database section:

```
autoMigrateEnabled = true
```

### DIFF
--- a/model/defaults/migrate.go
+++ b/model/defaults/migrate.go
@@ -22,7 +22,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// Migrate executes the table migrations for the auth module.
+// Migrate executes the table migrations for the defaults module.
 func Migrate(db *gorm.DB, logger logrus.FieldLogger) error {
 	tables := []interface{}{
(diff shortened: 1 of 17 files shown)


model/defaults/migrate.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/jinzhu/gorm"
+	"github.com/sirupsen/logrus"
+)
+
+// Migrate executes the table migrations for the auth module.
```
waynz0r: bad comment

spotguide/migrate.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/jinzhu/gorm"
+	"github.com/sirupsen/logrus"
+)
+
+// Migrate executes the table migrations for the auth module.
```
waynz0r: bad comment

PR #1232, opened 2018-10-18
Skip already deleted resources during EKS cluster deletion

stoader:


cluster/manager_delete.go
```diff
@@ -137,7 +137,7 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 		err = helm.DeleteAllDeployment(c)
 		if err != nil {
 			if force {
-				logger.Errorf("deleting all deployments failed: %s", err.Error())
+				logger.Errorln("deleting deployments failed: ", err.Error())
```
bonifaido: When using `...-ln` methods you don't need to add the extra space in between the arguments.

PR #1237, opened 2018-10-18
EKS: wait for ASGs

waynz0r:
Wait for ASGs to get into desired state in cluster create/update requests

### DIFF
--- a/pkg/providers/amazon/autoscaling/autoscaling.go
+++ b/pkg/providers/amazon/autoscaling/autoscaling.go
@@ -1,0 +1,125 @@
+// Copyright © 2018 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package autoscaling
+
+import (
+	"github.com/aws/aws-sdk-go/service/autoscaling"
+	awsEC2 "github.com/aws/aws-sdk-go/service/ec2"
+
+	"github.com/banzaicloud/pipeline/pkg/providers/amazon/ec2"
+)
+
+// Group extends autoscaling.Group
+type Group struct {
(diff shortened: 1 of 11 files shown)


pkg/providers/amazon/autoscaling/autoscaling.go
```diff
@@ -0,0 +1,125 @@
...
+	if lc.SpotPrice != nil && *lc.SpotPrice == "" {
+		return nil, nil
+	}
+
+	requests := make([]*ec2.SpotInstanceRequest, 0)
+	for _, res := range result.SpotInstanceRequests {
+		if *res.LaunchSpecification.IamInstanceProfile.Name == *lc.IamInstanceProfile {
+			requests = append(requests, ec2.NewSpotInstanceRequest(res))
+		}
+	}
```
martonsereg: I have a question about this, because I think it may not be enough to properly match spot instance requests to auto-scaling-groups in general. `IamInstanceProfile` can be empty or they can be reused among multiple AS groups.
Or maybe it's a very Pipeline-specific thing, because it's always creating a new instance profile for every AS group?
Are we sure that it will stay this way?

I've looked into it and the only way I've found to match these 2 things is this call, that seems to list the spot instance requests (but only while those are still pending, and it can only be parsed from the status message):
```
aws autoscaling describe-scaling-activities --region us-east-1 --auto-scaling-group-name marton-teszt2
{
    "Activities": [
        {
            "ActivityId": "9ae59a2f-c5bd-52b3-7233-3f245bd5e73d",
            "AutoScalingGroupName": "marton-teszt2",
            "Description": "Placed Spot instance request.  Status Reason: Placed Spot instance request: sir-wt5856dj. Waiting for instance(s)",
            "Cause": "At 2018-10-19T10:49:04Z a user request created an AutoScalingGroup changing the desired capacity from 0 to 1.  At 2018-10-19T10:49:07Z a difference between desired and actual capacity changing the desired capacity, increasing the capacity from 0 to 1.",
            "StartTime": "2018-10-19T10:49:07.668Z",
            "StatusCode": "WaitingForSpotInstanceId",
            "StatusMessage": "Placed Spot instance request: sir-wt5856dj. Waiting for instance(s)",
            "Progress": 20,
            "Details": "{\"Subnet ID\":\"subnet-0b421a12a03098de4\",\"Availability Zone\":\"us-east-1a\"}"
        }
    ]
}
```

Can you please check it out if this call can be used instead of instance profiles?

waynz0r: We create an instance profile for every ASG right now. It seems that currently it is the best possible way to match the spot requests for a given ASG. I commented the code about this necessity.

PR #1238, opened 2018-10-18
Block huge zero tars

ahma:


helm/helm_test.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func TestDownloadFile_TooBig(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Length", strconv.FormatInt(maxCompressedDataSize+1, 10))
+	}))
+	defer ts.Close()
+
+	_, err := DownloadFile(ts.URL)
+	assert.EqualError(t, err, "chart data is too big")
```
stoader: You could check the type of the returned error whether is of `chartDataIsTooBigError` type as well

PR #1244, opened 2018-10-19
EKS: get removable nodepool stack names from CF too

waynz0r:


pkg/providers/amazon/cloudformation/cloudformation.go
```diff
@@ -0,0 +1,54 @@
...
+	names := make([]string, 0)
+
+	err := cfSvc.DescribeStacksPages(&cloudformation.DescribeStacksInput{}, func(page *cloudformation.DescribeStacksOutput, lastPage bool) bool {
+		for _, stack := range page.Stacks {
+			stackTags := getFlattenedTags(stack.Tags)
+			ok := true
+			for k, v := range tags {
+				if stackTags[k] != v {
+					ok = false
+					continue
```
bonifaido: This should be `break`, right?

waynz0r: sure, @stoader fixed it in the #1243 which also contains this PR's changes

PR #1248, opened 2018-10-20
[WIP ]Fix: wrong AWS creds passed to External DNS

stoader:


cluster/hooks.go
```diff
@@ -894,11 +894,16 @@ func RegisterDomainPostHook(input interface{}) error {
 		log.Infof("Domain '%s' already registered", domain)
 	}
 
-	secretSources, err := InstallSecrets(
+	route53Secret, err := secret.Store.GetByName(orgId, route53.IAMUserAccessKeySecretName)
+	if err != nil {
+		log.Errorf("Failed to install route53 secret into cluster: %s", err.Error())
+		return err
```
sagikazarmark: Can't the error be passed to an error handler in the parent instead? (Actually in the cluster flow it is I think), so double logging is not just unnecessary, but harmful as well.

You can wrap the error if necessary.

PR #1273, opened 2018-10-29
Limit spotguide sync API requests to 1 per 2 minutes by organization

tothandras:


spotguide/spotguide.go
```diff
@@ -48,7 +48,6 @@ const PipelineYAMLPath = ".banzaicloud/pipeline.yaml"
 const ReadmePath = ".banzaicloud/README.md"
 const IconPath = ".banzaicloud/icon.svg"
 const CreateClusterStep = "create_cluster"
-const DeployApplicationStep = "deploy_application"
```
sagikazarmark: This change is totally unrelated here. Is it by accident?

cmd/pipeline/main.go
```diff
@@ -204,7 +205,7 @@ func main() {
 			orgs.Use(api.OrganizationMiddleware)
 
 			orgs.GET("/:orgid/spotguides", api.GetSpotguides)
-			orgs.PUT("/:orgid/spotguides", api.SyncSpotguides)
+			orgs.PUT("/:orgid/spotguides", middleware.NewRateLimiterByOrgID(1.0/60/2 /* 1 request per 2 minutes */), api.SyncSpotguides)
```
sagikazarmark: This value should be extracted to a constant, somewhere near the action itself.

spotguide/spotguide.go
sagikazarmark: > The change itself is not related, could be in a different commit.

I think it should be a separate PR, otherwise we won't know why it changed. CS fixes and stuff like that doesn't matter, but this is totally unrelated here.

sagikazarmark: We discussed it privately: the next time we have a confusion because of an unrelated change, @tothandras pays lunch for a week for everyone. 😄

Issue #1285, opened 2018-10-30
Remove the created VCN if cluster create fails


waynz0r:
Pipeline should remove the automatically created preconfigured VCN if the cluster creation fails.

PR #1297, opened 2018-11-04
OKE enhancements

### LINKED ISSUE #1284: Return OKE service limit errors of node pool operations
Pipeline should log and report service limit errors happens during node pool operations.

waynz0r:
- Remove the preconfigured VCN if cluster creation request validation fails
- Return node errors raised during node pool manipulations
- Trying to get OKE cluster by it's name if OCID is empty
- Fix node pool config persistance in case of an error

Fixes #1284 and #1285

### DIFF
--- a/pkg/providers/oracle/oci/ce_cluster.go
+++ b/pkg/providers/oracle/oci/ce_cluster.go
@@ -21,6 +21,7 @@ import (
 
 	pipConfig "github.com/banzaicloud/pipeline/config"
 	"github.com/go-errors/errors"
+	"github.com/goph/emperror"
 	"github.com/oracle/oci-go-sdk/common"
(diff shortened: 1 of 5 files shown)


pkg/providers/oracle/oci/ce_cluster.go
```diff
@@ -192,10 +192,11 @@ func (ce *ContainerEngine) WaitingForClusterNodePoolActiveState(clusterID *strin
 			return err
 		}
 
-		ok := true
+		ok := false
 		for _, np := range nodePools {
-			if !ce.IsNodePoolActive(np.Id) {
-				ok = false
+			ok, err = ce.IsNodePoolActive(np.Id)
```
bonifaido: There should be an early return here if `!ok`.

PR #1311, opened 2018-11-06
Do not render markdown as HTML

sagikazarmark:


bonifaido:
I think blackfriday can be removed now from Gopkg.toml as well.

PR #1313, opened 2018-11-06
Fix volume size for traefik and prometheus in case of alibaba

baluchicken:


cluster/hooks.go
```diff
@@ -218,6 +218,13 @@ func InstallMonitoring(input interface{}) error {
 			},
 		},
 	}
+	if cluster.GetCloud() == pkgCluster.Alibaba {
+		prometheusVolumeSize := grafanaValues["prometheus"].(map[string]interface{})["server"].(map[string]interface{})
+		prometheusVolumeSize["persistentVolume"] = map[string]interface{}{
```
stoader: Can't we set the value of "size" directly instead of creating a new map?

baluchicken:
After further discussion we are going to handle this in a different PR

PR #1325, opened 2018-11-08
add pipeline metrics

prekoa:


cluster/pipeline_monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/banzaicloud/pipeline/config"
+	intCluster "github.com/banzaicloud/pipeline/internal/cluster"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type PipelineMetrics struct {
+	clusterStatus *prometheus.GaugeVec
+
+	metricsMtx sync.RWMutex
+	sync.RWMutex
```
sagikazarmark: Do not embed sync primitives. Also, mutexes are usually called "mu" as a variable. Also, do you need RWMutex? As far as I can see Mutex is enough. Also, do you need to mutexes?

cluster/pipeline_monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"sync"
+
+	"github.com/banzaicloud/pipeline/config"
+	intCluster "github.com/banzaicloud/pipeline/internal/cluster"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type PipelineMetrics struct {
```
sagikazarmark: This struct does not have to be exported as we use it as a prometheus exporter, implementing an interface.

cluster/pipeline_monitoring.go
```diff
@@ -0,0 +1,115 @@
...
+	go p.scrape(clusterTotal)
+	p.setClusterMetrics(clusterTotal)
+
+	p.clusterStatus.Collect(ch)
+}
+
+func (p *PipelineMetrics) scrape(scrapesTotalCluster chan<- scrapeResultTotalCluster) {
+
+	defer close(scrapesTotalCluster)
+	clusters := intCluster.NewClusters(config.DB())
```
sagikazarmark: This instance should be injected into the pipeline metrics struct, to avoid initializing it every time scrape is called. Generally, avoid global scope instances.

cluster/pipeline_monitoring.go
prekoa: fixed, thx

PR #1331, opened 2018-11-08
Followup on Use VolumeBindingWaitForFirstConsumer in EKS use only in …

baluchicken:
…case of kubernetes version 1.12


cluster/eks.go
```diff
@@ -262,8 +263,22 @@ func (c *EKSCluster) CreateCluster() error {
 		return err
 	}
 
+	constraint, err := semver.NewConstraint(">= 1.12")
+	if err != nil {
+		return err
```
stoader: return error that contains stack info as well

cluster/eks.go
```diff
@@ -262,8 +263,22 @@ func (c *EKSCluster) CreateCluster() error {
...
+	}
+	kubeVersion, err := semver.NewVersion(c.modelCluster.EKS.Version)
+	if err != nil {
+		return err
```
stoader: return error that contains stack info as well

PR #1342, opened 2018-11-09
Fix missing client code generation

bonifaido:
If this will be rebased to master it will get fixed since the YAML is fixed there already.


bonifaido:
Failed, so far so good.

PR #1344, opened 2018-11-09
Delete cluster fixes

orymate:
deleteCluster: improve error reporting

DeleteCluster: clean up Kubernetes resources properly  …
Clean up as much Kubernetes resources from the cluster as possible to
prevent leaking cloud resources.

The old implementation
* did not work in any case
* did not consider PVCs and single Pods
* did retry even if the deletion succeeded (but never did)


cluster/common.go
```diff
@@ -138,7 +141,9 @@ func (c *CommonClusterBase) getConfig(cluster CommonCluster) ([]byte, error) {
 		log.Debug("k8s config is nil.. load from vault")
 		var loadedConfig []byte
 		configSecret, err := getSecret(cluster.GetOrganizationId(), cluster.GetConfigSecretId())
-		if err != nil {
+		if err == secret.ErrSecretNotExists {
+			return nil, ErrConfigNotExists
```
stoader: Should we enrich error message with cluster name for which the config was not found or rather do that in all of the callers?

cluster/manager_delete.go
```diff
@@ -56,38 +54,177 @@ func (m *Manager) DeleteCluster(ctx context.Context, cluster CommonCluster, forc
 	return nil
 }
 
-func deleteAllResource(kubeConfig []byte, logger *logrus.Entry) error {
+func retry(function func() error, count int, delaySeconds int) error {
```
stoader: This function should probably be factored out into a utility package.

sagikazarmark: Definitely not. Utility packages are evil. Until we have proper test coverage talking about code reusability makes no sense at all.

BUT: I would rather say, that we should probably use some sort of backoff library: https://medium.com/@lestrrat/yak-shaving-with-backoff-libraries-in-go-80240f0aa30c

Ultimately this code go into a pkg/backoff package itself, but there are already great implementations for that.

orymate: we can always move this to a utility package, if it can be used at least in another source file. we could also add an external dependency, but the boilerplate to use any of them is longer than this function alone. i would not change this for now.

PR #1352, opened 2018-11-12
Gke enable preemptible nodes

sancyx:


martonsereg:
thanks @sancyx, since [this PR](https://github.com/banzaicloud/pipeline/pull/1359) was merged earlier, can you add the GKE implementation of the general `spot` field?

Issue #1133, opened 2018-10-02
Pin the IAM authenticator to specific version from the next release


sagikazarmark:
As soon as the IAM authenticator reaches it's next release.

PR #1357, opened 2018-11-12
Use specific version of aws-iam-authenticator

bonifaido:
Fixes: https://github.com/banzaicloud/pipeline/issues/1133


sagikazarmark:
we should probably use prebuilt binaries

PR #1359, opened 2018-11-13
Add `spot` flag to cluster status response

martonsereg:
The new flag signals if the cluster has at least one node pool with spot instances configured.

This is needed for the UI implementation of setting "minimum on-demand percentages for replicas".
The GKE implementation will need to be updated once we have [this PR](https://github.com/banzaicloud/pipeline/pull/1352) merged.

Example cluster status call:
GET {{url}}/api/v1/orgs/:orgId/clusters/:clusterId
```
{
    "status": "ERROR",
    "statusMessage": "failed to create kubernetes tunnel: Unauthorized",
    "name": "ekscluster-marton-177",
    "location": "us-east-1",
    "cloud": "amazon",
    "distribution": "eks",
*   "spot": true,
    "version": "1.10",
    "id": 45,
    "nodePools": {
        "pool1": {
            "count": 1,
            "instanceType": "m5.xlarge",
            "spotPrice": "0",
            "minCount": 1,
            "maxCount": 1,
            "image": "ami-0440e4f6b9713faf6"
        },
        "pool2": {
            "count": 1,
            "instanceType": "c5.xlarge",
            "spotPrice": "0.4",
            "minCount": 1,
            "maxCount": 2,
            "image": "ami-0440e4f6b9713faf6"
        }
    },
    "createdAt": "2018-11-09T09:51:17+01:00",
    "creatorName": "martonsereg",
    "creatorId": 1
}
```

### DIFF
--- a/pkg/cluster/base.go
+++ b/pkg/cluster/base.go
@@ -162,6 +162,7 @@ type GetClusterStatusResponse struct {
 	Location      string                     `json:"location"`
 	Cloud         string                     `json:"cloud"`
 	Distribution  string                     `json:"distribution"`
+	Spot          bool                       `json:"spot"`
 	Version       string                     `json:"version,omitempty"`
 	ResourceID    uint                       `json:"id"`
 	NodePools     map[string]*NodePoolStatus `json:"nodePools,omitempty"`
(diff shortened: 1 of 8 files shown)


pkg/cluster/base.go
```diff
@@ -162,6 +162,7 @@ type GetClusterStatusResponse struct {
 	Location      string                     `json:"location"`
 	Cloud         string                     `json:"cloud"`
 	Distribution  string                     `json:"distribution"`
+	Spot          bool                       `json:"spot"`
```
stoader: Shouldn't this have `omitempty` annotation since this field is optional?

martonsereg: We can add it, but `bool` will always have `false` as the default value in go, so it will never really be empty.

martonsereg: I've added it just in case.

PR #1377, opened 2018-11-14
added wait option to create deployment

martonsereg:
From helm docs:

> if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment are in a ready state before marking the release as successful

It's fixing the issue with spot mutating webhook that didn't wait until the webhook started before moving on the next deployment.

When sending a create deployment request, it is now configurable on the API.

I've set wait to `false` for all default posthooks other then spot-config-webhook. It was the default before, and setting it to true may increase the cluster creation time.


martonsereg:
After a discussion with @baluchicken, I've set `wait` to `true` for the `logging-operator` as well.

PR #1378, opened 2018-11-14
EC2 instance startup and spot request fulfillment metrics

waynz0r:


pkg/providers/amazon/autoscaling/prometheus.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	key := *instance.InstanceId
+	if timers[key] != nil {
+		return timers[key]
+	}
+
+	m.logger.WithField("instance-id", *instance.InstanceId).Debug("start metric timer")
+
+	lc, err := instance.GetLaunchConfiguration()
```
martonsereg: instead of relying on `LaunchConfigurations` to determine if the instance is spot or not, please consider using `DescribeInstances` instead and parsing `instanceLifecycle` from the response. There is only a slim chance, but `LaunchConfigs` may change or be deleted after the instance is started.

waynz0r: Sure, coming right up! 👍

PR #1380, opened 2018-11-14
More exact build version

bonifaido:
`git describe --tags` shows out the correct tag (currently it writes out master or HEAD).


Makefile
```diff
@@ -11,7 +11,7 @@ OPENAPI_DESCRIPTOR = docs/openapi/pipeline.yaml
 # Build variables
 BUILD_DIR ?= build
 BUILD_PACKAGE = ${PACKAGE}/cmd/pipeline
-VERSION ?= $(shell git rev-parse --abbrev-ref HEAD)
+VERSION ?= $(shell git describe --tags)
```
sagikazarmark: This will output the latest tag even where it should output the branch name. Try this:

https://stackoverflow.com/questions/18659425/get-git-current-branch-tag-name

PR #1381, opened 2018-11-14
fixed listing buckets-buckets being deleted are ignored

lpuskas:
* buckets being deleted are ignored while retrieving
* sneaked in a few error - related fixes (contextual errors, proper and consistent messages)

### DIFF
--- a/internal/providers/amazon/objectstore.go
+++ b/internal/providers/amazon/objectstore.go
@@ -293,7 +293,11 @@ func (s *objectStore) ListBuckets() ([]*objectstore.BucketInfo, error) {
 			bucketInfo.Location = region
 			bucketList = append(bucketList, bucketInfo)
 		} else {
-			logger.WithField("bucket", bucket).WithError(err).Warn("skipping bucket from the list")
+			if _, ok := err.(bucketNotFoundError); ok {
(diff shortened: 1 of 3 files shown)


internal/providers/amazon/objectstore.go
```diff
@@ -269,41 +263,38 @@ func (s *objectStore) CheckBucket(bucketName string) error {
...
 		}
 
-		region, err := s.objectStore.GetRegion(bucket)
-		if err != nil {
-			return nil, err
+		if region, err := s.objectStore.GetRegion(bucket); err == nil {
+			bucketInfo.Location = region
+			bucketList = append(bucketList, bucketInfo)
+		} else {
+			logger.WithField("bucket", bucket).WithError(err).Warn("skipping bucket from the list")
```
martonsereg: I think we may need to differentiate between error codes here: if the bucket is not found, it can be skipped, but this API call can fail for other reasons as well, like missing permissions.
In those cases we should still return with an error.
see [docs](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html) about error handling

lpuskas: @martonsereg  thanks for spotted this out; fixed

PR #1382, opened 2018-11-14
Refactor authorization middleware

sagikazarmark:


internal/auth/middleware.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+type enforcer interface {
+	Enforce(rvals ...interface{}) bool
+}
+
+// NewMiddleware returns a new gin middleware that checks user authorization.
+func NewMiddleware(e enforcer, basePath string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		a := &middleware{
```
bonifaido: This could be created outside the anonymous function in `NewMiddleware` and captured.

sagikazarmark: Good catch

PR #1384, opened 2018-11-14
Refactor ASG instance fulfillment wait

waynz0r:
Refactor ASG instance fulfillment wait from separate action to the `CreateUpdateNodePoolStackAction` action.


pkg/cluster/eks/action/actions.go
```diff
@@ -822,6 +879,14 @@ func (a *CreateUpdateNodePoolStackAction) ExecuteAction(input interface{}) (outp
 		}
 	}
 
+	// wait for goroutines to finish
+	for i := 0; i < waitRoutines; i++ {
+		waitErr := <-waitChan
+		if waitErr != nil {
+			err = waitErr
```
stoader: Wouldn't this hide the errors collected in the above loop? Should we introduce an aggregate error that collects all errors and return the aggregate error to the caller?

waynz0r: 👍 As we discussed an error aggregator will be introduced in an upcoming PR.

PR #1385, opened 2018-11-15
Wait more for namespace termination during cluster deletion

orymate:
In more than one case the deletion of a cluster with monitoring timeouted because the deleted pipeline-system namespace was still in terminating state after 3 minutes.

### DIFF
--- a/cluster/manager_delete.go
+++ b/cluster/manager_delete.go
@@ -25,6 +25,7 @@ import (
 	pkgCluster "github.com/banzaicloud/pipeline/pkg/cluster"
 	"github.com/banzaicloud/pipeline/pkg/k8sclient"
 	"github.com/goph/emperror"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


cluster/manager_delete.go
```diff
@@ -129,14 +129,14 @@ func deleteUserNamespaces(kubeConfig []byte, logger *logrus.Entry) error {
 				continue
 			default:
 				logger.Infof("namespace %q still %s", ns.Name, ns.Status)
-				left++
+				left = append(left, ns.Name)
```
stoader: Create structured log. See as an example how is done in `func (m *Manager) deleteCluster`.

cluster/manager_delete.go
```diff
@@ -129,14 +129,14 @@ func deleteUserNamespaces(kubeConfig []byte, logger *logrus.Entry) error {
...
 			}
 		}
-		if left > 0 {
-			return fmt.Errorf("%d namespaces remained after deletion", left)
+		if len(left) > 0 {
+			return fmt.Errorf("%d namespaces remained after deletion: %v", len(left), left)
```
stoader: use `erros.Errorf` to capture stack info as well or `emperror.With(errors.New("namespaces remained after deletion"), "count", Len(left), "namespaces", left)`

orymate: Thanks, added context/stack. I would keep the list in the message, so the user can see it without specifically handling this error somewhere in the callers.

PR #1386, opened 2018-11-15
Fix make debug

pbalogh-sa:
- make debug now build linux binary
- Dockerfile.local entrypoint /pipeline-debug (without delve)
- add vscode remote debug option


Makefile
```diff
@@ -96,11 +96,17 @@ docker-build: ## Builds go binary in docker image
 debug: GOARGS += -gcflags "-N -l"
 debug: BINARY_NAME = pipeline-debug
 debug: build ## Builds binary package
+debug: GO_OS = linux
```
sagikazarmark: use `export GOOS = linux` instead

Makefile
```diff
@@ -96,11 +96,17 @@ docker-build: ## Builds go binary in docker image
...
+debug: GO_ARCH = amd64
```
sagikazarmark: This is the default, not necessary

Makefile
```diff
@@ -86,7 +86,7 @@ build: ## Build a binary
 ifneq (${IGNORE_GOLANG_VERSION_REQ}, 1)
 	@printf "${GOLANG_VERSION}\n$$(go version | awk '{sub(/^go/, "", $$3);print $$3}')" | sort -t '.' -k 1,1 -k 2,2 -k 3,3 -g | head -1 | grep -q -E "^${GOLANG_VERSION}$$" || (printf "Required Go version is ${GOLANG_VERSION}\nInstalled: `go version`" && exit 1)
 endif
-	go build ${GOARGS} ${BUILD_PACKAGE}
+	GOOS=${GO_OS} GOARCH=${GO_ARCH} go build ${GOARGS} ${BUILD_PACKAGE}
```
sagikazarmark: Once export is used, this is not necessary anymore.

Dockerfile.local
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.11-alpine
+RUN apk add --update --no-cache ca-certificates git
+RUN go get -d github.com/kubernetes-sigs/aws-iam-authenticator/cmd/aws-iam-authenticator
+RUN cd $GOPATH/src/github.com/kubernetes-sigs/aws-iam-authenticator && \
+    git checkout 981ecbe && \
+    go install ./cmd/aws-iam-authenticator
+
+FROM alpine:3.7
+RUN apk add --no-cache ca-certificates libc6-compat
```
sagikazarmark: libc6-compat is not necessary here since we don't use delve.
Certificates can be copied from the previous layer

sagikazarmark: But there is no delve anywhere around this docker image

PR #1399, opened 2018-11-19
Export spot request metrics collected from running EKS clusters

waynz0r:


internal/monitor/spotmetrics.go
```diff
@@ -0,0 +1,143 @@
...
+	credentials, err := config.Credentials.Get()
+	if err != nil {
+		return nil, err
+	}
+
+	key := *config.Region + "-" + credentials.AccessKeyID
+	if e.ec2Clients[key] == nil {
+		sess, err := session.NewSession(&config)
+		if err != nil {
+			return nil, err
```
stoader: Use `emperror` to collect stack and provide error context

internal/monitor/spotmetrics.go
```diff
@@ -0,0 +1,152 @@
...
+// Run runs the metrics collections with the given interval
+func (e *spotMetricsExporter) Run(interval time.Duration) {
+	e.logger.WithField("interval", interval.String()).Debug("collecting spot request metrics from EKS clusters")
+	err := e.collectMetrics()
+	if err != nil {
+		e.errorHandler.Handle(emperror.Wrap(err, "could not collect spot metrics"))
+	}
+
+	ticker := time.NewTicker(interval)
+	func() {
```
stoader: What's the purpose of using an inner func. here?

PR #1405, opened 2018-11-20
bucket deletion error, if not empty

prekoa:


api/bucket.go
```diff
@@ -651,6 +651,14 @@ func errorResponseFrom(err error) *pkgCommon.ErrorResponse {
 		}
 	}
 
+	if err.Error() == "bucket: "+pkgErrors.ErrorBucketDeleteNotEmpty.Error() {
```
stoader: Simply check that the type of `err` is of `pkgErrors.ErrorBucketDeleteNotEmpty`

pkg/errors/errors.go
```diff
@@ -55,4 +55,5 @@ var (
 	ErrorAlibabaNodePoolFieldIsEmpty  = errors.New("At least one 'nodePool' is required.")
 	ErrorAlibabaNodePoolFieldLenError = errors.New("Only one 'nodePool' is supported.")
 	ErrorAlibabaMinNumberOfNodes      = errors.New("'num_of_nodes' must be greater than zero.")
+	ErrorBucketDeleteNotEmpty         = errors.New("failed to delete bucket, it was not empty")
```
stoader: I'd change the error message to `non empty buckets can not be deleted`

internal/providers/amazon/objectstore.go
```diff
@@ -242,7 +242,7 @@ func (s *objectStore) deleteFailed(bucket *ObjectStoreBucketModel, reason error)
 	if err := s.db.Save(bucket).Error; err != nil {
 		return emperror.WrapWith(err, "failed to delete bucket", "bucket", bucket.Name)
 	}
-	return nil
+	return emperror.WrapWith(reason, "bucket", bucket.Name)
```
lpuskas: this seems to be wrong here, if the bucket can't be deleted on the provider and the operation is not forced this code should not be reached, and the error returned by the provider is returned

pkg/providers/alibaba/objectstore/objectstore.go
```diff
@@ -115,6 +116,15 @@ func (o *objectStore) CheckBucket(bucketName string) error {
 
 // DeleteBucket deletes a bucket from the object store
 func (o *objectStore) DeleteBucket(bucketName string) error {
+	obj, err := o.ListObjects(bucketName)
+	if err != nil {
+		return emperror.With(emperror.Wrap(err, "could not list objects"), "bucket", bucketName)
```
lpuskas: use emperror's WrapWith insted; the woding should be "faieled to list objects" (to be consistent across providers)

pkg/providers/amazon/objectstore/objectstore.go
```diff
@@ -178,7 +179,15 @@ func (s *objectStore) DeleteBucket(bucketName string) error {
 		Bucket: aws.String(bucketName),
 	}
 
-	_, err := s.client.DeleteBucket(input)
+	obj, err := s.ListObjects(bucketName)
+	if err != nil {
```
lpuskas: use emperror's WrapWith insted; the woding should be "faieled to list objects" (to be consistent across providers)

pkg/providers/amazon/objectstore/objectstore.go
```diff
@@ -178,7 +179,15 @@ func (s *objectStore) DeleteBucket(bucketName string) error {
...
+		return emperror.With(emperror.Wrap(err, "could not list objects"), "bucket", bucketName)
+	}
+
+	if len(obj) > 0 {
+		return emperror.With(pkgErrors.ErrorBucketDeleteNotEmpty, "bucket", bucketName)
+	}
+	_, err = s.client.DeleteBucket(input)
 	if err != nil {
 		err = s.convertError(err)
 		return emperror.With(emperror.Wrap(err, "bucket deletion failed"), "bucket", bucketName)
```
lpuskas: use emperror's WrapWith instead; the woding should be "faieled to list objects" (to be consistent across providers)

internal/providers/amazon/objectstore.go
lpuskas: fixed - the original error is funneled through this method ...

pkg/errors/errors.go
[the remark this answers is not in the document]
lpuskas: let's let this as is for the time being, error handling is being refactored as part of other tasks

PR #1408, opened 2018-11-20
Cert generation

sagikazarmark:


cluster/hooks_ingress.go
```diff
@@ -0,0 +1,132 @@
...
+
+		defaultCertSecretRequest := &secret.CreateSecretRequest{
+			Name: defaultCertSecretName,
+			Type: pkgSecret.GenericSecret, // bypassing secret validation and generation
+			Values: map[string]string{
+				pkgSecret.ServerCert: string(cert),
+				pkgSecret.ServerKey:  string(key),
+			},
+			Tags: []string{
+				pkgSecret.TagBanzaiReadonly,
```
stoader: It might be useful to add another tag as well which indicates that this secret is a server cert or is dedicated to ingress controller if shouldn't be reused for other purposes.

pkg/crypto/cert/generator.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// NewGenerator returns a new Generator instance.
+func NewGenerator(caLoader CALoader) *Generator {
+	return &Generator{
+		caLoader: caLoader,
+	}
+}
+
+// GenerateServerCertificate generates a self-signed cert-key pair for server usage.
```
orymate:
```suggestion
// GenerateServerCertificate generates a cert-key pair for server usage issued by the loaded CA.
```

pkg/crypto/cert/generator.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/banzaicloud/bank-vaults/pkg/tls"
+	"github.com/pkg/errors"
+)
+
+const (
+	validity = 365 * 24 * time.Hour
+)
+
+// Generator generates a self-signed cert-key pair.
```
orymate:
```suggestion
// Generator creates a cert-key pair issued by the loaded CA.
```

cluster/hooks_ingress.go
sagikazarmark: We decided to use TLS secret type.

PR #1409, opened 2018-11-20
[WIP] New anchore version

pbalogh-sa:


api/security.go
```diff
@@ -289,7 +296,14 @@ func CreatePolicy(c *gin.Context) {
 	if !ok {
 		return
 	}
-	response, err := anchore.MakeAnchoreRequest(commonCluster.GetOrganizationId(), commonCluster.GetUID(), http.MethodPost, "policies", policyBundle)
+	anchoreRequest := anchore.AnchoreRequest{
+		OrgID:     commonCluster.GetOrganizationId(),
+		ClusterID: commonCluster.GetUID(),
+		Method:    http.MethodPost,
+		URL:       "policies",
```
stoader: It may worth factoring out `policies` into a const in case it has to be replaced in the future.

internal/security/anchore.go
```diff
@@ -43,104 +44,172 @@ type AnchoreError struct {
...
+	anchoreRequest := AnchoreRequest{
+		AdminUser: true,
+		Method:    http.MethodPost,
+		URL:       "accounts",
+		Body:      anchoreAccount,
+	}
+	_, err := MakeAnchoreRequest(anchoreRequest)
 	if err != nil {
-		logger.Panic("invalid Anchore database config: ", err.Error())
+		return err
```
stoader: Use emperror.Wrap to collect stack trace and add error context

internal/security/anchore.go
```diff
@@ -43,104 +44,172 @@ type AnchoreError struct {
...
+	anchoreRequest := AnchoreRequest{
+		AdminUser: true,
+		Method:    http.MethodPost,
+		URL:       endPoint,
+		Body:      anchoreUser,
+	}
+	_, err := MakeAnchoreRequest(anchoreRequest)
 	if err != nil {
-		logger.Panic("failed to initialize db: ", err.Error())
+		return err
```
stoader: Use emperror.Wrap to collect stack trace and add error context

internal/security/anchore.go
```diff
@@ -204,39 +276,54 @@ func RemoveAnchoreUser(orgId uint, clusterId string) {
...
+	}
+	logger.Debugf("Anchore user %v deleted.", anchorUserName)
+	if deleteAnchoreAccount(anchorUserName) != http.StatusNoContent {
+		logger.Errorf("Error deleting Anchore account: %v", anchorUserName)
+	}
+	logger.Debugf("Anchore account %v deleting.", anchorUserName)
 }
 
-func MakeAnchoreRequest(orgId uint, clusterId string, method string, url string, body interface{}) (*http.Response, error) {
+// MakeAnchoreRequest do anchore api call
```
bonifaido: I would rename this to `DoAnchoreRequest`, Do* naming is some kind of convention in HTTP client libs.

internal/security/anchore.go
```diff
@@ -204,39 +276,54 @@ func RemoveAnchoreUser(orgId uint, clusterId string) {
...
+	if req.Body != nil {
 		var buf io.ReadWriter
 		buf = new(bytes.Buffer)
-		err := json.NewEncoder(buf).Encode(body)
+		err := json.NewEncoder(buf).Encode(req.Body)
 		if err != nil {
 			return nil, err
 		}
-		request, _ = http.NewRequest(method, AnchorEndpoint+"/"+path.Join("v1", url), buf)
+		request, err = http.NewRequest(req.Method, AnchoreEndpoint+"/"+path.Join("v1", req.URL), buf)
```
bonifaido: I would move the AnchoreEndpoint into the path.Join call

internal/security/anchore.go
```diff
@@ -317,13 +317,13 @@ func MakeAnchoreRequest(req AnchoreRequest) (*http.Response, error) {
 		if err != nil {
 			return nil, err
 		}
-		request, err = http.NewRequest(req.Method, AnchoreEndpoint+"/"+path.Join("v1", req.URL), buf)
+		request, err = http.NewRequest(req.Method, path.Join(AnchoreEndpoint+"/v1", req.URL), buf)
```
stoader: `path.Join(AnchoreEndpoint, "v1", req.URL)`

internal/security/anchore.go
```diff
@@ -317,13 +317,13 @@ func MakeAnchoreRequest(req AnchoreRequest) (*http.Response, error) {
...
 		if err != nil {
 			return nil, err
 		}
 	} else {
 		var err error
-		request, err = http.NewRequest(req.Method, AnchoreEndpoint+"/"+path.Join("v1", req.URL), nil)
+		request, err = http.NewRequest(req.Method, path.Join(AnchoreEndpoint+"/v1", req.URL), nil)
```
stoader: `path.Join(AnchoreEndpoint, "v1", req.URL)`

PR #1412, opened 2018-11-21
cluster: allow non-force delete of incomplete clusters

orymate:
if the kubeconfig is not available, then we were not able to create
resources on the cluster, so we don't have to remove them. this makes it
possible to delete incomplete clusters on a controlled way (get a
feedback if the deletion of the cloud resources did not succeed)

~~I would like to get review about the idea before testing the proposed solution.~~

Tested with gke.


tarokkk:
In my opinion this could help on a specific edge. I can't recall any way where we delete kubeconfig before cluster delete. 👍

PR #1414, opened 2018-11-21
Use bank-vaults configure for configuring Vault in docker-compose env

bonifaido:


sagikazarmark:
This is going to be integrated in  #1424

PR #1421, opened 2018-11-21
Fix NaN resource calculation

sagikazarmark:


orymate:
handle +/-inf as well

PR #1432, opened 2018-11-22
Test parallelism

sagikazarmark:


sagikazarmark:
For reference, here is the last build:

https://circleci.com/workflow-run/b4c264c4-3409-48a4-96d2-7e6624159cd0

This is already faster than the sequential one.

I already have a few ideas to utilize Makefile for tests and make the config better, but I have to test those first.

sagikazarmark:
We decided to delay this until we actually need it.

Issue #1428, opened 2018-11-22
Implement securityScan is-enabled API.


pbalogh-sa:
Security scan can be enabled or disabled in pipeline, so it should be queryable.

PR #1472, opened 2018-11-29
Is security scan enabled globally API

pbalogh-sa:
Add API endpoint to check whether security scan is enabled in ENV.
fixes: #1428

### DIFF
--- a/api/security.go
+++ b/api/security.go
@@ -536,9 +536,13 @@ func GetReleaseScanLog(c *gin.Context) (map[string]bool, bool) {
 func SecurytiScanEnabled(c *gin.Context) {
 
 	if viper.GetBool("anchore.enabled") {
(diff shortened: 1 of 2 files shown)


api/security.go
```diff
@@ -530,3 +531,14 @@ func GetReleaseScanLog(c *gin.Context) (map[string]bool, bool) {
...
 }
+
+// SecurytiScanEnabled checks if security scan is enabled in pipeline
+func SecurytiScanEnabled(c *gin.Context) {
+
+	if viper.GetBool("anchore.enabled") {
+		c.Status(http.StatusOK)
+		return
+	}
+	c.Status(http.StatusBadRequest)
```
bonifaido: I would return some other 200 always and the body should contain the boolean I guess, because this is not a client error.

PR #1497, opened 2018-12-04
Spotguide refactor & sync & introduce pipeline supported versions for releases

tothandras:

PR #1498, opened 2018-12-04
[wip] spotguide: fix GetSpotguide

orymate:
GetSpotguideIcon paniced if no result found

### DIFF
--- a/spotguide/spotguide.go
+++ b/spotguide/spotguide.go
@@ -286,22 +286,23 @@ func GetSpotguides(orgID uint) ([]*SpotguideRepo, error) {
 	return spotguides, err
 }
 
-func GetSpotguide(orgID uint, name, version string) ([]SpotguideRepo, error) {
+func GetSpotguide(orgID uint, name, version string) (*SpotguideRepo, error) {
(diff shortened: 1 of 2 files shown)


spotguide/spotguide.go
```diff
@@ -286,23 +286,24 @@ func GetSpotguides(orgID uint) ([]*SpotguideRepo, error) {
...
 
-func GetSpotguide(orgID uint, name, version string) ([]SpotguideRepo, error) {
+func GetSpotguide(orgID uint, name, version string) (*SpotguideRepo, error) {
 	db := config.DB()
 	where := SpotguideRepo{OrganizationID: orgID, Name: name, Version: version}
-	repo := []SpotguideRepo{}
-	err := db.Find(&repo, where).Error
-	return repo, err
+	var repo *SpotguideRepo
+	err := db.Where(where).First(&repo).Error
```
stoader: Don't you need to pass only repo instead &repo since it's already a pointer?

orymate:
Have to check, but it cant be nil then

orymate:
#1497 will handle this

PR #1510, opened 2018-12-06
ARK fixes

waynz0r:
* delete stale ark deployments on cluster delete
* only sync backups and restores for cluster in RUNNING or WARNING state

### DIFF
--- a/internal/ark/events/cluster_event_handler.go
+++ b/internal/ark/events/cluster_event_handler.go
@@ -47,8 +47,8 @@ func NewClusterEventHandler(events clusterEvents, db *gorm.DB, logger logrus.Fie
 // RemoveStaleDeployments deletes stale ARK deployment records from database
 func (eh *ClusterEventHandler) DeleteStaleARKDeployments(orgID uint) error {
 	var deployments []*ark.ClusterBackupDeploymentsModel
-
-	eh.logger.WithField("org", orgID).Debug("removing stale ark deployment records")
+	log := eh.logger.WithField("org", orgID)
+	log.Debug("removing stale ark deployment records")
 
 	err := eh.db.Where(ark.ClusterBackupDeploymentsModel{OrganizationID: orgID}).Preload("Cluster").Find(&deployments).Error
 	if err != nil {
@@ -59,7 +59,7 @@ func (eh *ClusterEventHandler) DeleteStaleARKDeployments(orgID uint) error {
 		if deployment.ID > 0 && deployment.Cluster.ID == 0 {
 			err = eh.db.Delete(&deployment).Error
 			if err != nil {
-				eh.logger.Error(emperror.Wrap(err, "could not delete deployment record"))
+				log.Error(emperror.Wrap(err, "could not delete deployment record"))
 			}
 		}
 	}
--- a/internal/ark/events/cluster_event_handler.go
+++ b/internal/ark/events/cluster_event_handler.go
@@ -1,0 +1,68 @@
+// Copyright © 2018 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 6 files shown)


internal/ark/events/cluster_event_handler.go
```diff
@@ -0,0 +1,68 @@
...
+	err := eh.db.Where(ark.ClusterBackupDeploymentsModel{OrganizationID: orgID}).Preload("Cluster").Find(&deployments).Error
+	if err != nil {
+		return err
+	}
+
+	for _, deployment := range deployments {
+		if deployment.ID > 0 && deployment.Cluster.ID == 0 {
+			err = eh.db.Delete(&deployment).Error
+			if err != nil {
+				eh.logger.Error(emperror.Wrap(err, "could not delete deployment record"))
```
stoader: shouldn't the previously created logger be used here in order to have the org id in the log message?
```
log := eh.logger.WithField("org", orgID)
log.Debug("removing stale ark deployment records")
...
...
log.Error(emperror.Wrap(err, "could not delete deployment record"))
```

internal/ark/events/cluster_events.go
```diff
@@ -0,0 +1,49 @@
...
+type eventBus interface {
+	SubscribeAsync(topic string, fn interface{}, transactional bool) error
+}
+
+type clusterEventBus struct {
+	eb eventBus
+}
+
+const (
+	clusterCreatedTopic = "cluster_created"
```
stoader: Instead of introducing constants here for `cluster_created` and `cluster_deleted` can't we just create some global costs for these and use everywhere? There are constants for these event literals in `internal/monitor/events.go`, `cluster/events.go`. (see `OrganizationRegisteredTopic`)

internal/ark/events/cluster_event_handler.go
waynz0r: sure, good catch!

internal/ark/events/cluster_events.go
waynz0r: I had a conversation about this with @sagikazarmark yesterday. The main reason is that these are separate consumers and as such is better to have separate interfaces for each, this gives us the ability to move those consumers out from pipeline easier. Surely Mark can explain it more clearly.

sagikazarmark: Please remove the word "global" from your dictionaries. 🙂

internal/ark/sync/backups_sync_svc.go
```diff
@@ -70,8 +72,19 @@ func (s *BackupsSyncService) SyncBackups(clusterManager *cluster.Manager) error
 	// iterate through clusters and sync backups for each of them
 	for _, cluster := range clusters {
 		log := s.logger.WithField("clusterID", cluster.GetID())
+
+		status, err := cluster.GetStatus()
+		if err != nil {
+			log.Error(emperror.Wrap(err, "could not get cluster status"))
```
sagikazarmark: Use an error handler here

internal/ark/sync/restores_sync_svc.go
```diff
@@ -60,9 +61,21 @@ func (s *RestoresSyncService) SyncRestores(clusterManager *cluster.Manager) erro
 	}
 
 	for _, cluster := range clusters {
-		err := s.syncRestoresForCluster(cluster)
+		log := s.logger.WithField("clusterID", cluster.GetID())
+
+		status, err := cluster.GetStatus()
+		if err != nil {
+			log.Error(emperror.Wrap(err, "could not get cluster status"))
```
sagikazarmark: Use an error handler here.

internal/ark/sync/backups_sync_svc.go
waynz0r: Thanks @sagikazarmark , I'll fix this in an upcoming PR!

internal/ark/sync/restores_sync_svc.go
waynz0r: Thanks @sagikazarmark , I'll fix this in an upcoming PR!

PR #1514, opened 2018-12-07
Add location information to alibaba bucket listing

baluchicken:


internal/providers/alibaba/objectstore.go
```diff
@@ -194,8 +195,14 @@ func (os *objectStore) ListBuckets() ([]*objectstore.BucketInfo, error) {
...
 		}
-
-		bucketList = append(bucketList, bucketInfo)
+		if location, err := os.objectStore.GetLocation(bucket); err == nil {
+			// Removing oss- from the beginning of the location because we are storing and using it
+			// without oss- prefix
+			bucketInfo.Location = strings.TrimPrefix(location, "oss-")
+			bucketList = append(bucketList, bucketInfo)
+		} else {
+			return nil, err
```
stoader: Use emperror.Wrap to provider context and preserve call stack

baluchicken: I already did this in the s.objectStore.GetLocation function, the call stack is inside the error, maybe I can add additional context here, but I think the bucket name provided earlier is enough

PR #1515, opened 2018-12-07
gke: add preemtible flag to Profile API

sancyx:


sancyx:
Db column is already added this is just a small fix to actually set the field value in profile.

PR #1520, opened 2018-12-07
Extend cluster details

orymate:
All fields of the cluster status request (the one using only DB) are added to cluster details request (which calls out to the cluster).


api/cluster.go
```diff
@@ -372,7 +371,8 @@ func GetClusterDetails(c *gin.Context) {
 	if ok != true {
 		return
 	}
-	log.Info("getting cluster details")
+	log.Debugf("getting cluster details for %v", commonCluster)
```
sagikazarmark: Use structured logging.

api/cluster.go
```diff
@@ -384,14 +384,24 @@ func GetClusterDetails(c *gin.Context) {
 		return
 	}
 
-	log.Info("Start getting API endpoint")
+	status, err := commonCluster.GetStatus()
+	if err != nil {
+		log.Errorf("Error during getting status: %s", err.Error())
```
sagikazarmark: Use error handler

api/cluster.go
```diff
@@ -384,14 +384,24 @@ func GetClusterDetails(c *gin.Context) {
...
+		c.JSON(http.StatusBadRequest, pkgCommon.ErrorResponse{
+			Code:    http.StatusBadRequest,
```
tothandras: `NotFound` might be better here

api/cluster.go
orymate: #1523

PR #1549, opened 2018-12-13
Enable nodegroups with 0 node count & nodepool fetch/update endpoint

sancyx:


api/cluster.go
```diff
@@ -413,6 +414,87 @@ func GetClusterDetails(c *gin.Context) {
...
+	}
+
+	c.JSON(http.StatusOK, response)
+}
+
+func getActualNodeCounts(commonCluster cluster.CommonCluster) map[string]int {
+	nodePoolCounts := make(map[string]int)
+	kubeConfig, err := commonCluster.GetK8sConfig()
+	if err != nil {
+		return nodePoolCounts
```
stoader: Is it fine to return and empty collection instead of returning the error? If yes than we should log it with the appropriate context.

PR #1557, opened 2018-12-17
Distribution

Ecsy:


cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+func (c *EC2ClusterBanzaiCloudDistribution) GetSshSecretId() string {
+	return c.model.Cluster.SSHSecretID
+}
+
+func (c *EC2ClusterBanzaiCloudDistribution) SaveSshSecretId(sshSecretId string) error {
+	c.model.Cluster.SSHSecretID = sshSecretId
+
+	err := c.db.Save(&c.model).Error
+	if err != nil {
+		return errors.Wrap(err, "failed to save ssh secret id")
```
baluchicken: Because the logger lacks the context in this package please use emperor.Wrapwith and add context to the error.

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+}
+
+func convertTaints(taints banzaicloud.Taints) (result banzaicloudDB.Taints) {
+	for _, taint := range taints {
+		result = append(result, banzaicloudDB.Taint(taint))
+	}
+	return
+}
+
+func createEC2BanzaiCloudNetworkFromRequest(network banzaicloud.Network, userId uint) (banzaicloudDB.Network, error) {
```
baluchicken: Do not return error with this function because it is always nil

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+	}
+
+	c := &EC2ClusterBanzaiCloudDistribution{
+		db:    db,
+		model: &m,
+	}
+	return c, nil
+}
+
+func createEC2BanzaiCloudNodePoolsFromRequest(pools banzaicloud.NodePools, userId uint) (banzaicloudDB.NodePools, error) {
```
baluchicken: Do not return error in this function because it is always nil

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+	}
+	n.CreatedBy = userId
+	return n, nil
+}
+
+func convertNetworkProvider(provider banzaicloud.NetworkProvider) (result banzaicloudDB.NetworkProvider) {
+	return banzaicloudDB.NetworkProvider(provider)
+}
+
+func createEC2BanzaiCloudKubernetesFromRequest(kubernetes banzaicloud.Kubernetes, userId uint) (banzaicloudDB.Kubernetes, error) {
```
baluchicken: Do not return error in this function because it is always nil

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+	k := banzaicloudDB.Kubernetes{
+		Version: kubernetes.Version,
+		RBAC:    banzaicloudDB.RBAC{Enabled: kubernetes.RBAC.Enabled},
+	}
+	k.CreatedBy = userId
+	return k, nil
+}
+
+func createEC2BanzaiCloudKubeADMFromRequest(kubernetes banzaicloud.KubeADM, userId uint) (banzaicloudDB.KubeADM, error) {
```
baluchicken: Do not return error in this function because it is always nil

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+
+func convertExtraArgs(extraArgs banzaicloud.ExtraArgs) banzaicloudDB.ExtraArgs {
+	res := make(banzaicloudDB.ExtraArgs, len(extraArgs))
+	for k, v := range extraArgs {
+		res[k] = banzaicloudDB.ExtraArg(v)
+	}
+	return res
+}
+
+func createEC2BanzaiCloudCRIFromRequest(cri banzaicloud.CRI, userId uint) (banzaicloudDB.CRI, error) {
```
baluchicken: Do not return error in this function because it is always nil

cluster/ec2_banzaicloud.go
```diff
@@ -0,0 +1,504 @@
...
+		RuntimeConfig: cri.RuntimeConfig,
+	}
+	c.CreatedBy = userId
+	return c, nil
+}
+
+func getMasterInstanceTypeAndImageFromNodePools(nodepools banzaicloudDB.NodePools) (masterInstanceType string, masterImage string, err error) {
+	for _, nodepool := range nodepools {
+		for _, role := range nodepool.Roles {
+			if banzaicloudDB.RoleMaster == role {
```
baluchicken: Inside this if there is a dead code, banzaicloudDB.RoleMaster is a constant and the value is "master"

PR #1559, opened 2018-12-17
[WIP] Add a transparent cache layer to the helm client

sagikazarmark:
Unfortunately this does not speed things up as much as we hoped. I already investigated this a lot, but I guess I will have to do some profiling to see where we lose performance.

### DIFF
--- a/pkg/helm/cache.go
+++ b/pkg/helm/cache.go
@@ -1,0 +1,81 @@
+// Copyright © 2018 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


pkg/helm/cache.go
```diff
@@ -0,0 +1,81 @@
...
+	rls "k8s.io/helm/pkg/proto/hapi/services"
+)
+
+var globalCache *cache.Cache
+
+func init() {
+	globalCache = cache.New(1*time.Minute, 5*time.Minute)
+}
+
+type releaseCache interface {
```
bonifaido: If this is going to be a general releaseCache interface for future cache implementations, maybe error handling should be added (I'm not sure though), or just in case of an error we should return cache-miss (?).

bonifaido:
Some background reading: https://github.com/helm/helm/issues/3500

sagikazarmark:
This should be reimplemented if necessary, once helm deployments are refactored.

PR #1561, opened 2018-12-20
Partial details

sagikazarmark:
⚠️ DO NOT SQUASH COMMITS IF YOU MERGE THIS PR ⚠️


api/common_stuff.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// listPods returns list of pods in all namespaces.
+func listPods(client *kubernetes.Clientset, fieldSelector string, labelSelector string) ([]v1.Pod, error) {
```
waynz0r: Is this func necessary? It looks out of place (hence common), we calls this from 3 places and we don't use any selectors at all at 2 of those. I suggest we just use `client.CoreV1().Pods("").List()` where it is needed atm.

sagikazarmark: This isn't a new function, I just moved it to a common place. I wanted to call it `shit.go`, but that's not PC. You are right, but this function should be destroyed in a subsequent PR.

internal/cluster/resourcesummary/summary.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// GetTotalSummary calculates all resource summary for a cluster.
+func GetTotalSummary(client kubernetes.Interface) (*Summary, error) {
+	nodeList, err := client.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list nodes")
+	}
+
+	podList, err := client.CoreV1().Pods("").List(metav1.ListOptions{})
```
baluchicken: just a minor pls use instead of "" -> metav1.NamespaceAll

PR #1575, opened 2018-12-31
Fixing possible nil pointer exceptions

waynz0r:


pkg/providers/amazon/autoscaling/autoscaling.go
```diff
@@ -97,6 +104,10 @@ func (group *Group) getSpotRequests() ([]*ec2.SpotInstanceRequest, error) {
 		return nil, err
 	}
 
+	if lc == nil {
+		return nil, nil
```
stoader: I think it would be useful to log that no launch config is available thus we can not retrieve spot requests.

stoader: I'd log in the ```
if res.LaunchSpecification == nil || res.LaunchSpecification.IamInstanceProfile == nil || res.LaunchSpecification.IamInstanceProfile.Name == nil || lc.IamInstanceProfile == nil {
    continue
}```
section due to which missing field is a spot instance request being ignored.

waynz0r: If that condition is true that means that the spot request is not related for the ASG as far as we can tell. I would not treat it as an error.

PR #1582, opened 2019-01-03
Local dev setup enhancements

Laci21:
On Ubuntu 16.04, the local pipeline setup was not working instantly based on the instructions of `developer.md`. 

- The first issue was that the generated `~/.vault-token` file was owned by root on my machine instead of my local user and it caused an error when starting `pipeline`. With the help of @martonsereg and @bonifaido we have identified that the error might be caused by the fact that Docker handles users differently on Linux and Mac. On Mac, the local user is mapped to Docker automatically, but on Linux we need to do this manually. The first commit of this PR aims to start Docker containers with the local user so that at the end the generated and then copied `~/.vault-token` file will be owned by the local user and hence can be accessed from `pipeline`.
**Please, test that this solution also works on Mac, before approving!**

- In the second commit, I added a few hints which I think would be useful to have in the `developer.md` file for a newcomer to get started with Pipelines.


docs/developer.md
```diff
@@ -80,6 +80,24 @@ For accessing Vault the `VAULT_ADDR` env var has to be set, Pipeline stores JWT
...
+```
+
+If you happen to get an error similar to this on the first run:
+```
+Error 1146: Table 'pipeline.amazon_eks_profiles' doesn't exist
+```
+
+You should set `autoMigrateEnabled = true` in the database section in the `config/config.toml` file.
+
+You should now be able to login on the Pipeline UI: http://localhost:4200/ui
```
siliconbrain:
```diff
- login
+ log in
```

docs/developer.md
```diff
@@ -48,9 +48,9 @@ As of now the example config enables OAuth2 based authentication. It can be chan
 
 OAuth2 based authentication requires GitHub application, this can be created by following this 
 [tutorial](https://developer.github.com/apps/building-oauth-apps/creating-an-oauth-app/).
-Please set the `clientid` and the `clientsecret` in the auth section, with the GitHub generated values.
+Please set the `token` in the github section and the `clientid` and the `clientsecret` in the auth section, with the GitHub generated values.
```
siliconbrain: There's no reference to any documentation explaining where this token comes from.

docs/developer.md
Laci21: fixed, thx!

docs/developer.md
Laci21: I added a short section about it to the `github-app.md` where I feel the description belongs to. Let me know if it makes sense.

docs/developer.md
```diff
@@ -80,6 +80,24 @@ For accessing Vault the `VAULT_ADDR` env var has to be set, Pipeline stores JWT
 export VAULT_ADDR=http://127.0.0.1:8200
 ```
 
+#### Start Pipeline
+
+Once you have the docker containers running for the development environment, you should be able to start the pipeline platform.
+
+You can install and then run it with:
+```bash
+$ go install ./cmd/pipeline/
```
sagikazarmark: This should rather be `make build` and `build/pipeline`. (Event better: mention that both VSCode and Goland has builtin tasks for this)

docker-compose.override.yml.dist
```diff
@@ -2,6 +2,7 @@ version: "3.1"
 
 services:
     db:
+        user: "${uid}:${gid}"
```
sagikazarmark: How about this: comment these lines and in the `docker-compose.override.yml` target check if the os is linux. If yes, remove the comment and place the uid-gid pair of the current user there. That way the make file stays clean and the override file contain actual values, thus it works without make as well (with plain docker-compose commands).

I highly doubt multiple users would use the same workspace.

Makefile
```diff
@@ -55,18 +55,23 @@ docker-compose.override.yml: ## Create docker compose override file
 docker-compose.anchore.yml: ## Create docker compose override file with anchore
 	cp docker-compose.anchore.yml.dist docker-compose.override.yml
 
+create-docker-dirs: ## Create .docker directories with your user
```
sagikazarmark: Is this actually necessary? Isn't the user setting enough?

Makefile
```diff
@@ -38,8 +38,8 @@ up: vendor start config/config.toml ## Set up the development environment
 
 .PHONY: down
 down: clean ## Destroy the development environment
-	docker-compose down
-	rm -rf .docker/
+	uid=$(shell id -u) gid=$(shell id -g) docker-compose down
+	@ if [[ "$$OSTYPE" == "linux-gnu" ]]; then sudo rm -rf .docker/; else rm -rf .docker/; fi
```
sagikazarmark: Isn't this resolved by the user setting?

docs/developer.md
Laci21: Made the changes, thx!

docker-compose.override.yml.dist
Laci21: I really like the idea, thank you! I added the changes accordingly. I also added a comment stating `needed for Linux users` above the `# user: "${uid}:${gid}"` lines, just in case later someone would see these commented lines and would just want to delete them. Hopefully, with this comment, they'll think twice before deleting it.

Makefile
Laci21: In my experience, it is not enough. Even though I was using my own user inside some of the docker containers, the `.docker` directory (and folders under it) was owned by root after creation. This caused `permission denied` errors inside containers where my local Linux user was used. That's why I used this solution so that these dirs will be owned by my Linux user and hence can operate on their mounts inside the containers as well.

Makefile
Laci21: Unfortunately, it was not resolved for me which is not so nice. Even though I created the folders with my Linux user, the folders under `.docker/volumes/vault/file` were created by root. That's why I could not delete the whole folder without `sudo`.

 As far as I know, the `vault` container creates these files, so I've also tried to launch that container as well with my local user to solve this issue. In that case, the `vault` container exited with a `permission denied` error. That's why I simply solved the issue with `sudo rm`, but let me know if you have another idea!

bonifaido: Unexisting external mount points are created by the Docker daemon implicitly, which runs as root, so on Linux the implicitly created folders will be owned by root.

Makefile
sagikazarmark: I would probably add these commands to the start target, before the compose up part. This target will never be executed on it's own.

Also, a shorter form: `mkdir -p .docker/volumes/{mysql,vault/file,vault/keys}`

Makefile
sagikazarmark: But the mount points are created `make start`...anyway, I guess some processes (eg mysql) could run with different users, so we probably have limited options here.

Makefile
Laci21: This is needed for the `anchorestart` target as well on Linux, that's why created this as a new target to avoid code duplication. In this case, do you think it's reasonable like this? (I'll commit the shorter `mkdir` form, thx!)

Makefile
Laci21: @sagikazarmark I'm not sure I understand, are you suggesting some changes with your last comment? For me, at least, it seemed to work.

sagikazarmark: No-no, it's okay as it is now

Makefile
sagikazarmark: anchorestart will be removed soon, but I don't want to block this PR, so I will just simplify the make targets once I get there.

sagikazarmark:
Thx @Laci21

PR #1583, opened 2019-01-04
Save kubernetes version in case of ACK cluster create

baluchicken:
We need to store Kubernetes Version in case of Alibaba clusters to handle ClusterDetails similar to other providers.  #1561

### DIFF
--- a/cluster/acsk.go
+++ b/cluster/acsk.go
@@ -288,6 +288,7 @@ func (c *ACSKCluster) CreateCluster() error {
 	if !ok {
 		return errors.New("could not cast cluster create response")
 	}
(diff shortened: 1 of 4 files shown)


cluster/acsk.go
```diff
@@ -288,6 +288,7 @@ func (c *ACSKCluster) CreateCluster() error {
 	if !ok {
 		return errors.New("could not cast cluster create response")
 	}
+	c.modelCluster.ACSK.KubernetesVersion = castedValue.KubernetesVersion
```
sagikazarmark: Past tense of cast is cast as well

baluchicken: Going to add a PR about Alibaba Nodepools soon where I will address this, thanks:)

PR #1597, opened 2019-01-08
Schema diff test

sagikazarmark:
⚠️ DO NOT SQUASH COMMITS ⚠️


internal/providers/google/gke_cluster_model.go
```diff
@@ -85,7 +85,7 @@ type GKENodePoolModel struct {
 
 	Name             string `gorm:"unique_index:idx_cluster_id_name"`
 	Autoscaling      bool   `gorm:"default:false"`
-	Preemptible      bool
+	Preemptible      bool   `gorm:"default:0"`
```
orymate:
```suggestion
	Preemptible      bool   `gorm:"default:false"`
```

internal/providers/oracle/objectstore_model.go
```diff
@@ -36,7 +36,7 @@ type ObjectStoreBucketModel struct {
 
 	SecretRef string
 	Status    string
-	StatusMsg string
+	StatusMsg string `sql:"type:text;"`
```
orymate:
```suggestion
	StatusMsg string `gorm:"type:text;"`
```

sagikazarmark: Moved to #1600

internal/providers/google/gke_cluster_model.go
sagikazarmark: Thanks!

PR #1603, opened 2019-01-10
PKE: add spot price as a provider config property, add ondemand labels

sancyx:


cluster/ec2_banzaicloud.go
```diff
@@ -229,8 +230,13 @@ func (c *EC2ClusterBanzaiCloudDistribution) GetStatus() (*pkgCluster.GetClusterS
 		nodePools[np.Name] = &pkgCluster.NodePoolStatus{
 			Count:             len(np.Hosts),
 			InstanceType:      providerConfig.AutoScalingGroup.InstanceType,
+			SpotPrice:         providerConfig.AutoScalingGroup.SpotPrice,
 			CreatorBaseFields: *NewCreatorBaseFields(np.CreatedAt, np.CreatedBy),
 		}
+
+		if providerConfig.AutoScalingGroup.SpotPrice != "" && providerConfig.AutoScalingGroup.SpotPrice != "0" {
```
stoader: In case spot price is  expressed as "0.0" this check won't work correctly. If SpotPrice != "" than parse it to float and compare agains the float number.

PR #1614, opened 2019-01-11
Install plain value to secret

tothandras:
- [x] API docs
- [x] generate client


api/cluster_secret.go
```diff
@@ -74,13 +75,23 @@ func InstallSecretToCluster(c *gin.Context) {
...
+			Value:     spec.Value,
 		}
 	}
 
 	secretName := c.Param("secretName")
 
+	// Either spec is not defined (empty) or at least one spec is not a value
+	hasSourceSecret := len(secretRequest.Spec) == 0
+	for _, spec := range secretRequest.Spec {
+		if spec.Source != "" || (spec.SourceMap != nil && len(spec.SourceMap) != 0) {
```
sagikazarmark: `len` handles if the value is nil, no need to manually check it.

api/cluster_secret.go
```diff
@@ -74,13 +75,23 @@ func InstallSecretToCluster(c *gin.Context) {
...
 			Source:    spec.Source,
 			SourceMap: spec.SourceMap,
+			Value:     spec.Value,
 		}
 	}
 
 	secretName := c.Param("secretName")
 
+	// Either spec is not defined (empty) or at least one spec is not a value
+	hasSourceSecret := len(secretRequest.Spec) == 0
```
sagikazarmark: `needsSecret`?

cluster/secrets.go
```diff
@@ -164,25 +165,43 @@ func InstallSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string,
...
-	} else if err != nil {
-		return nil, emperror.With(errors.Wrap(err, "failed to get secret"), "secret", req.SourceSecretName)
-	}
-
 	kubeSecretRequest := intSecret.KubeSecretRequest{
 		Name:      secretName,
 		Namespace: req.Namespace,
-		Type:      secretItem.Type,
-		Values:    secretItem.Values,
+		Type:      "",
```
sagikazarmark: Use default empty value

cluster/secrets.go
```diff
@@ -164,25 +165,43 @@ func InstallSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string,
...
+		Values:    nil,
```
sagikazarmark: Use default empty value

cluster/secrets.go
```diff
@@ -164,25 +165,43 @@ func InstallSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string,
...
+
+	if req.SourceSecretName != "" {
+		secretItem, err := secret.Store.GetByName(orgID, req.SourceSecretName)
+		if err == secret.ErrSecretNotExists {
+			return nil, ErrSecretNotFound
+		} else if err != nil {
+			return nil, emperror.With(errors.Wrap(err, "failed to get secret"), "secret", req.SourceSecretName)
+		}
+
+		kubeSecretRequest = intSecret.KubeSecretRequest{
```
sagikazarmark: How about just setting the necessary values, instead of reinitializing the struct?

cluster/secrets.go
```diff
@@ -164,25 +165,43 @@ func InstallSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string,
...
+			Name:      secretName,
+			Namespace: req.Namespace,
+			Type:      secretItem.Type,
+			Values:    secretItem.Values,
+			Spec:      intSecret.KubeSecretSpec{},
```
sagikazarmark: We know the size of the map, right (`req.Spec`)? Maybe initialize it first time the struct is created?

cluster/secrets.go
```diff
@@ -225,11 +242,36 @@ func MergeSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string, re
...
 
-	secretItem, err := secret.Store.GetByName(orgID, req.SourceSecretName)
-	if err == secret.ErrSecretNotExists {
-		return nil, ErrSecretNotFound
-	} else if err != nil {
-		return nil, emperror.With(errors.Wrap(err, "failed to get secret"), "secret", req.SourceSecretName)
+	kubeSecretRequest := intSecret.KubeSecretRequest{
+		Name:      secretName,
+		Namespace: req.Namespace,
+		Type:      "",
```
sagikazarmark: Use default empty value

cluster/secrets.go
```diff
@@ -225,11 +242,36 @@ func MergeSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string, re
...
+		Values:    nil,
```
sagikazarmark: Use default empty value

cluster/secrets.go
```diff
@@ -225,11 +242,36 @@ func MergeSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string, re
...
+		Spec:      intSecret.KubeSecretSpec{},
```
sagikazarmark: Do we know the size, to preallocate the map?

cluster/secrets.go
```diff
@@ -225,11 +242,36 @@ func MergeSecretByK8SConfig(kubeConfig []byte, orgID uint, secretName string, re
...
+
+	if req.SourceSecretName != "" {
+		secretItem, err := secret.Store.GetByName(orgID, req.SourceSecretName)
+		if err == secret.ErrSecretNotExists {
+			return nil, ErrSecretNotFound
+		} else if err != nil {
+			return nil, emperror.With(errors.Wrap(err, "failed to get secret"), "secret", req.SourceSecretName)
+		}
+
+		kubeSecretRequest = intSecret.KubeSecretRequest{
```
sagikazarmark: Do not reinitialize the struct, if possible.

internal/secret/kubernetes.go
```diff
@@ -70,7 +71,9 @@ func CreateKubeSecret(req KubeSecretRequest) (v1.Secret, error) {
 		}
 	} else {
 		for key, specItem := range req.Spec {
-			if specItem.Source != "" { // Map one secret
+			if specItem.Value != "" { // Map value directly
```
sagikazarmark: I think the value should be the last checked item. Partly to maintain backward compatibility, partly because an empty value is/can be a valid value, but empty secret references cannot.

PR #1622, opened 2019-01-14
Fix backward compatibility by not dropping columns from DB

baluchicken:


model/cluster.go
```diff
@@ -79,18 +79,21 @@ type ClusterModel struct {
...
-	Delete          bool `gorm:"-"`
+	ID                           uint `gorm:"primary_key"`
+	CreatedAt                    time.Time
+	CreatedBy                    uint
+	ClusterID                    uint   `gorm:"unique_index:idx_cluster_id_name"`
+	Name                         string `gorm:"unique_index:idx_cluster_id_name"`
+	InstanceType                 string
+	DeprecatedSystemDiskCategory string `gorm:"column:system_disk_category"`
+	DeprecatedSystemDiskSize     int    `gorm:"column:system_disk_size"`
+	DeprecatedImage              string `gorm:"column:image"`
```
stoader: This should pointer to make GROM write `NULL` to the database if no value is provided for this field. The same is for `DeprecatedSystemDiskCategory` and `DeprecatedSystemDiskSize`

PR #1623, opened 2019-01-14
eks supports 1.11 k8s version, remove deprecated API call

prekoa:


cluster/eks.go
```diff
@@ -1224,18 +1224,18 @@ func ListEksRegions(orgId uint, secretId string) ([]string, error) {
 	// TODO revisit this later when https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/ starts supporting AmazonEKS
 
 	eksRegionIds := make([]string, 0)
-	for region := range pkgEks.DefaultImages {
+	for region := range pkgEks.DefaultImages[pkgEks.DefaultK8sVersion] {
```
stoader: Instead of `pkgEks.DefaultK8sVersion` shouldn't this rather be `r.Properties.CreateClusterEKS.Version` ?

PR #1628, opened 2019-01-15
Cross-platform VPC network query API

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?
This PR implements a cross-platform, VPC network related resource (network, subnetwork, route table) querying API, whose primary goal is to enable UIs to present options for the user when selecting existing resources for cluster creation. Currently Google and Amazon is supported as provider, but more are coming.

### Why?
There was no such cross-platform API available.

### Checklist


api/network.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/banzaicloud/pipeline/auth"
+	"github.com/banzaicloud/pipeline/internal/platform/gin/correlationid"
+	ginutils "github.com/banzaicloud/pipeline/internal/platform/gin/utils"
+	"github.com/banzaicloud/pipeline/internal/providers"
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
+
+// NetworkAPI implements network functions
+type NetworkAPI struct{}
```
sagikazarmark: Would be nice to inject dependencies here instead of accessing the global versions. Those should eventually go away.

siliconbrain: Which dependencies do you mean? Could you give an example? I only found the secret store as a global dependency, but that has a "private" type, so it's not trivial to create a reference to it.

sagikazarmark: I mean the logger, errorHandler, db connection dependencies.

siliconbrain: I added the logger, but couldn't find any more dependencies.

Issue #1579, opened 2019-01-02
Support adding labels to node pools


stoader:
Pipeline already automatically applies labels to nodes of a node pool during cluster creation such as:

- **nodepool.banzaicloud.io/name** -> `<name of the nodepool>`
- **node.banzaicloud.io/ondemand** -> `<true for ondemand nodes, false for preemptive nodes>` (used by our custom scheduler for ondemand/spot instance percentage deployments)

Pipeline API needs to support adding any custom labels to node pools in order to leverage the various node instance metadata provided by Cloud Info. The Pipeline will pull instance types, price info, regions etc from Cloud Info and set these as node labels. This way these instance metadata become available inside the K8s cluster (and be used by e.g. advanced scheduling scenarios).

The payload for specifying labels for node pools:
```
"nodePools": {
        ...
        "<name>": {
            ...
            "labels": {
                "label1": "value1",
                "label2", "value2"
            }
            ...
        },
```

Users should be able to specify node label during cluster creation/update they want to place on the nodes of a nodepool. These will be added to nodes in addition to what comes from Cloud Info.

If the user specifies the same label name as what Pipeline uses than the user’s settings will override the Pipeline provided value.

sagikazarmark:
> If the user specifies the same label name as what Pipeline uses than the user’s settings will override the Pipeline provided value.

Is this a good idea? Do we rely on those labels? If yes, we shouldn't allow the user to do that, should we?

waynz0r:
+1 for not a good idea to allow users to overwrite Pipeline labels

stoader:
Agree that Pipeline provided labels should not be overridable as Pipeline relies on those.

smartyjohn:
There does need to be some distinction between node pools, e.g. pipeline/system and worker node pools in particular. That is, labels added to nodes may only apply to one, some, or all node pools.  I believe it would be sufficient if each node pool's labels were managed separately.

PR #1638, opened 2019-01-16
Create GKE cluster with custom labels

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | partially #1579 

### What's in this PR?

- Make it possible to add custom labels for Kubernetes nodes from create GKE cluster API call
- Store the labels in DB
- Add a function to create foreign and referenced keys
- Add foreign keys for GKE specific DB tables
- Add the related SQL scripts
- Make sure delete from tables works with foreign keys
- Moved deletion of unused secrets from `DeleteFromDatabase()` as it is not a DB related method
- Modified Postman collection to show how to add labels from the create GKE cluster API
- Made possible to run `make generate-client` on Linux (earlier it failed for me)
- Added client generated code
- Modified test `TestCreateNodePoolsModelFromRequest` to use GKE labels

### Additional context

- Edge case: when a user tries to set a Pipeline specific label such as `nodepool.banzaicloud.io/name` or `node.banzaicloud.io/ondemand`. In this case they should recieve 400 response code as those labels are not allowed to be overwritten.

### Checklist


cluster/gke_node_pools.go
```diff
@@ -63,13 +71,19 @@ func createNodePoolsFromClusterModel(clusterModel *google.GKEClusterModel) ([]*g
 	for i := 0; i < nodePoolsCount; i++ {
 		nodePoolModel := clusterModel.NodePools[i]
 
+		labels := map[string]string{
+			pkgCommon.LabelKey:         nodePoolModel.Name,
+			pkgCommon.OnDemandLabelKey: strconv.FormatBool(!nodePoolModel.Preemptible),
+		}
+
+		for _, label := range nodePoolModel.Labels {
+			labels[label.Name] = label.Value
```
stoader: Return with error if user provided labels conflicts with Pipeline reserved labels

cluster/manager_delete.go
```diff
@@ -247,6 +248,17 @@ func deleteDnsRecordsOwnedByCluster(cluster CommonCluster) error {
 	return nil
 }
 
+func deleteUnusedSecrets(cluster CommonCluster) error {
```
stoader: Make this a method of `Manager` and use the logger that `Manager` holds.

cluster/manager_delete.go
```diff
@@ -352,6 +364,11 @@ func (m *Manager) deleteCluster(ctx context.Context, cluster CommonCluster, forc
 	// delete from proxy from kubeProxyCache if any
 	m.DeleteKubeProxy(cluster)
 
+	err = deleteUnusedSecrets(cluster)
```
stoader: This supersedes `ClusterModel.preDelete` thus `ClusterModel.preDelete` can be deleted.

cluster/gke_node_pools.go
Laci21: Isn't it enough to check earlier in the `CreateClusterGKE.Validate()` function, [here](https://github.com/Laci21/pipeline/blob/create-gke-cluster-with-custom-labels/pkg/cluster/gke/gke.go#L119)?
(I'll modify that implementation to check for some kind of Banzai Cloud or Pipeline specific label name, instead of checking all labels one by one, once that pattern is decided.)

cluster/manager_delete.go
Laci21: How about the solution I added in [this](https://github.com/banzaicloud/pipeline/pull/1638/commits/c7c2fb5c11448722be0adb857cee563aeea69603) commit? This way, this function uses the logger of `Manager` (as it gets it as a parameter), but I also kept the way how similar functions are implemented in this file, like e.g. [DeleteAllResources()](https://github.com/banzaicloud/pipeline/blob/c7c2fb5c11448722be0adb857cee563aeea69603/cluster/manager_delete.go#L72) or[ deleteDnsRecordsOwnedByCluster()](https://github.com/banzaicloud/pipeline/blob/c7c2fb5c11448722be0adb857cee563aeea69603/cluster/manager_delete.go#L233), so to keep it as a simple function in this file and not a method of `Manager`. Do you think this is reasonable like this?

cluster/manager_delete.go
Laci21: I think I have already deleted it in [this](https://github.com/banzaicloud/pipeline/pull/1638/commits/cb32a0b6f7feb5d35cba22d9e31e216312083adb) commit.

Makefile
```diff
@@ -95,7 +95,7 @@ endif
 
 .PHONY: docker-build
 docker-build: ## Builds go binary in docker image
-	docker run -it -v $(PWD):/go/src/${PACKAGE} -w /go/src/${PACKAGE} golang:${GOLANG_VERSION}-alpine go build -o pipeline_linux ${BUILD_PACKAGE}
+	docker run -it -v $$(PWD):/go/src/${PACKAGE} -w /go/src/${PACKAGE} golang:${GOLANG_VERSION}-alpine go build -o pipeline_linux ${BUILD_PACKAGE}
```
sagikazarmark:
```suggestion
	docker run -it -v $${PWD}:/go/src/${PACKAGE} -w /go/src/${PACKAGE} golang:${GOLANG_VERSION}-alpine go build -o pipeline_linux ${BUILD_PACKAGE}
```

internal/providers/google/gke_cluster_model.go
```diff
@@ -22,6 +22,13 @@ import (
 	"github.com/jinzhu/gorm"
 )
 
+// TableName constants
+const (
+	GKEClusterModelTableName  = "google_gke_clusters"
```
sagikazarmark: If not used outside of this package, these constants should not be exported.

internal/providers/google/model.go
```diff
@@ -42,5 +44,25 @@ func Migrate(db *gorm.DB, logger logrus.FieldLogger) error {
 		"table_names": strings.TrimSpace(tableNames),
 	}).Info("migrating provider tables")
 
-	return db.AutoMigrate(tables...).Error
+	err := db.AutoMigrate(tables...).Error
+	if err != nil {
+		return err
+	}
+
+	err = model.AddForeignKey(db, logger, &model.ClusterModel{}, &GKEClusterModel{}, "ClusterID")
```
sagikazarmark: IIRC foreign keys can be defined on structs as tags. Although visually not really appealing, it might be better to have all table definitions in a single place (the model).

pkg/errors/errors.go
```diff
@@ -26,40 +26,41 @@ var (
...
+	ErrorClusterNotReady                       = errors.New("Cluster not ready yet")
+	ErrorNilCluster                            = errors.New("<nil> cluster")
+	ErrorWrongKubernetesVersion                = errors.New("Wrong kubernetes version for master/nodes. The required minimum kubernetes version is 1.8.x ")
+	ErrorDifferentKubernetesVersion            = errors.New("Different kubernetes version for master and nodes")
+	ErrorLocationEmpty                         = errors.New("Location field is empty")
+	ErrorRequiredLocation                      = errors.New("location is required")
+	ErrorRequiredSecretId                      = errors.New("Secret id is required")
+	ErrorCloudInfoK8SNotSupported              = errors.New("Not supported key in case of amazon")
+	ErrorNodePoolNotProvided                   = errors.New("At least one 'nodepool' is required for creating or updating a cluster")
+	ErrorNodePoolLabelClashesWithPipelineLabel = errors.New("NodePool label cannot be set to Pipeline specific label")
```
sagikazarmark:
```suggestion
	ErrorNodePoolLabelClashesWithPipelineLabel = errors.New("NodePool label conflicts with Pipeline reserved label name")
```

internal/providers/google/model.go
stoader: GORM doesn't create FKs based on struct tags, it uses the "foreighkey" struct tag solely for tracking the dependencies between models.

Makefile
Laci21: Ah, I missed that, thank you!

internal/providers/google/gke_cluster_model.go
Laci21: Understood and fixed. Need to get to used to this naming convention :)

internal/providers/google/model.go
sagikazarmark: Sad :(

PR #1652, opened 2019-01-21
istio mtls configurable

martonsereg:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Mutual TLS can be enabled in the Istio posthook, by setting the property `"mtls": true`.

It sets the corresponding property in Helm, that:
1. creates a mesh-wide authentication policy that specifies that all services in the mesh will only accept TLS requests.
2. creates a `DestinationRule` that configures the client-side part, so all internal requests will include a client certificate.
3. creates another `DestinationRule` that disables `mtls` to the Kubernetes API server.

### Checklist

### DIFF
--- a/cluster/hooks_istio.go
+++ b/cluster/hooks_istio.go
@@ -43,17 +43,22 @@ func InstallServiceMesh(cluster CommonCluster, param cluster.PostHookParam) erro
 
 	log.Infof("istio params: %#v", params)
 
-	values := map[string]interface{}{}
+	values := map[string]interface{}{
+		"global": map[string]interface{}{
+			"mtls": map[string]interface{}{
+				"enabled": params.EnableMtls,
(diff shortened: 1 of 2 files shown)


cluster/hooks_istio.go
```diff
@@ -43,17 +43,22 @@ func InstallServiceMesh(cluster CommonCluster, param cluster.PostHookParam) erro
 
 	log.Infof("istio params: %#v", params)
 
-	values := map[string]interface{}{}
+	values := map[string]interface{}{
```
stoader: Can we make this a struct?

martonsereg: yep.

PR #1659, opened 2019-01-24
fix: save github ID as well

tothandras:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

`GithubID` has to be saved alongside of name when inserting shared Spotguide org.

### Why?

Several query in the application uses both `Name` and `GithubID` in the where clause.

### Checklist

### DIFF
--- a/cmd/pipeline/main.go
+++ b/cmd/pipeline/main.go
@@ -59,6 +59,7 @@ import (
 	gormadapter "github.com/casbin/gorm-adapter"
 	"github.com/gin-contrib/cors"
 	"github.com/gin-gonic/gin"
+	"github.com/google/go-github/github"
 	"github.com/goph/emperror"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


cmd/pipeline/main.go
```diff
@@ -277,9 +278,14 @@ func main() {
 	})
 
 	// insert shared organization to DB if not exists
-	sharedOrg := auth.Organization{Name: viper.GetString(config.SpotguideSharedLibraryGitHubOrganization)}
-	if err := db.Where(sharedOrg).FirstOrCreate(&sharedOrg).Error; err != nil {
+	sharedOrgName := viper.GetString(config.SpotguideSharedLibraryGitHubOrganization)
+	if org, _, err := github.NewClient(nil).Organizations.Get(context.Background(), sharedOrgName); err != nil {
 		log.Errorf("failed to create shared organization: %s", err.Error())
```
orymate:
```suggestion
		log.Errorf("failed to query shared Github organization: %s", err.Error())
```

stoader: If the shared spotguide insertion into db is required by Pipeline than we should exit with error if the insertion fails.

Issue #1026, opened 2018-09-12
Cluster name uniqueness within org is not enforced


orymate:
**Describe the bug**
You can create multiple clusters in an organization with the same name. This is both confusing and makes the cluster instance refer to the same resources if the provider is the same as well.

**Steps to reproduce the issue:**
Post the same cluster request twice

**Expected behavior**
http 409

**Additional context**
nullable deleted_at is part of the unique constraint, and null != null in sql.

sagikazarmark:
One possible solution would be dropping soft delete entirely and introducing some sort of archive for deleted clusters.

Another solution would be making the deleted_at field non-null (eg. zero date) and overriding the default gorm deleted_at hooks to check this value instead of null.

PR #1662, opened 2019-01-24
cluster: fix Clusters.Exists()

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Workarounds #1026

### What's in this PR?
Fix Clusters.Exists method's incorrect return expression

### Checklist


internal/cluster/clusters.go
```diff
@@ -42,7 +42,7 @@ func (c *Clusters) Exists(organizationID uint, name string) (bool, error) {
 		return false, errors.Wrap(err, "could not check cluster existence")
 	}
 
-	return existingCluster.ID == 0, nil
+	return existingCluster.ID != 0, nil
```
stoader: `existingCluster.ID > 0` I think it's safer

orymate: to be honest, I don't really know the intents of this check, but would think that it's checked against the default value of `uint` here.

PR #1664, opened 2019-01-24
ARK - Use RBAC for AKS as well

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Enables RBAC for ARK on AKS as well

### Checklist

### DIFF
--- a/internal/ark/config.go
+++ b/internal/ark/config.go
@@ -103,9 +103,10 @@ type ConfigRequest struct {
 }
 
 type clusterConfig struct {
-	Name     string
-	Provider string
-	Location string
+	Name        string
+	Provider    string
+	Location    string
+	RBACEnabled bool
 
 	azureClusterConfig
 }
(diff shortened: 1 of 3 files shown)


internal/ark/config.go
```diff
@@ -156,19 +156,14 @@ func (req ConfigRequest) Get() (values ValueOverrides, err error) {
...
-
 	return ValueOverrides{
 		Configuration: configuration{
 			PersistentVolumeProvider: pvp,
 			BackupStorageProvider:    bsp,
 			RestoreOnlyMode:          req.RestoreMode,
 		},
 		RBAC: rbac{
-			Create: rbacEnabled,
+			Create: true,
```
stoader: This should come from the cluster's RbacEnabled property.

waynz0r: 👍

PR #1668, opened 2019-01-25
Remove azure-aks-client dependency

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
The azure-aks-client was written as a temporary fix while the missing AKS service features are integrated into the official azure-sdk-for-go. The missing features have since been implemented, so azure-aks-client is no longer required, but implements several extensions and helper methods to those provided by the official SDK. These changes remove the azure-aks-client dependency, copying only necessary features from it. It also tries to simplify and improve the implementation of AKS related features.

### Why?
azure-aks-client shouldn't be used because it was only meant as a temporary solution.

### Checklist


cluster/aks.go
```diff
@@ -323,178 +492,141 @@ func (c *AKSCluster) GetStatus() (*pkgCluster.GetClusterStatusResponse, error) {
...
+	if err != nil {
+		return emperror.Wrap(err, "failed to retrieve AKS cluster")
 	}
 
-	return nil
-}
+	// AKS currently only supports a single node pool
+	// For more info, see: https://feedback.azure.com/forums/914020-azure-kubernetes-service-aks/suggestions/34917127-support-multiple-node-pool
+	if len(request.AKS.NodePools) > 1 {
+		return fmt.Errorf("Too many node pools: %d", len(request.AKS.NodePools))
```
stoader: This is actually not correct. AKS allows creating clusters with multiple nodepools. It doesn't allow adding new nodepools to or removing existing nodepools from a running cluster. Only the size (min, max, node count) of the current nodepools can be adjusted on a running cluster.

siliconbrain:
moved loosely related commits to #1700

PR #1669, opened 2019-01-25
Add cluster status history

### LINKED ISSUE #1619: Proper Cluster State History
**Describe the bug**
When the cluster has been deleted we have lost all of the previous states(RUNNING,ERROR...).

* create cluster
* delete cluster immediately

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1619

### What's in this PR?

This PR adds tracking for cluster status changes.

### Checklist

### DIFF
--- a/cluster/ec2_banzaicloud.go
+++ b/cluster/ec2_banzaicloud.go
@@ -161,13 +161,24 @@ func (c *EC2ClusterBanzaiCloudDistribution) Persist(string, string) error {
 }
 
 func (c *EC2ClusterBanzaiCloudDistribution) UpdateStatus(status, statusMessage string) error {
-	if c.model.Cluster.ID != 0 && c.model.Cluster.Status != status {
+	originalStatus := c.model.Cluster.Status
+	originalStatusMessage := c.model.Cluster.StatusMessage
+
+	c.model.Cluster.Status = status
+	c.model.Cluster.StatusMessage = statusMessage
+
+	err := c.db.Save(&c.model).Error
+	if err != nil {
+		return errors.Wrap(err, "failed to update status")
+	}
+
+	if originalStatus != status {
 		statusHistory := &cluster.StatusHistoryModel{
 			ClusterID:   c.model.Cluster.ID,
 			ClusterName: c.model.Cluster.Name,
 
-			FromStatus:        c.model.Cluster.Status,
-			FromStatusMessage: c.model.Cluster.StatusMessage,
+			FromStatus:        originalStatus,
+			FromStatusMessage: originalStatusMessage,
 			ToStatus:          status,
 			ToStatusMessage:   statusMessage,
 		}
@@ -178,13 +189,6 @@ func (c *EC2ClusterBanzaiCloudDistribution) UpdateStatus(status, statusMessage s
 		}
 	}
 
-	c.model.Cluster.Status = status
-	c.model.Cluster.StatusMessage = statusMessage
-
-	err := c.db.Save(&c.model).Error
-	if err != nil {
-		return errors.Wrap(err, "failed to update status")
-	}
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


orymate:
missing at least from `func (c *EC2ClusterBanzaiCloudDistribution) UpdateStatus(status, statusMessage string) error`

database/migrations/1548428537_create_cluster_status_history_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE `cluster_status_history` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `cluster_id` int(10) unsigned NOT NULL,
```
stoader: Do we want to keep cluster status history record beyond the lifetime of the cluster record? If not let's add a foreign key to `cluster_id`

cluster/ec2_banzaicloud.go
```diff
@@ -161,6 +161,23 @@ func (c *EC2ClusterBanzaiCloudDistribution) Persist(string, string) error {
...
+			ClusterID:   c.model.Cluster.ID,
+			ClusterName: c.model.Cluster.Name,
+
+			FromStatus:        c.model.Cluster.Status,
+			FromStatusMessage: c.model.Cluster.StatusMessage,
+			ToStatus:          status,
+			ToStatusMessage:   statusMessage,
+		}
+
+		err := c.db.Save(&statusHistory).Error
```
stoader: Can't this be saved via `err := c.db.Save(&c.model)` ? What we need is to have a field in Cluster model that collects status history model entries related to the cluster.

cluster/gke.go
```diff
@@ -1845,12 +1845,29 @@ func (c *GKECluster) getProjectId() (string, error) {
...
+			ClusterID:   c.model.Cluster.ID,
+			ClusterName: c.model.Cluster.Name,
+
+			FromStatus:        c.model.Cluster.Status,
+			FromStatusMessage: c.model.Cluster.StatusMessage,
+			ToStatus:          status,
+			ToStatusMessage:   statusMessage,
+		}
+
+		err := c.db.Save(&statusHistory).Error
```
stoader: Same as above

model/cluster.go
```diff
@@ -385,6 +386,25 @@ func (a *ACSKClusterModel) AfterUpdate(scope *gorm.Scope) error {
...
+
+			FromStatus:        cs.Status,
+			FromStatusMessage: cs.StatusMessage,
+			ToStatus:          status,
+			ToStatusMessage:   statusMessage,
+		}
+
+		db := config.DB()
+
+		err := db.Save(&statusHistory).Error
```
stoader: Same as above

database/migrations/1548428537_create_cluster_status_history_table.up.sql
sagikazarmark: I deliberately made the decision not to add foreign keys:

1. With the deleted at behavior clusters never get deleted from the clusters table
2. The purpose of this table is to give an overview of the cluster status changes (eg. ERROR->DELETING). If the cluster ever gets deleted, this information is lost.
3. This is a temporary solution, so I'm not afraid of the table getting large.

Please confirm if the purpose of this solution changed @stoader @ahma

sagikazarmark:
>  We also need to expose the status history through the API such as clients (UI) can query it.

@stoader This wasn't in the ticket and AFAIK it's purpose is not to provide information to the users, but to us. Also, it's a temporary solution until the cluster state machine is refactored.

If there is a change in the purpose of this solution please discuss it with @ahma so that we can create a solution targeted at the problem. This is a quick fix, so we can avoid using triggers.

cluster/ec2_banzaicloud.go
sagikazarmark: See my other comments.

PR #1670, opened 2019-01-26
fix masterAddress and Token properties

tarokkk:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Hotfix to make kubeadm compatible output

### Checklist

### DIFF
--- a/api/cluster.go
+++ b/api/cluster.go
@@ -41,6 +41,7 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"net/url"
 )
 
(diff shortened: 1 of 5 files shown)


api/cluster.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sort"
 	"strings"
 
+	"net/url"
```
Ecsy: Please pay attention to imports sorting. I know IDEs tend to mess this up. Please move net/url to the internal packages block.

orymate: fmt cares with ordering inside groups only. and the linter is useless in this case with its 50 error reports, none of which about the import grouping.

PR #1674, opened 2019-01-29
PKE bootstrap commands

### LINKED ISSUE #1673: Create user data script calling banzai-kube install

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | closes #1673

### What's in this PR?
This PR adds the capability of creating Pipeline API tokens for the PKE installer (or other cluster service users), and to generate bootstrap commands that can be directly used as user-data scripts or from the CLI.

* New API endpoint: GET clusters/x/pke/commands
* Commands made available to EC2/PKE CreateCluster method for upcoming work

Limitation: jwt token gives general org-wide privileges instead of cluster-only

### Checklist

- [x] Implementation tested (with at least one cloud provider) -- PKE GKE
- [x] Related Helm chart(s) updated (needed!)

### To Do
- [x] test with a managed provider
- [x] add external url config to chart
- [x] `git rebase -i --autosquash origin/master`


cluster/common.go
```diff
@@ -63,7 +63,7 @@ type CommonCluster interface {
 	DeleteFromDatabase() error
 
 	// Cluster management
-	CreateCluster() error
+	CreateCluster(*Manager) error
```
orymate: no

api/cluster_create.go
```diff
@@ -149,7 +149,7 @@ func (a *ClusterAPI) CreateCluster(
 		PostHooks:      postHooks,
 	}
 
-	creator := cluster.NewCommonClusterCreator(createClusterRequest, commonCluster)
+	creator := cluster.NewClusterCreator(createClusterRequest, commonCluster, a.tokenGenerator, a.externalBaseURL)
```
sagikazarmark: Let's create a separate function for the common and pke cluster creator and conditionally call them here.

orymate: why do you think that it's the api's responsibility to decide on how a specific type of clusters should be created?

sagikazarmark: It's not, but for now it's an acceptable compromise, until we figure out how to properly do this. Once we have a few more creators, having a single constructor will become unmaintainable. Refactoring a conditional logic is much easier.

PR #1679, opened 2019-01-29
storage account validation

prekoa:


internal/providers/azure/objectstore.go
```diff
@@ -164,6 +168,12 @@ func (s *ObjectStore) getLogger() logrus.FieldLogger {
 	})
 }
 
+func isValidStorageAccount(storageAccount string) bool {
```
lpuskas: reaname it to isValidStorageAccountName, or simply isValid and add godoc explayning what "valid" means eventually a link to the regexp or the description of a valid storage account name

prekoa: thx, fixed

internal/providers/azure/objectstore.go
```diff
@@ -164,6 +168,13 @@ func (s *ObjectStore) getLogger() logrus.FieldLogger {
 	})
 }
 
+// https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-storage-account-name-errors
+func isValidStorageAccountName(storageAccount string) bool {
+	match, _ := regexp.MatchString(`^[a-z0-9]{3,24}$`, storageAccount)
```
stoader: This regex could be made as a const.

prekoa: fixed, thx

PR #1699, opened 2019-02-01
doc: extend authentication docs

bonifaido:
Since Dex was introduced, the authentication flow has changed (basically it has opened up a lot of new opportunities, and resolved some old issues). This PR tries to extend the docs to cover this and adds new examples as well.


bonifaido:
Ready for review.

PR #1701, opened 2019-02-01
Configure bypass egress traffic on EKS for Istio posthook

Laci21:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | none

### What's in this PR?

- In struct `pkgCluster.Ipv4Cidrs` the fields which hold the IP ranges for Kubernetes pods and services were changed to type string slice instead of simply string, because they can contain multiple ranges (they do e.g. on EKS)
-  Added implementation to `GetK8sIpv4Cidrs()` on EKS to be able to access external services (by [including internal IP ranges](https://istio.io/docs/tasks/traffic-management/egress/#calling-external-services-directly) only)

### Why?

The functionality to configure the `BypassEgressTraffic` variable was only added for GKE in #1643.

### Additional context

- Launched an EKS cluster with **`BypassEgressTraffic` set to `False`** and `autoSidecarInjectNamespaces` set to `default`. Launced a simple pod in the `default` namespace for which the `istio-proxy` sidecar container was automatically launched as well. Entered the pod's shell and tried to run `wget google.hu`. The result was `404 Not Found`, as expected, because this is what Istio does by default, the `istio-proxy` sidecar container hijacked the request and did not allow it to access the external network by default.
- Then launched an EKS cluster with **`BypassEgressTraffic` set to `True`**. From the same container with the same `wget` call I was able to access the external network. 

### Checklist

### To Do

Still need to implement the `GetK8sIpv4Cidrs()` method for the other cloud providers (except for GKE and EKS).


cluster/eks.go
```diff
@@ -1215,8 +1215,20 @@ func (c *EKSCluster) GetConfigSecretId() string {
 }
 
 func (c *EKSCluster) GetK8sIpv4Cidrs() (*pkgCluster.Ipv4Cidrs, error) {
-	//TODO
-	return nil, errors.New("not implemented")
+	// On EKS the services IP range is chosen from the two possible ranges below
```
stoader: Add GoDoc for `GetK8sIpv4Cidrs`

cluster/eks.go
```diff
@@ -1215,8 +1215,20 @@ func (c *EKSCluster) GetConfigSecretId() string {
...
+	// source: https://forums.aws.amazon.com/thread.jspa?messageID=859958
+	eksServiceClusterIPRanges := []string{"172.20.0.0/16", "10.100.0.0/16"}
+
+	// On EKS the pods IP range is coming from the CIDR's of the subnets
+	var eksPodIPRanges []string
+	for _, subnet := range c.GetModel().EKS.Subnets {
```
stoader: If the cluster is created with default settings than the default CIDR defined in the CloudFormation template is used. In that case the subnet CIDR is not persisted to the database. Currently we persist to db only the CIDR passed in by user.

PR #1704, opened 2019-02-01
Check security_scan flag

### LINKED ISSUE #1703: Create proper error message ins case of disabled security scan.
**Describe the bug**
In case of disabled security scan would better to create proper error message when  get policies or initiate image scan.

Get policies on cluster where security scan isn't enabled.

**Expected behavior**
Create a clear error message in this case ("security scan isn't enabled")

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #1703 

### What's in this PR?

Checking security_scan flag before anchore requests

### Why?

It makes error messages more exact.

### Checklist

### To Do

### DIFF
--- a/api/imagescan.go
+++ b/api/imagescan.go
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"path"
 
+	"github.com/banzaicloud/pipeline/api/ark/common"
 	apiclient "github.com/banzaicloud/pipeline/client"
 	"github.com/banzaicloud/pipeline/internal/security"
 	pkgCommmon "github.com/banzaicloud/pipeline/pkg/common"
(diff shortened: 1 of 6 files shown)


api/imagescan.go
```diff
@@ -73,6 +73,16 @@ func ScanImages(c *gin.Context) {
 	if !ok {
 		return
 	}
+
+	if !commonCluster.GetSecurityScan() {
+		c.JSON(http.StatusNotFound, pkgCommmon.ErrorResponse{
```
stoader: Can't `common.ErrorResponseWithStatus` be used here to send back error response?

pbalogh-sa: Thanks for mentioning common.ErrorResponseWithStatus function.

matyix: thx @pbalogh-sa

api/imagescan.go
```diff
@@ -18,6 +18,7 @@ import (
 	"net/http"
 	"path"
 
+	"github.com/banzaicloud/pipeline/api/ark/common"
```
tarokkk: Is there any specific reason that we import this from api/ark/common and not refactor it to api/common?

stoader: Agree this should go into common.

pbalogh-sa: I'm on it

PR #1705, opened 2019-02-01
WIP Allow custom VPC for AKS 

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | BC only
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
An optional virtual subnetwork ID parameter is added for AKS cluster creation requests to allow specifying an existing VPC subnet to create the cluster in. Also, because in Azure resources are linked to resource groups instead of regions, the `region` parameter is not required when creating a cluster in Azure but the new `resourceGroup` parameter is.

### Why?
Allow specifying an existing VPC subnet when creating clusters in AKS.

### Checklist


internal/providers/azure/network.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func getCIDR(addressPrefixes []string) string {
+	return strings.Join(addressPrefixes, ", ")
+}
+
+func (ns *azureNetworkService) ListNetworks() ([]intNetwork.Network, error) {
+	rp, err := ns.client.List(context.TODO(), ns.resourceGroupName)
+	if err != nil {
+		return nil, err
```
stoader: Add context to the error.

internal/providers/azure/network.go
```diff
@@ -0,0 +1,152 @@
...
+		for _, vn := range rp.Values() {
+			res = append(res, &azureNetwork{
+				cidr: getCIDR(*vn.AddressSpace.AddressPrefixes),
+				id:   *vn.Name, // this is what we want as ID
+				name: *vn.Name,
+			})
+		}
+		err = rp.NextWithContext(context.TODO())
+		if err != nil {
+			return nil, err
```
stoader: Add context to the error.

internal/providers/azure/network.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (ns *azureNetworkService) ListRouteTables(networkID string) ([]intNetwork.RouteTable, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (ns *azureNetworkService) ListSubnets(networkID string) ([]intNetwork.Subnet, error) {
+	vn, err := ns.client.Get(context.TODO(), ns.resourceGroupName, networkID, "")
+	if err != nil {
+		return nil, err
```
stoader: Add context to the error.

internal/providers/azure/network.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/banzaicloud/pipeline/secret"
+
+	intNetwork "github.com/banzaicloud/pipeline/internal/network"
+	pkgAzure "github.com/banzaicloud/pipeline/pkg/providers/azure"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type azureNetwork struct {
+	cidr string
```
stoader: Why not make it `[]string` since its a list of cidrs?

cluster/aks.go
```diff
@@ -189,9 +197,10 @@ func (c *AKSCluster) CreateCluster() error {
 			count := int32(np.Count)
 			name := np.Name
 			profiles = append(profiles, containerservice.ManagedClusterAgentPoolProfile{
-				Name:   &name,
-				Count:  &count,
-				VMSize: containerservice.VMSizeTypes(np.NodeInstanceType),
+				Name:         &name,
+				Count:        &count,
+				VMSize:       containerservice.VMSizeTypes(np.NodeInstanceType),
+				VnetSubnetID: getVNetSubnetID(np),
```
stoader: Is it a valid scenario to pass in a subnet that belongs to a resource group other than the resource group the cluster is created in? If not than add validation to check that the resource group of the subnet matches `c.GetResourceGroupName()` which is used for creating the cluster.

model/cluster.go
```diff
@@ -185,6 +185,7 @@ type AKSNodePoolModel struct {
 	NodeMaxCount     int
 	Count            int
 	NodeInstanceType string
+	VNetSubnetID     string
```
stoader: Add migration sql scripts for the changes in the model.

internal/providers/azure/network.go
siliconbrain: If it was a `[]string` then either the getter would have to do the concatenation or the query interface would have to be changed to propagate the new return type (and that would break the API)

cluster/aks.go
siliconbrain: The Azure web interface only lists VNs from the same resource group, so I think no, the subnet must be in the same resource group as the cluster. Also, it has to be under the same subscription. But should we check that? Azure will check these for us if and when it needs to, I don't think we should duplicate the logic here. The UI already provides a way to select a subnet that should work, and users will always be able to shoot themselves in the foot if they really want to.

PR #1707, opened 2019-02-03
EKS: add ASG tags from cf template, swith off Cluster Autoscaler if Hollowtrees is enabled

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### Checklist

### To Do

### DIFF
--- a/pkg/cluster/eks/action/actions.go
+++ b/pkg/cluster/eks/action/actions.go
@@ -29,21 +29,22 @@ import (
 	"github.com/aws/aws-sdk-go/service/eks"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/iam"
-	"github.com/banzaicloud/pipeline/pkg/common"
-	"github.com/goph/emperror"
-	"github.com/pkg/errors"
-	"github.com/satori/go.uuid"
-	"github.com/sirupsen/logrus"
-
+	"github.com/banzaicloud/pipeline/config"
 	"github.com/banzaicloud/pipeline/model"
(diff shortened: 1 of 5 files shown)


pkg/cluster/eks/action/actions.go
```diff
@@ -800,15 +804,26 @@ func (a *CreateUpdateNodePoolStackAction) ExecuteAction(input interface{}) (outp
 				{Key: aws.String("pipeline-stack-type"), Value: aws.String("nodepool")},
 			}
 
-			if nodePool.Autoscaling {
```
stoader: What's the reason for moving the logic of setting this autoscaler related to tag down to the CF template?

sancyx: The collection of tags you can pass to stack call, are applied to all resources and if nothing else changes in input params, then tags are no updated.

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -283,6 +292,12 @@ Resources:
       - Key: !Sub 'kubernetes.io/cluster/${ClusterName}'
         Value: 'owned'
         PropagateAtLaunch: 'true'
+      - Key: !If [ AutoscalerEnabled, 'k8s.io/cluster-autoscaler/enabled', 'k8s.io/cluster-autoscaler/disabled' ]
```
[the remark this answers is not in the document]
sancyx: Unfortunately this doesn't seem to work in case of tags, stack create is failing with: `CREATE_FAILED Key is empty or not found in properties`

PR #1708, opened 2019-02-03
[WIP] Contribution

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
This PR contains a few changes required by the contribution blog post.


config/config.toml.example
```diff
@@ -25,8 +25,8 @@ user = "sparky"
 role = "pipeline"
 password = "sparky123"
 dbname = "pipeline"
-logging = true
-autoMigrateEnabled = false
+logging = false # Turn this to true to see all queries in the log
+autoMigrateEnabled = true # This should be false in production
```
kozmagabor: It should be false by default

sagikazarmark: It is false by default, this is just an example configuration file which is used to generate a development configuration. Locally it should be true IMHO.

kozmagabor: i see, thanks

PR #1719, opened 2019-02-05
WIP Oracle VCN query API

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
This PR implements the virtual network/subnetwork/route table querying API for Oracle.

### Why?
Allow API users to query OKE VCN resources to simplify creating clusters in existing VCNs.

### Checklist

### DIFF
--- a/internal/providers/oracle/network.go
+++ b/internal/providers/oracle/network.go
@@ -1,0 +1,157 @@
+// Copyright © 2019 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


internal/providers/oracle/network.go
```diff
@@ -0,0 +1,157 @@
...
+	for idx, item := range rts {
+		routeTables[idx] = &oracleRouteTable{
+			id:   deref(item.Id),
+			name: deref(item.DisplayName),
+		}
+	}
+	return routeTables, nil
+}
+
+func deref(p *string) string {
```
stoader: Should this be moved to some common package such as it can be used in other places as well?

siliconbrain: It could be, but then it'd have to be renamed to something like `deref_string_or_empty` to avoid confusion and collisions and that wouldn't be so succinct anymore.

PR #1747, opened 2019-02-13
openapi: support nullable attribute with temporary generator image

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/OpenAPITools/openapi-generator/pull/1869

### What's in this PR?
The OpenAPI specification gives support for nullable fields via the [nullable attribute](https://swagger.io/docs/specification/data-models/data-types/#nullable) but this feature has not landed yet in the official generator CLI, thus I have updated the image built from the PR branch until the feature arrives.

### Why?
It's not possible to use pointer fields for example `*time.Time` is not achievable, however null and time.Time = 0 is very different.

### Additional Context
`oneOf` was broken (made no effect on the generated code, it still generated `map[string]interface{}`) previously as well, but now it becomes broken (the hard way) since it references a type called `OneOfAorBorC` but that actual type is not generated (results in compile time error). For the time being, I have commented out the broken `oneOf` parts from the API specification.

### Checklist


ackintosh:
FYI:  https://github.com/OpenAPITools/openapi-generator/pull/1869 has been merged into master. 😉

sagikazarmark:
Thanks for the heads up @ackintosh !

PR #1770, opened 2019-02-19
[WIP] PKE cluster create API v2

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | BC only
| Deprecations?   | yes
| Related tickets | -

### What's in this PR?
This PR implements part of the new cluster API (creation requests) for the PKE-on-AWS cluster type as an alternative to the legacy API. (Both work right now, the legacy API will be phased out eventually.)

### Why?
The current API is coupled too deeply with the business logic and data layers, which is bad. We should move toward a new API, building on a better architecture. 

### Checklist

### DIFF
--- a/api/cluster/requests.go
+++ b/api/cluster/requests.go
@@ -1,0 +1,338 @@
+// Copyright © 2019 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


siliconbrain:
Instead of doing this for PKE only, the new API will be implemented in parallel on another branch.

PR #1772, opened 2019-02-20
Pkecadence refactor

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR contains generic improvements to the PKE distribution workflows.

There are three main improvements in this PR:

- Introduced cluster level secret store
- Added additional parameters to the workflow input
- AWS steps are no longer depending on clusters directly, unless necessary (POC, will be finished in a subsequent PR)

### Why?

After finishing the POC implementations, it's time to improve the code.

### Checklist

### DIFF
--- a/internal/cluster/clustersecret/store.go
+++ b/internal/cluster/clustersecret/store.go
@@ -64,7 +64,7 @@ type SecretStore interface {
 	EnsureSecretExists(organizationID uint, secret SecretCreateRequest) (string, error)
 }
 
-// EnsureSecretExists creates a secret for an organization if it cannot be found and returns it's ID.
+// EnsureSecretExists creates a secret for a cluster if it cannot be found and returns it's ID.
 func (s *Store) EnsureSecretExists(ctx context.Context, clusterID uint, secret SecretCreateRequest) (string, error) {
 	cluster, err := s.clusters.GetCluster(ctx, clusterID)
 	if err != nil {
--- a/internal/cluster/clustersecret/store.go
+++ b/internal/cluster/clustersecret/store.go
@@ -60,11 +60,11 @@ type Cluster interface {
 
 // SecretStore is a generic secret store.
 type SecretStore interface {
-	// EnsureSecretExists creates a secret for an organization if it cannot be found.
+	// EnsureSecretExists creates a secret for an organization if it cannot be found and returns it's ID.
 	EnsureSecretExists(organizationID uint, secret NewSecret) (string, error)
 }
 
-// EnsureSecretExists creates a secret for a cluster if it cannot be found.
+// EnsureSecretExists creates a secret for an organization if it cannot be found and returns it's ID.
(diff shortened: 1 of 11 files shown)


internal/providers/pke/pkeworkflow/generate_certificates_activity.go
```diff
@@ -16,51 +16,44 @@ package pkeworkflow
...
-func NewGenerateCertificatesActivity(clusters Clusters) *GenerateCertificatesActivity {
+func NewGenerateCertificatesActivity(secrets ClusterSecretStore) *GenerateCertificatesActivity {
 	return &GenerateCertificatesActivity{
-		clusters: clusters,
+		secrets: secrets,
 	}
 }
 
+type ClusterSecretStore interface {
+	EnsureSecretExists(ctx context.Context, clusterID uint, secret clustersecret.NewSecret) (string, error)
```
siliconbrain: I think there should be a doc comment here too, stating what the method returns, because it is not obvious.

internal/cluster/clustersecret/store.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// NewSecret represents a new secret.
+type NewSecret struct {
```
orymate: i don't like this terminology. this is not a secret, this is only the input for creating one. and the term also collides with all the New* functions we have here. request is easier to understand

internal/providers/pke/pkeworkflow/generate_certificates_activity.go
sagikazarmark: Fair enough, I'll add better doc

internal/cluster/clustersecret/store.go
sagikazarmark: Well, it **is** a secret, a new one which is not persisted yet. I removed request on purpose, because this is not a request-response scenario, but I see how it might be confusing, so I will try to come up with something better. Feel free to suggest alternatives though.

internal/cluster/clustersecret/store.go
```diff
@@ -60,11 +60,11 @@ type Cluster interface {
 
 // SecretStore is a generic secret store.
 type SecretStore interface {
-	// EnsureSecretExists creates a secret for an organization if it cannot be found.
+	// EnsureSecretExists creates a secret for an organization if it cannot be found and returns it's ID.
 	EnsureSecretExists(organizationID uint, secret NewSecret) (string, error)
 }
 
-// EnsureSecretExists creates a secret for a cluster if it cannot be found.
+// EnsureSecretExists creates a secret for an organization if it cannot be found and returns it's ID.
```
orymate: for a cluster

PR #1773, opened 2019-02-21
Fix spot price parameter

Ecsy:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

Fix worker CF resulting in rollback, due to assuming spot based instances incorrectly.

### Why?

CloudFormation expects empty spot price in case of non spot based ASG.

### Checklist

### To Do

### DIFF
--- a/internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
+++ b/internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
@@ -94,7 +94,8 @@ func (a *CreateWorkerPoolActivity) Execute(ctx context.Context, input CreateWork
 		return "", emperror.Wrap(err, "loading CF template")
 	}
 
-	if input.Pool.SpotPrice == "0.0" {
+	spotPrice, err := strconv.ParseFloat(input.Pool.SpotPrice, 64)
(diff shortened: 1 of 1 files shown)


internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
```diff
@@ -94,6 +94,10 @@ func (a *CreateWorkerPoolActivity) Execute(ctx context.Context, input CreateWork
 		return "", emperror.Wrap(err, "loading CF template")
 	}
 
+	if input.Pool.SpotPrice == "0.0" {
```
stoader: if the zero value in `input.Pool.SpotPrice` can take various forms (e.g. `0.0; 0; .0`)  than it's safer to use something like:
```
if p, err := strconv.ParseFloat(input.Pool.SpotPrice, 64); err != nil || p == 0.0 {
  input.Pool.SpotPrice = ""
}
```

sagikazarmark: Already done 😄

PR #1776, opened 2019-02-22
Run vendor on demand

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Update vendor before build if needed, and add runall/run-worker targets to Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,8 +95,16 @@ config/dex.yml:
 
 .PHONY: run
 run: GOTAGS += dev
-run: build ## Build and execute a binary
+run: build-pipeline ## Build and execute a binary
 	PIPELINE_CONFIG_DIR=$${PWD}/config VAULT_ADDR="http://127.0.0.1:8200" ${BUILD_DIR}/${BINARY_NAME} ${ARGS}
+
+.PHONY: run-worker
+run-worker: GOTAGS += dev
+run-worker: build-worker ## Build and execute a binary
+	PIPELINE_CONFIG_DIR=$${PWD}/config VAULT_ADDR="http://127.0.0.1:8200" ${BUILD_DIR}/worker ${ARGS}
+
+.PHONY: runall ## Run worker and pipeline in foreground. Use with make -j.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ GOLANG_VERSION = 1.11
 GOFILES_NOVENDOR = $(shell find . -type f -name '*.go' -not -path "./vendor/*" -not -path "./client/*")
 
 .PHONY: up
-up: vendor config/dex.yml start config/config.toml ## Set up the development environment
+up: vendor config/dex.yml start config/config.toml ## Set up the development environment (override existing configuration)
```
sagikazarmark: It does not override existing configuration, creates configuration if there is none.

Makefile
```diff
@@ -81,6 +81,11 @@ bin/dep-${DEP_VERSION}:
 .PHONY: vendor
 vendor: bin/dep ## Install dependencies
 	bin/dep ensure -v -vendor-only
+	@touch vendor/.auto
+
+vendor/.auto: Gopkg.lock bin/dep # install/update dependencies if needed
```
sagikazarmark: This won't update anything, only install dependencies for the first time, same way as `make up`. Not sure I understand the added value.

orymate: I mean updates the vendor folder to match the revisions/versions in the lock file. So you won't be confused about broken builds which would succeed if you would have run `make vendor -vendor-only` before the build.

Makefile
orymate: Right, no source files for configs targets. I always want to run `make up` instead of `make start`, but only remembered it's not what I should do, but the reason is not that it overrides config files but that it runs an unnecessary vendoring.

Makefile
orymate: (Also see the other change in this commit)

sagikazarmark: Okay, got it, thanks!

PR #1789, opened 2019-02-25
import user name as organization during org sync

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
If a user logs in with some other provider than GitHub it's convenient to have its user name synced into an org when the Sync Orgs feature is called.

### Why?
To enable CI/CD for non through-GitHub registered users.

### Checklist

### To Do

### DIFF
--- a/auth/github.go
+++ b/auth/github.go
@@ -88,7 +88,7 @@ func getGithubOrganizations(token string) ([]organization, error) {
 
 	memberships, _, err := githubClient.Organizations.ListOrgMemberships(oauth2.NoContext, nil)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to list organization memberships")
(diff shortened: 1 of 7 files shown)


auth/github.go
```diff
@@ -103,6 +103,19 @@ func getGithubOrganizations(token string) ([]organization, error) {
...
 
+	user, _, err := githubClient.Users.Get(context.Background(), "")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to fetch user from github")
+	}
+
+	userOrg := organization{
+		name:     *user.Login,
+		role:     "admin",
+		provider: "user",
```
tothandras: Providers should probably come from constants.

PR #1802, opened 2019-02-27
Dex Cluster auth for PKE

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
This PR enabled Dex OIDC authentication for PKE Kubernetes clusters.

### Checklist

### To Do


database/migrations/1551272865_add_dex_enabled_field_to_amazon_ec2_clusters.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE `amazon_ec2_clusters` ADD COLUMN `dex_enabled` tinyint(1) DEFAULT NULL;
```
sagikazarmark: Shouldn't this be `NOT NULL DEFAULT 0`?

database/migrations/1551272865_add_dex_enabled_field_to_amazon_ec2_clusters.up.sql
```diff
@@ -0,0 +1,4 @@
...
+
+UPDATE `amazon_ec2_clusters` SET `dex_enabled` = NULL;
```
sagikazarmark: Default value should be 0 IMHO.

Gopkg.toml
```diff
@@ -284,3 +284,7 @@ required = [ "github.com/apache/thrift/lib/go/thrift"]
 [[constraint]]
   name = "github.com/spf13/pflag"
   version = "1.0.1"
+
+[[constraint]]
+  name = "github.com/dexidp/dex"
```
sagikazarmark: Instead of installing dex as yet another dependency, I think it would be better to retrieve the proto file (download, submodule, whatever) and generate the client ourselves.

This is the recommendation in the dex docs BTW: https://github.com/dexidp/dex/blob/master/Documentation/api.md#generating-clients

See an example for handling proto files in a "real life" application here: https://github.com/sagikazarmark/modern-go-application

Although this would be nice to have in this PR, postponing to a subsequent would also be acceptable by me.

bonifaido: Good idea, however, I would push that into a separate PR.

sagikazarmark: See #1810

PR #1813, opened 2019-02-28
Switch to Go Modules

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR switches to Go Modules from Dep.

### Why?

Go modules is the future.

### Checklist


api/common_stuff.go
```diff
@@ -30,7 +30,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/api/googleapi"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
pbalogh-sa: I found three different naming of  k8s.io/apimachinery/pkg/apis/meta/v1, it would be fine to use one common name insted of metav1, v1meta and meta_v1.

api/cluster.go
```diff
@@ -42,7 +42,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"go.uber.org/cadence/client"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
pbalogh-sa: I found three different naming of  k8s.io/apimachinery/pkg/apis/meta/v1, it would be fine to use one common name insted of metav1, v1meta and meta_v1

api/custom_test.go
```diff
@@ -20,9 +20,9 @@ import (
 	"testing"
 
 	pkgHelm "github.com/banzaicloud/pipeline/pkg/helm"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
-	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
```
pbalogh-sa: I found three different naming of  k8s.io/apimachinery/pkg/apis/meta/v1, it would be fine to use one common name insted of metav1, v1meta and meta_v1

api/common_stuff.go
sagikazarmark: I would rather do this in a subsequent PR as I think this is probably a broader problem.

pbalogh-sa:
I opened a new issue about import naming:  #1814

pbalogh-sa:
thx @sagikazarmark it works for me properly with `GO111MODULE=on make build`

PR #1825, opened 2019-02-28
Fix delete all deployments

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

### Why?

If a helm upgrade of a deployment fails the deployment will have two states: the original state and a PENDING_UPGRADE state. This caused Pipeline to delete such deployments twice the second attempt failing since the first already deleted the deployment.

### Checklist

### DIFF
--- a/helm/helm.go
+++ b/helm/helm.go
@@ -191,18 +191,19 @@ func DeleteAllDeployment(log logrus.FieldLogger, kubeconfig []byte) error {
 		//      release name = release1, status = DEPLOYED
 		//
 		// we need only the release name for deleting a release
-		releases := make(map[string]bool)
+		deletedDeployments := make(map[string]bool)
 		for _, r := range releaseResp.Releases {
-			releases[r.Name] = true
-		}
+			if _, ok := deletedDeployments[r.Name]; !ok {
+				log.Infoln("deleting deployment", r.Name)
 
-		for release := range releases {
-			log.Infoln("deleting deployment", release)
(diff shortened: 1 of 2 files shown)


helm/helm.go
```diff
@@ -176,22 +177,32 @@ func GetChartFile(file []byte, fileName string) (string, error) {
...
+		//
+		// we need only the release name for deleting a release
+		releases := make(map[string]bool)
 		for _, r := range releaseResp.Releases {
-			log.Info("Trying to delete deployment ", r.Name)
-			err := DeleteDeployment(r.Name, kubeconfig)
+			releases[r.Name] = true
+		}
+
+		for release := range releases {
```
sagikazarmark: Why not merge the two loops and skip already deleted releases?

stoader: done

helm/helm.go
```diff
@@ -176,22 +177,33 @@ func GetChartFile(file []byte, fileName string) (string, error) {
...
+func DeleteAllDeployment(log logrus.FieldLogger, kubeconfig []byte) error {
+	log.Info("getting deployments....")
 	filter := ""
 	releaseResp, err := ListDeployments(&filter, "", kubeconfig)
 	if err != nil {
-		return err
+		return emperror.Wrap(err, "failed to get deployments")
 	}
-	log.Info("Starting deleting deployments")
+	log.Info("start deleting deployments")
```
orymate:
```suggestion
```

PR #1833, opened 2019-03-04
Simplify cluster search implementation

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Move the cluster query filter generation out of the common method and into the specialization methods.

### Why?
The current implementation uses too much type checking.

### Checklist

### DIFF
--- a/internal/cluster/clusters.go
+++ b/internal/cluster/clusters.go
@@ -71,17 +71,30 @@ func (c *Clusters) FindByOrganization(organizationID uint) ([]*model.ClusterMode
 
 // FindOneByID returns a cluster instance for an organization by cluster ID.
 func (c *Clusters) FindOneByID(organizationID uint, clusterID uint) (*model.ClusterModel, error) {
-	return c.findOneBy(organizationID, "id", clusterID)
+	cluster, err := c.findOneBy(model.ClusterModel{
+		OrganizationId: organizationID,
+		ID:             clusterID,
+	})
+	if err != nil {
+		return nil, emperror.Wrap(err, "could not find cluster by ID")
+	}
+	return cluster, nil
 }
 
 // FindOneByName returns a cluster instance for an organization by cluster name.
 func (c *Clusters) FindOneByName(organizationID uint, clusterName string) (*model.ClusterModel, error) {
-	return c.findOneBy(organizationID, "name", clusterName)
+	cluster, err := c.findOneBy(model.ClusterModel{
+		OrganizationId: organizationID,
+		Name:           clusterName,
+	})
+	if err != nil {
+		return nil, emperror.Wrap(err, "could not find cluster by name")
+	}
+	return cluster, nil
 }
 
 type clusterModelNotFoundError struct {
-	cluster        interface{}
-	organizationID uint
+	cluster model.ClusterModel
 }
 
 func (e *clusterModelNotFoundError) Error() string {
@@ -90,8 +103,9 @@ func (e *clusterModelNotFoundError) Error() string {
 
 func (e *clusterModelNotFoundError) Context() []interface{} {
 	return []interface{}{
-		"cluster", e.cluster,
-		"organization", e.organizationID,
+		"clusterID", e.cluster.ID,
+		"clusterName", e.cluster.Name,
+		"organizationID", e.cluster.OrganizationId,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


internal/cluster/clusters.go
```diff
@@ -100,44 +114,20 @@ func (e *clusterModelNotFoundError) NotFound() bool {
...
-	err := c.db.Where(cluster).Preload("ScaleOptions").First(&cluster).Error
+func (c *Clusters) findOneBy(cluster model.ClusterModel) (*model.ClusterModel, error) {
+	var result model.ClusterModel
+	err := c.db.Where(cluster).Preload("ScaleOptions").First(&result).Error
 	if gorm.IsRecordNotFoundError(err) {
 		return nil, errors.WithStack(&clusterModelNotFoundError{
-			cluster:        criteria,
-			organizationID: organizationID,
+			cluster:        cluster,
+			organizationID: cluster.OrganizationId,
```
sagikazarmark: Since the cluster contains the organization ID, I wonder if this is necessary.

internal/cluster/clusters.go
```diff
@@ -100,44 +114,20 @@ func (e *clusterModelNotFoundError) NotFound() bool {
...
 		})
-	} else if err != nil {
-		return nil, emperror.With(
-			errors.Wrapf(err, "could not get cluster by %s", field),
-			"cluster", criteria,
-			"organization", organizationID,
-		)
+	}
+	if err != nil {
+		return nil, emperror.With(err, "cluster", cluster, "organization", cluster.OrganizationId)
```
sagikazarmark: Check `clusterModelNotFoundError` struct: it already extracts the context from itself, although you might have to update it as well.

internal/cluster/clusters.go
```diff
@@ -71,17 +71,30 @@ func (c *Clusters) FindByOrganization(organizationID uint) ([]*model.ClusterMode
...
+	if err != nil {
+		return nil, emperror.Wrap(err, "could not find cluster by name")
+	}
+	return cluster, nil
 }
 
 type clusterModelNotFoundError struct {
-	cluster        interface{}
-	organizationID uint
+	cluster interface{}
```
sagikazarmark: This could be `ClusterModel`

internal/cluster/clusters.go
```diff
@@ -91,7 +104,6 @@ func (e *clusterModelNotFoundError) Error() string {
 func (e *clusterModelNotFoundError) Context() []interface{} {
 	return []interface{}{
 		"cluster", e.cluster,
-		"organization", e.organizationID,
```
sagikazarmark: Maybe use some values here instead of the whole struct.

internal/cluster/clusters.go
```diff
@@ -100,44 +112,19 @@ func (e *clusterModelNotFoundError) NotFound() bool {
...
 		})
-	} else if err != nil {
-		return nil, emperror.With(
-			errors.Wrapf(err, "could not get cluster by %s", field),
-			"cluster", criteria,
-			"organization", organizationID,
-		)
+	}
+	if err != nil {
+		return nil, emperror.With(err, "cluster", cluster)
```
sagikazarmark: `emperror.With` is not necessary at all IMHO.

internal/cluster/clusters.go
```diff
@@ -100,44 +114,23 @@ func (e *clusterModelNotFoundError) NotFound() bool {
...
-		name, ok := criteria.(string)
-		if !ok {
-			return nil, errors.New("criteria is not a valid string value for name")
-		}
-
-		cluster.Name = name
-	}
-
-	err := c.db.Where(cluster).Preload("ScaleOptions").First(&cluster).Error
+func (c *Clusters) findOneBy(cluster model.ClusterModel) (*model.ClusterModel, error) {
```
orymate: i would trigger an error somewhere if all IDs are zero (default).

orymate: (also take into consideration that iirc there were some calls with orgid=0 and clusterid!=0 that take advantage of that clusterid is unique in itself)

internal/cluster/clusters.go
```diff
@@ -90,8 +103,9 @@ func (e *clusterModelNotFoundError) Error() string {
 
 func (e *clusterModelNotFoundError) Context() []interface{} {
 	return []interface{}{
-		"cluster", e.cluster,
-		"organization", e.organizationID,
+		"clusterID", e.cluster.ID,
```
orymate: don't we pass these fields to the API? please update openapi description if we do

siliconbrain: I see no relevant use of `cluster` or `organization` as property name in the API description, nor does `ClusterNotFound` refer to them, so I don't think they're passed anywhere.

PR #1853, opened 2019-03-06
dex: propagate client delete error

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Propagate low-level dex error correctly from the activity.

### Why?
To see the root cause of deletion issues.

### Checklist

### To Do

### DIFF
--- a/internal/providers/pke/pkeworkflow/delete_dex_client.go
+++ b/internal/providers/pke/pkeworkflow/delete_dex_client.go
@@ -16,10 +16,9 @@ package pkeworkflow
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/banzaicloud/pipeline/internal/cluster/auth"
-	"github.com/pkg/errors"
+	"github.com/goph/emperror"
 )
 
 const DeleteDexClientActivityName = "pke-delete-dex-client-activity"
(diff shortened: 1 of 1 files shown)


internal/providers/pke/pkeworkflow/delete_dex_client.go
```diff
@@ -48,7 +47,7 @@ func (a *DeleteDexClientActivity) Execute(ctx context.Context, input DeleteDexCl
 
 	err = a.clusterAuthService.UnRegisterCluster(ctx, cluster.GetUID())
 	if err != nil {
-		return errors.New(fmt.Sprintf("can't delete dex client for cluster %t", cluster))
+		return emperror.Wrapf(err, "can't delete dex client for cluster: %d", input.ClusterID)
```
orymate:
```suggestion
		return emperror.Wrap(err, "can't delete dex client for cluster")
```

the  cluster id should be clear from the context added during error handling

bonifaido: Thanks!

PR #1855, opened 2019-03-07
Fix node label for standalone master

tarokkk:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The kubernetes controller-manager won't attach nodes with the master
role (node-role.kubernetes.io/master-worker) active.
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/service/service_controller.go#L588

### DIFF
--- a/internal/providers/pke/pkeworkflow/set_master_taint_activity.go
+++ b/internal/providers/pke/pkeworkflow/set_master_taint_activity.go
@@ -16,19 +16,18 @@ package pkeworkflow
 
 import (
 	"context"
-	"encoding/json"
 
 	"github.com/banzaicloud/pipeline/pkg/k8sclient"
 	"github.com/goph/emperror"
 	"go.uber.org/cadence/activity"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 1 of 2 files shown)


internal/providers/pke/pkeworkflow/set_master_taint_activity.go
```diff
@@ -62,21 +60,14 @@ func (a *SetMasterTaintActivity) Execute(ctx context.Context, input SetMasterTai
...
-	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: selector})
+	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{LabelSelector: masterKey})
 	if err != nil {
 		return emperror.Wrap(err, "failed to list master nodes")
 	}
 
 	for _, node := range nodes.Items {
 		logger := logger.With("node", node.Name)
-		taints := []v1.Taint{}
+		taints := make([]v1.Taint, 0)
```
siliconbrain: This creates an empty slice unnecessarily, instead:
```go
var taints []v1.Taint
```
where `taints` will be `nil` which is an empty slice. `append` will allocate a sufficiently sized slice on-the-fly anyway.

orymate:
```suggestion
		var taints []v1.Taint
```

PR #1867, opened 2019-03-08
pke: bump pkeVersion

orymate:
### What's in this PR?
Update pke tool


tarokkk:
Won't be better if we move this to configuration?

PR #1877, opened 2019-03-12
Speed up storage listing in case of Amazon & Alibaba

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
- In case of (non managed) storages we don't add location field to the response items. 
- In install logging posthook check the region field, if it's empty we send a get region request to cloud and set the region field
- To initialize Amazon & Alibaba clients we use default regions
- Make configurable these default regions

### Why?
The current storage listing is extremly slow (compared to other providers). These numbers are depend on how many storage we have.

In case of 95 Amazon storage it takes `~8 sec` to get a response.
In case of 8 Alibaba storage it takes `~10-12 sec` to get a response.

With this solution:
In case of 95 Amazon storage it takes `~1-2 sec` to get a response.
In case of 8 Alibaba storage it takes `~1-2 sec` to get a response.

### Checklist

### DIFF
--- a/api/bucket.go
+++ b/api/bucket.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/banzaicloud/pipeline/auth"
+	pipConfig "github.com/banzaicloud/pipeline/config"
 	"github.com/banzaicloud/pipeline/internal/objectstore"
 	"github.com/banzaicloud/pipeline/internal/platform/gin/correlationid"
 	ginutils "github.com/banzaicloud/pipeline/internal/platform/gin/utils"
@@ -35,6 +36,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
 )
 
 const (
@@ -108,15 +110,10 @@ func ListBuckets(c *gin.Context) {
 	}
 
 	switch cloudType {
-	case pkgProviders.Alibaba, pkgProviders.Amazon:
-		location, ok := ginutils.RequiredQueryOrAbort(c, "location")
-		if !ok {
-			logger.Debug("missing location")
-
-			return
-		}
-
-		objectStoreCtx.Location = location
+	case pkgProviders.Amazon:
+		objectStoreCtx.Location = viper.GetString(pipConfig.AmazonInitializeRegionKey)
+	case pkgProviders.Alibaba:
+		objectStoreCtx.Location = viper.GetString(pipConfig.AlibabaInitializeRegionKey)
 	}
 
 	objectStore, err := providers.NewObjectStore(objectStoreCtx, logger)
--- a/cluster/gke.go
+++ b/cluster/gke.go
@@ -1961,7 +1961,7 @@ func (c *GKECluster) IsReady() (bool, error) {
 		return false, err
 	}
 
-	c.log.Debug("Get gke cluster with name %s", c.model.Cluster.Name)
+	c.log.Debug("Get gke cluster")
 	cl, err := svc.Projects.Zones.Clusters.Get(secretItem.GetValue(pkgSecret.ProjectId), c.model.Cluster.Location, c.model.Cluster.Name).Context(context.Background()).Do()
 	if err != nil {
 		apiError := getBanzaiErrorFromError(err)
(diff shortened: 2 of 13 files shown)


api/bucket.go
```diff
@@ -108,15 +110,10 @@ func ListBuckets(c *gin.Context) {
...
-	case pkgProviders.Alibaba, pkgProviders.Amazon:
-		location, ok := ginutils.RequiredQueryOrAbort(c, "location")
-		if !ok {
-			logger.Debug("missing location")
-
-			return
-		}
-
-		objectStoreCtx.Location = location
+	case pkgProviders.Amazon:
```
stoader: We should fall back to the default location only when no "location" query param is provided by the caller.

internal/providers/amazon/objectstore.go
```diff
@@ -338,3 +330,23 @@ func (s *objectStore) searchCriteria(bucketName string) *ObjectStoreBucketModel
...
 	}
 }
+
+// GetBucketRegion returns with the given bucket's region from Amazon
+func GetBucketRegion(secret *secret.SecretItemResponse, bucketName string, orgID uint, log logrus.FieldLogger) (region string, err error) {
+
+	var org *auth.Organization
+	org, err = auth.GetOrganizationById(orgID)
+	if err != nil {
+		return
```
stoader: Use emperror to provide error context

internal/providers/amazon/objectstore.go
```diff
@@ -338,3 +330,23 @@ func (s *objectStore) searchCriteria(bucketName string) *ObjectStoreBucketModel
...
+	}
+
+	var s *objectStore
+	defaultRegion := viper.GetString(pipConfig.StorageDefaultAmazonRegionKey)
+	// we don't need DB here, this bucket information came from the cloud
+	s, err = NewObjectStore(defaultRegion, secret, org, nil, log, false)
+	if err != nil {
+		return
```
stoader: Provide error context to the error.

internal/providers/alibaba/objectstore.go
```diff
@@ -319,3 +314,29 @@ func (os *objectStore) deleteFailed(bucket *ObjectStoreBucketModel, reason error
 	}
 	return reason
 }
+
+func GetBucketLocation(secret *secret.SecretItemResponse, bucketName string, orgID uint, log logrus.FieldLogger) (location string, err error) {
+
+	var org *auth.Organization
+	org, err = auth.GetOrganizationById(orgID)
+	if err != nil {
+		return
```
stoader: Use emperror to provide error context

internal/providers/alibaba/objectstore.go
```diff
@@ -319,3 +314,29 @@ func (os *objectStore) deleteFailed(bucket *ObjectStoreBucketModel, reason error
...
+	}
+
+	var s *objectStore
+	defaultRegion := viper.GetString(pipConfig.StorageDefaultAlibabaRegionKey)
+	// we don't need DB here, this bucket information came from the cloud
+	s, err = NewObjectStore(defaultRegion, secret, org, nil, log, false)
+	if err != nil {
+		return
```
stoader: Use emperror to provide error context

internal/providers/alibaba/objectstore.go
```diff
@@ -319,3 +314,29 @@ func (os *objectStore) deleteFailed(bucket *ObjectStoreBucketModel, reason error
...
+	}
+
+	location, err = s.objectStore.GetLocation(bucketName)
+	if err != nil {
+		return
```
stoader: Use emperror to provide error context

cluster/gke.go
```diff
@@ -1961,7 +1961,7 @@ func (c *GKECluster) IsReady() (bool, error) {
 		return false, err
 	}
 
-	c.log.Debug("Get gke cluster with name %s", c.model.Cluster.Name)
+	c.log.Debugf("Get gke cluster with name %s", c.model.Cluster.Name)
```
stoader:
```suggestion
	c.log.Debug("Get gke cluster with name")
```

either cluster id or cluster name field is already set on `c.log`

config/config.toml.dist
```diff
@@ -205,3 +205,9 @@ port = 7933
 domain = "pipeline"
 createNonexistentDomain = true
 workflowExecutionRetentionPeriodInDays = 3
+
+[storage]
```
stoader: Since this is not specific to cloud provider storage API I'd name it something like
[amazon]
defaultApiRegion

[alibaba]
defaultApiRegion

colin014: thx, i renamed these

PR #1888, opened 2019-03-19
cloudinfo: additional logging & locking

sancyx:
- Remove location from cache key, since there's no difference in instance types
- Add some log & locking around instancetype map 

### Checklist

### To Do

### DIFF
--- a/internal/cloudinfo/machine_types.go
+++ b/internal/cloudinfo/machine_types.go
@@ -60,7 +60,7 @@ type VMKey struct {
 }
 
 type InstanceTypeMap struct {
-	sync.RWMutex
+	lock     sync.RWMutex
 	internal map[VMKey]MachineDetails
 }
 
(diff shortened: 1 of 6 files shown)


internal/cloudinfo/machine_types.go
```diff
@@ -77,6 +103,7 @@ func fetchMachineTypes(cloud string, service string, region string, location str
 	ciRequest.Header.Set("Content-Type", "application/json")
 	httpClient := &http.Client{}
 
+	log.Infof("fetching machine types from %s for %s / %s / %s", cloudInfoEndPoint, cloud, region, service)
```
stoader: Use `log.WithFields` for structured logging.

internal/cloudinfo/machine_types.go
```diff
@@ -54,14 +56,38 @@ type VMKey struct {
...
-	location     string
 	instanceType string
 }
 
+type InstanceTypeMap struct {
+	sync.RWMutex
+	internal map[VMKey]MachineDetails
+}
+
+func NewInstanceTypeMap() *InstanceTypeMap {
```
stoader: Inject logger and error handler as input parameters to this factory method similar how it's done in `cluster.NewManager`

internal/cloudinfo/machine_types.go
```diff
@@ -54,14 +56,38 @@ type VMKey struct {
...
+	im.RUnlock()
+	return result, ok
+}
+
+// nolint: gochecknoglobals
+var instanceTypeMap = NewInstanceTypeMap()
+
 // nolint: gochecknoglobals
-var instanceTypeMap = make(map[VMKey]MachineDetails)
+var log logrus.FieldLogger
```
stoader: Inject the logger instead of using global logger.

internal/cloudinfo/machine_types.go
```diff
@@ -54,14 +56,38 @@ type VMKey struct {
...
 
-func fetchMachineTypes(cloud string, service string, region string, location string) error {
+func init() {
```
stoader: Inject the logger instead of using global logger.

internal/cloudinfo/machine_types.go
```diff
@@ -54,14 +56,38 @@ type VMKey struct {
...
+	log = config.Logger()
+}
+
+func fetchMachineTypes(cloud string, service string, region string) error {
```
stoader: Can we inject `cloudInfoEndPoint` as parameter instead of `viper.GetString(config.CloudInfoEndPoint)`?

internal/cloudinfo/machine_types.go
```diff
@@ -54,14 +56,38 @@ type VMKey struct {
 	cloud        string
 	service      string
 	region       string
-	location     string
 	instanceType string
 }
 
+type InstanceTypeMap struct {
+	sync.RWMutex
```
sagikazarmark: Do not embed synchronization structures, it makes them publicly available.

internal/cloudinfo/machine_types.go
sancyx: logger is created from context

internal/cloudinfo/machine_types.go
sancyx: logger is created from context

internal/cloudinfo/machine_types.go
sancyx: fixed

internal/cloudinfo/machine_types.go
```diff
@@ -85,37 +107,40 @@ func fetchMachineTypes(cloud string, service string, region string, location str
 	var vmDetails CloudInfoResponse
 	json.Unmarshal(respBody, &vmDetails)
 
+	instanceTypeMap.lock.Lock()
```
stoader: IMHO we never should use `instanceTypeMap.lock` from outside of the enclosing struct.

I'd add a `update(*InstanceTypeMap)` method to `InstanceTypeMap` or `setVM(VMKey, *MachineDetails)` and lock/unlock inside the method

PR #1895, opened 2019-03-20
add rewrite transport to pipeline api proxy

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add url rewrite functionality to Pipeline Proxy. This is need in case the proxied application is using absolute url's like Kiali.

### Checklist


cluster/manager.go
```diff
@@ -106,15 +107,15 @@ func (m *Manager) getPrometheusTimer(provider, location, status string, orgId ui
...
+func (m *Manager) GetKubeProxy(requestUrl *url.URL, apiProxyPrefix string, commonCluster CommonCluster) (*KubeAPIProxy, error) {
 	// Currently we do not lock this transaction of getting and optionally creating a KubeAPIProxy.
 	// The worst thing that could happen is that for a short period (a Go GC period) there will be
 	// an extra KubeAPIProxy object in memory, but we can keep this method lock-free I think this is a good trade-off.
 	kubeProxy, found := m.kubeProxyCache.Get(commonCluster.GetUID())
 	if !found {
 		var err error
 
-		kubeProxy, err = NewKubeAPIProxy(apiProxyPrefix, commonCluster, defaultProxyExpirationMinutes*time.Minute)
+		kubeProxy, err = NewKubeAPIProxy(requestUrl, apiProxyPrefix, commonCluster, defaultProxyExpirationMinutes*time.Minute)
```
sagikazarmark: The proxy is cached globally for a cluster UID.

sancyx: Yes the request url is the same for each cluster, so it's not a problem that we have the cache by clusterId.

sagikazarmark: Well, that's not necessarily true (eg. if I add a random query to the API, the URL won't be the same. As far as I can see only the host and the scheme is used at the moment. Maybe it's worth a comment somewhere saying that relying on other values might cause caching issues.

PR #1918, opened 2019-03-27
Add created by field to cluster status history

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

The cluster status history will contain the creator from now on.

### Checklist


orymate:
this does not solve the whole problem

PR #1924, opened 2019-03-29
Better logging for stack trace

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR includes better logging for error stack traces: proper panic position will be logged.

Also, panicking cluster operations will be recorded in cluster status.

### Why?

Debugging based on logs is not really effective at the moment as most of the error information is not in the logs. This PR aims to improve that.

### Checklist


cluster/manager.go
```diff
@@ -94,6 +96,40 @@ func (m *Manager) getErrorHandler(ctx context.Context) emperror.Handler {
 	return pipelineContext.ErrorHandlerWithCorrelationID(ctx, m.errorHandler)
 }
 
+type clusterErrorHandler struct {
+	handler       emperror.Handler
+	status        string
+	statusMessage string
+	cluster       CommonCluster
```
orymate: check this and if org/userid is needed

PR #1925, opened 2019-03-29
use cloudinfo client

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes

### Why?
Use CloudInfo client instead of using default http client to fetch VM machine details.

### Checklist

- [x] Error handling code meets the [guideline]


internal/cloudinfo/machine_types.go
```diff
@@ -94,39 +96,42 @@ func (im *InstanceTypeMap) setMachines(cloud string, service string, region stri
...
 	}
-	respBody, _ := ioutil.ReadAll(ciResponse.Body)
-	var vmDetails CloudInfoResponse
-	json.Unmarshal(respBody, &vmDetails)
-
-	instanceTypeMap.setMachines(cloud, service, region, vmDetails.Products)
+	if getProductsOK.Payload != nil {
+		instanceTypeMap.setMachines(cloud, service, region, getProductsOK.Payload.Products)
+	} else {
+		return errors.New("error fetching machine types from CloudInfo: empty Payload in response")
```
baluchicken: use emperror wrap

internal/cloudinfo/machine_types.go
```diff
@@ -94,39 +96,42 @@ func (im *InstanceTypeMap) setMachines(cloud string, service string, region stri
...
-	}
-	cloudInfoUrl := fmt.Sprintf(
-		"%s/providers/%s/services/%s/regions/%s/products",
-		cloudInfoEndPoint, cloud, service, region)
-	ciRequest, err := http.NewRequest(http.MethodGet, cloudInfoUrl, nil)
-	if err != nil {
-		return errors.New("error fetching machine types from CloudInfo")
+	ciHostName := viper.GetString(config.CloudInfoHostName)
+	if len(ciHostName) == 0 {
+		return emperror.With(errors.New("missing config"), "ciHostName", config.CloudInfoHostName)
```
baluchicken: I don't think ciHostName is good name for the context, I would use fmt.Errorf("missing config: %s", config.CloudInfoHostName) instead of just errors.New and a context.

sancyx: renamed

internal/cloudinfo/machine_types.go
sancyx: the error is wrapped in calling function

PR #1934, opened 2019-04-02
Azure PKE integration

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | yes
| Related tickets | -

### What's in this PR?
Implementation of PKE on Azure

### Why?
PKE + Azure + ??? = $$$

### Checklist


api/cluster/base.go
```diff
@@ -0,0 +1,127 @@
...
+	pkgCommon "github.com/banzaicloud/pipeline/pkg/common"
+	"github.com/banzaicloud/pipeline/secret"
+)
+
+// CreateClusterRequestBase describes the common base of cluster creation requests
+type CreateClusterRequestBase struct {
+	Name         string               `json:"name" yaml:"name" binding:"required"`
+	Features     []Feature            `json:"features"`
+	SecretID     string               `json:"secretId" yaml:"secretId"`
+	SecretName   string               `json:"secretName" yaml:"secretName"`
```
sagikazarmark: Why do we need both secret name and secret ID?

siliconbrain: No idea. We had them in the old API, I thought it had significance. I'd gladly remove one of them.

orymate: Because the user of the API might know either of them, and there is no API exposed to exchange them other than looking up in all secret's list. The mapping from name to ID is not defined from end user's perspective (and I hope it will change at some time).

api/cluster/pke_azure.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	intPKE "github.com/banzaicloud/pipeline/internal/pke"
+	"github.com/banzaicloud/pipeline/internal/providers/azure/pke"
+	"github.com/banzaicloud/pipeline/internal/providers/azure/pke/driver"
+)
+
+const PKEOnAzure = pke.PKEOnAzure
+
+type NodePool struct {
```
sagikazarmark: Should this be called `AzureNodePool`?

api/cluster_create.go
```diff
@@ -16,28 +16,115 @@ package api
...
 	a.logger.Info("Cluster creation started")
 
-	a.logger.Debug("Bind json into CreateClusterRequest struct")
-	// bind request body to struct
+	ctx := ginutils.Context(context.Background(), c)
+
+	orgID := auth.GetCurrentOrganization(c.Request).ID
+	userID := auth.GetCurrentUser(c.Request).ID
+
+	a.logger.Debug("Try to bind JSON to CreateClusterRequest struct")
```
sagikazarmark: This log isn't necessary (or should be trace level instead). I would probably log the binding error instead (using info level, since this isn't an error from the application point of view).

docs/openapi/pipeline.yaml
```diff
@@ -5332,6 +5335,191 @@ components:
...
+                                                example: "10.0.0.0/24"
+                                    zones:
+                                        type: array
+                                        items:
+                                            type: string
+                                            enum:
+                                                - "1"
+                                                - "2"
+                                                - "3"
+                                        example: ["1", "3"]
```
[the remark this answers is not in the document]
tarokkk: This is purely Azure implementation, although I think more than 3 zone can be possible.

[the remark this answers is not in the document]
siliconbrain: It seems there are only the three options: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-use-availability-zones

client/api/openapi.yaml
```diff
@@ -5670,6 +5670,190 @@ components:
...
+            example: westeurope
+            type: string
+          resourceGroup:
+            example: my-resource-group-1
+            type: string
+          network:
+            properties:
+              name:
+                type: string
+              cidr:
```
stoader: shouldn't this be a collection of cidrs?

siliconbrain: I'm not sure if we'd like to support creating and automatically allocating subranges from a multi-range network. When the network already exists only its name has to be specified as reference.

PR #1976, opened 2019-04-18
Ack fixes

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

- Access ESS service at the endpoints listed in here: https://www.alibabacloud.com/help/doc-detail/25927.htm
- Fix the logic that determines what node pool a node belongs to
- After tiller is upgraded wait until the old tiller pod stops and the new one comes up before considering it as complete

### Checklist

### DIFF
--- a/cluster/ack.go
+++ b/cluster/ack.go
@@ -132,12 +132,12 @@ func (c *ACKCluster) SetTTL(ttl time.Duration) {
 func (c *ACKCluster) ListNodeNames() (pkgCommon.NodeNames, error) {
 	essClient, err := c.GetAlibabaESSClient(nil)
 	if err != nil {
-		return nil, emperror.Wrap(err, "failed to get Alibaba ESS client")
+		return nil, err
 	}
 
 	ecsClient, err := c.GetAlibabaECSClient(nil)
 	if err != nil {
-		return nil, emperror.Wrap(err, "failed to get Alibaba ECS client")
+		return nil, err
 	}
 
 	request := ess.CreateDescribeScalingInstancesRequest()
@@ -161,7 +161,7 @@ func (c *ACKCluster) ListNodeNames() (pkgCommon.NodeNames, error) {
 
 		var instanceIds []string
 		for _, instance := range response.ScalingInstances.ScalingInstance {
-			instanceIds = append(instanceIds,  strconv.Quote(instance.InstanceId))
+			instanceIds = append(instanceIds, strconv.Quote(instance.InstanceId))
 		}
 
 		if len(instanceIds) > 0 {
--- a/cluster/ack.go
+++ b/cluster/ack.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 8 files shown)


cluster/ack.go
```diff
@@ -130,13 +132,24 @@ func (c *ACKCluster) SetTTL(ttl time.Duration) {
 func (c *ACKCluster) ListNodeNames() (pkgCommon.NodeNames, error) {
 	essClient, err := c.GetAlibabaESSClient(nil)
 	if err != nil {
-		return nil, err
+		return nil, emperror.Wrap(err, "failed to get Alibaba ESS client")
+	}
+
+	ecsClient, err := c.GetAlibabaECSClient(nil)
+	if err != nil {
+		return nil, emperror.Wrap(err, "failed to get Alibaba ECS client")
```
baluchicken: This log message is duplicated, we already put this message to the log inside c.GetAlibabaECSClient

cluster/ack.go
```diff
@@ -130,13 +132,24 @@ func (c *ACKCluster) SetTTL(ttl time.Duration) {
 func (c *ACKCluster) ListNodeNames() (pkgCommon.NodeNames, error) {
 	essClient, err := c.GetAlibabaESSClient(nil)
 	if err != nil {
-		return nil, err
+		return nil, emperror.Wrap(err, "failed to get Alibaba ESS client")
```
baluchicken: This log message is duplicated, we already put this message to the log inside cGetAlibabaESSClient

helm/install.go
```diff
@@ -355,3 +363,33 @@ func Install(helmInstall *phelm.Install, kubeConfig []byte) error {
 	log.Info("Helm install finished")
 	return nil
 }
+
+// watchTillerUntilReady waits for the tiller pod to become available.
+//
+// Returns true if it exists. If the timeout was reached and it could not find the pod, it returns false.
+func watchTillerUntilReady(namespace string, client *kubernetes.Clientset, timeout time.Duration, newImage string) bool {
+	deadlinePollingChan := time.NewTimer(timeout).C
```
baluchicken: minor: we already have a library which handles backoff, https://github.com/banzaicloud/pipeline/tree/master/internal/backoff

stoader: Removing this part as this is already handled in `WaitingForTillerComeUp`

pkg/providers/alibaba/provider.go
```diff
@@ -14,4 +14,32 @@
...
 const Provider = "alibaba"
+
+// GetESSServiceEndpoint returns the endpoint of the ESS Service in the
+// given region (https://www.alibabacloud.com/help/doc-detail/25927.htm)
+func GetESSServiceEndpoint(region string) string {
+
+	region = strings.ToLower(region)
+
+	switch strings.ToLower(region) {
+	case "cn-zhangjiakou":
```
bonifaido: Please add `fallthrough` statements.

PR #1987, opened 2019-05-03
Fix azure pke infra delete

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
If 'force' flag is passed than continue the deletion of infra resources in case an error is encountered.

### Checklist


internal/providers/azure/pke/driver/cluster_deleter.go
```diff
@@ -91,8 +93,8 @@ func (cd AzurePKEClusterDeleter) Delete(ctx context.Context, cluster pke.PKEOnAz
 		ssns[i] = pke.GetVMSSName(cluster.Name, np.Name)
 	}
 
-	servicePips, err := getServicesPublicIPs(cd.logger, k8sConfig)
-	if err != nil {
+	servicePips, err := getServicesPublicIPs(logger, k8sConfig)
+	if err != nil && !force {
```
orymate: I'd log the error if `force`

stoader: done

PR #1993, opened 2019-05-06
PKE-on-Azure enhancements

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Fixes for issues with the current PKE-on-Azure implementation, mostly for cluster deletion flow.

### Why?
Some steps were missing from the deletion flow. Some related parts were refactored.

### Checklist


internal/cluster/kubernetes/namespace_resources_deleter.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func MakeNamespaceResourcesDeleter(logger logrus.FieldLogger) NamespaceResourcesDeleter {
+	return NamespaceResourcesDeleter{
+		logger: logger,
+	}
+}
+
+func (d NamespaceResourcesDeleter) Delete(k8sConfig []byte, namespace string) error {
+	logger := d.logger.WithField("namespace", namespace)
```
orymate: We should somehow provide cluster information as well. This might be resolved later, and applies to the other deleters too.

internal/cluster/workflow/delete_k8s_resources.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	"github.com/goph/emperror"
+	"go.uber.org/cadence/workflow"
+)
+
+const DeleteK8sResourcesWorkflowName = "delete-k8s-resources"
+
+type DeleteK8sResourcesWorkflowInput struct {
+	K8sConfig []byte
```
orymate: `Force bool` option should be added (later?)

siliconbrain: This workflow doesn't need to know whether the operation is being forced or not, because it only affects how the failure of this workflow is handled by the caller.

internal/cluster/kubernetes/namespace_resources_deleter.go
siliconbrain: Done.

internal/cluster/workflow/delete_k8s_resources.go
orymate: I added a logic to ignore some of the helm step's legitimate errors. This solves my main concern.

PR #2028, opened 2019-05-20
PKE-on-Azure cluster update

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Cluster update/edit feature for PKE-on-Azure clusters.

### Checklist


database/migrations/postgres/1557939759_pke_azure_indices.up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE UNIQUE INDEX idx_azure_pke_np_cluster_id_name ON "azure_pke_node_pools" (cluster_id, "name");
```
orymate: what about deleted and re-created pools with the same name? (I would handle it in a new issue, with the solution that we discussed: generating new pool names server-side instead of the ui)

PR #2056, opened 2019-05-29
quick fix for GKE cluster creation in case of its status is ERROR

### LINKED ISSUE #2055: GKE cluster creation failed due to its status is ERROR
**Describe the bug**
If a GKE cluster status is error after creation, the posthook will start and fail.

Sometimes it happens.

**Expected behavior**
If cluster status was ERROR, cluster creation should return error.

**Additional context**
```
status: ERROR
statusMessage: Failed to create cluster
subnetwork: default
```

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2055

### What's in this PR?
Return error if the cluster status after GKE cluster creation is ERROR 
### Checklist

### DIFF
--- a/cluster/gke.go
+++ b/cluster/gke.go
@@ -61,6 +61,7 @@ import (
 const (
 	statusRunning = "RUNNING"
 	statusDone    = "DONE"
+	statusError   = "ERROR"
 )
 
 const (
@@ -319,6 +320,10 @@ func (c *GKECluster) CreateCluster() error {
 	if err != nil {
 		be := getBanzaiErrorFromError(err)
 		// TODO status code !?
+		return errors.New(be.Message)
+	}
+	if gkeCluster.Status == statusError {
+		be := getBanzaiErrorFromError(err)
 		return errors.New(be.Message)
 	}
 
--- a/cluster/gke.go
+++ b/cluster/gke.go
@@ -323,8 +323,7 @@ func (c *GKECluster) CreateCluster() error {
 		return errors.New(be.Message)
 	}
 	if gkeCluster.Status == statusError {
-		be := getBanzaiErrorFromError(err)
-		return errors.New(be.Message)
+		return errors.New(gkeCluster.StatusMessage)
 	}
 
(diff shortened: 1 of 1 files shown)


cluster/gke.go
```diff
@@ -321,6 +322,10 @@ func (c *GKECluster) CreateCluster() error {
 		// TODO status code !?
 		return errors.New(be.Message)
 	}
+	if gkeCluster.Status == statusError {
+		be := getBanzaiErrorFromError(err)
+		return errors.New(be.Message)
```
stoader: I'd include also the status message provided by `gkeCluster` or if we don't want to display that to the user than log the status message.

cluster/gke.go
```diff
@@ -61,6 +61,7 @@ import (
 const (
 	statusRunning = "RUNNING"
 	statusDone    = "DONE"
+	statusError   = "ERROR"
```
orymate: these are operation statuses. the cluster status can be used from `pkg/cluster.Error`

cluster/gke.go
```diff
@@ -321,6 +322,10 @@ func (c *GKECluster) CreateCluster() error {
 		// TODO status code !?
 		return errors.New(be.Message)
 	}
+	if gkeCluster.Status == statusError {
```
orymate: in which cases will `err == nil` be true (i.e. the function not returning in line 323), but the error extraction below work?

cluster/gke.go
orymate: this line will always be equivalent to `return errors.New("")`, won't it?

pbalogh-sa: @orymate , you are right I will fix it

cluster/gke.go
pbalogh-sa: If getClusterGoogle func return without any error and the gkeCluster.Status is ERROR.
`gkeCluster, err := getClusterGoogle(svc, cc)`

PR #2067, opened 2019-06-04
Label new node pools on cluster update

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Install labels to new node pools created when updating a cluster.

### Checklist


internal/providers/azure/pke/workflow/update_cluster.go
```diff
@@ -251,6 +251,20 @@ func UpdateClusterWorkflow(ctx workflow.Context, input UpdateClusterWorkflowInpu
 			return err
 		}
 	}
+	// install node pool labels
+	{
+		activityInput := cluster.RunPostHookActivityInput{
+			ClusterID: input.ClusterID,
+			HookName:  pkgCluster.InstallNodePoolLabelSetOperator,
```
stoader: This is wrong. We should call `SetupNodePoolLabelsSet` and pass it the appropriate `HookParam`

PR #2069, opened 2019-06-06
Azure PKE tags refactor

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
Use `Tag` struct instead of string pairs.

### Why?
Make working with tags easier.

### Checklist

### DIFF
--- a/internal/providers/azure/pke/workflow/azure.go
+++ b/internal/providers/azure/pke/workflow/azure.go
@@ -50,43 +50,52 @@ func (f *AzureClientFactory) New(organizationID uint, secretID string) (*pkgAzur
 	return cc, nil
 }
 
-func getOwnedTag(clusterName string) (string, string) {
-	return getOwnershipTagKey(clusterName), "owned"
+type Tag struct {
+	Key   string
+	Value string
 }
 
-func getSharedTag(clusterName string) (string, string) {
-	return getOwnershipTagKey(clusterName), "shared"
+func (t Tag) Map() map[string]string {
+	return map[string]string{
+		t.Key: t.Value,
+	}
+}
+
+func getOwnedTag(clusterName string) Tag {
+	return Tag{
+		Key:   getOwnershipTagKey(clusterName),
+		Value: "owned",
+	}
+}
+
+func getSharedTag(clusterName string) Tag {
+	return Tag{
+		Key:   getOwnershipTagKey(clusterName),
+		Value: "shared",
+	}
 }
 
 func getOwnershipTagKey(clusterName string) string {
 	return fmt.Sprintf("kubernetesCluster-%s", clusterName)
 }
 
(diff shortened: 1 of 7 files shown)


internal/providers/azure/pke/workflow/azure.go
```diff
@@ -50,43 +50,54 @@ func (f *AzureClientFactory) New(organizationID uint, secretID string) (*pkgAzur
...
 
 func getOwnershipTagKey(clusterName string) string {
 	return fmt.Sprintf("kubernetesCluster-%s", clusterName)
 }
 
-func tagsFrom(key, value string) map[string]string {
-	return map[string]string{
-		key: value,
+func tagsToMap(tags ...Tag) map[string]string {
+	m := make(map[string]string)
```
bonifaido: Could be `m := make(map[string]string, len(tags))`

internal/providers/azure/pke/workflow/azure.go
```diff
@@ -50,43 +50,54 @@ func (f *AzureClientFactory) New(organizationID uint, secretID string) (*pkgAzur
...
+	return hasTag(tags, getOwnedTag(clusterName))
 }
 
 func HasSharedTag(clusterName string, tags map[string]string) bool {
-	key, val := getSharedTag(clusterName)
-	return hasTag(tags, key, val)
+	return hasTag(tags, getSharedTag(clusterName))
 }
 
 func RemoveSharedTag(tags map[string]string, clusterName string) map[string]string {
```
orymate: shouldn't this either work on a copy of the input map, or have no return value? maps are not like slices in this regard

internal/providers/azure/pke/workflow/azure.go
```diff
@@ -50,43 +50,54 @@ func (f *AzureClientFactory) New(organizationID uint, secretID string) (*pkgAzur
...
 }
 
 func getOwnershipTagKey(clusterName string) string {
 	return fmt.Sprintf("kubernetesCluster-%s", clusterName)
 }
 
-func tagsFrom(key, value string) map[string]string {
-	return map[string]string{
-		key: value,
+func tagsToMap(tags ...Tag) map[string]string {
```
orymate:
```
func (t Tag) Map() map[string]string {
```
would be more legible for all current uses

internal/providers/azure/pke/workflow/azure.go
siliconbrain: returning the modified map makes the function "chainable", e.g.: `tags = RemoveSomeOtherTag(RemoveSharedTag(tags, "my-cluster"), "my-other-key")`

PR #2072, opened 2019-06-13
PKE on AWS: add cloudformation stack failure events to the error mess…

stoader:
…age returned in case of failure

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Collect CloudFormation stack failure events as that provides information on what caused a CloudFormation stack failure and propagate these events into the error that is returned to the user. 

### Why?

When a CloudFormation stack creation fails the user only receives the high-level error message that says that the stack failed but not the root cause. The root cause can be found among the failure events (e.g. quota limit exceeded) linked to the CloudFormation stack, thus these should be propagated back to the user.

### Checklist

### DIFF
--- a/internal/providers/pke/pkeworkflow/delete_vpc_activity.go
+++ b/internal/providers/pke/pkeworkflow/delete_vpc_activity.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
+	pkgCloudformation "github.com/banzaicloud/pipeline/pkg/providers/amazon/cloudformation"
 	"github.com/goph/emperror"
 	"github.com/pkg/errors"
 )
@@ -72,9 +73,6 @@ func (a *DeleteVPCActivity) Execute(ctx context.Context, input DeleteVPCActivity
 	}
 
 	err = cfClient.WaitUntilStackDeleteCompleteWithContext(ctx, &cloudformation.DescribeStacksInput{StackName: &stackName})
-	if err != nil {
-		return emperror.Wrap(err, "waiting for termination")
-	}
 
-	return nil
+	return emperror.Wrap(pkgCloudformation.NewAwsStackFailure(err, stackName, cfClient), "waiting for termination")
 }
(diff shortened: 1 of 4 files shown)


internal/providers/pke/pkeworkflow/delete_vpc_activity.go
```diff
@@ -72,7 +73,7 @@ func (a *DeleteVPCActivity) Execute(ctx context.Context, input DeleteVPCActivity
 	}
 
 	err = cfClient.WaitUntilStackDeleteCompleteWithContext(ctx, &cloudformation.DescribeStacksInput{StackName: &stackName})
-	if err != nil {
+	if err = amazon.NewAwsStackFailure(err, stackName, cfClient); err != nil {
```
orymate: do we want to collect errors even if the deletion was complete? (the err value from line 75 is not checked)

internal/providers/pke/pkeworkflow/wait_for_cf_completion_activity.go
```diff
@@ -50,7 +51,7 @@ func (a *WaitCFCompletionActivity) Execute(ctx context.Context, input WaitCFComp
 	describeStacksInput := &cloudformation.DescribeStacksInput{StackName: aws.String(input.StackID)}
 
 	err = cfClient.WaitUntilStackCreateCompleteWithContext(ctx, describeStacksInput)
-	if err != nil {
+	if err = amazon.NewAwsStackFailure(err, input.StackID, cfClient); err != nil {
```
orymate: do we want to collect errors even if the creation was complete? (the err value from line 54 is not checked)

internal/providers/pke/pkeworkflow/delete_vpc_activity.go
stoader: The `err` is checked in the if statement. See the `; err != nil`. If the original `err` is `nil` then `NewAwsStackFailure` returns `nil`

internal/providers/pke/pkeworkflow/wait_for_cf_completion_activity.go
stoader: The `err` is checked in the if statement. See the `; err != nil`. If the original `err` is `nil` then `NewAwsStackFailure` returns `nil`

internal/providers/amazon/cloudformation.go
```diff
@@ -0,0 +1,98 @@
+// Copyright © 2019 Banzai Cloud
```
siliconbrain: I think this should be under `pkg/providers/amazon` because it is not pipeline specific, it depends only on built-in, 3rd party, and pkg packages.

PR #2087, opened 2019-06-18
Ignore namespace deletion error if namespace status transitioned into…

stoader:
… 'Terminating' state

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

If namespace deletion returns with error but the namespace actually transitions into 'Terminating' status phase than ignore the error and continue with the deletion process.

### Why?

When namespace deletion returns with an error that error may indicate that Kubernetes will delete the namespace at a later stage thus it's not a real error. (e.g. `Operation cannot be fulfilled on namespaces "....": The system is ensuring all content is removed from this namespace. Upon completion, this namespace will automatically be purged by the system.`)

### Checklist

### DIFF
--- a/internal/cluster/kubernetes/user_namespace_deleter.go
+++ b/internal/cluster/kubernetes/user_namespace_deleter.go
@@ -82,8 +82,8 @@ func (d UserNamespaceDeleter) Delete(organizationID uint, clusterName string, k8
 		if err != nil {
 			return emperror.Wrap(err, "could not list remaining namespaces")
 		}
-		left = []string{}
-		gaveUp = []string{}
+		left = nil
+		gaveUp = nil
 		for _, ns := range namespaces.Items {
 			switch ns.Name {
 			case "default", "kube-system", "kube-public":
--- a/internal/cluster/kubernetes/user_namespace_deleter.go
+++ b/internal/cluster/kubernetes/user_namespace_deleter.go
@@ -19,7 +19,7 @@ import (
 	"github.com/goph/emperror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


internal/cluster/kubernetes/user_namespace_deleter.go
```diff
@@ -55,6 +56,17 @@ func (d UserNamespaceDeleter) Delete(organizationID uint, clusterName string, k8
...
 			err := client.CoreV1().Namespaces().Delete(ns.Name, &metav1.DeleteOptions{})
 			if err != nil {
+				// check if the namespace transitioned into 'Terminating' phase, if so
+				// ignore the error
+				namespace, errGet := client.CoreV1().Namespaces().Get(ns.Name, metav1.GetOptions{})
+				if errGet != nil {
+					return emperror.Wrapf(err, "failed to get %q namespace details", ns.Name)
+				}
+
+				if namespace.Status.Phase == corev1.NamespaceTerminating {
```
siliconbrain: I would negate the condition and return with the error inside the `if` instead of the `continue`

internal/cluster/kubernetes/user_namespace_deleter.go
```diff
@@ -71,6 +83,7 @@ func (d UserNamespaceDeleter) Delete(organizationID uint, clusterName string, k8
 			return emperror.Wrap(err, "could not list remaining namespaces")
 		}
 		left = []string{}
+		gaveUp = []string{}
```
siliconbrain: both assignments could be `nil` instead of allocating a zero-length slice

internal/cluster/kubernetes/user_namespace_deleter.go
stoader: If the state of the namespace is 'Terminating' than we can ignore the error it has no sense to retry the delete namespace operation. When the namespace deletion returns with `The system is ensuring all content is removed from this namespace. Upon completion, this namespace will automatically be purged by the system.` error the namespace gets into 'Terminating' state and eventually Kubernetes will delete it.

PR #2117, opened 2019-07-07
Make changes to bucket API to support CLI

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | banzaicloud/banzai-cli#79

### What's in this PR?

Make changes to the openapi descriptor and also to the API itself to support bucket commands in the CLI

### Why?

1. There were some missing information in the openapi descriptor
2. The openapi generator doesn't support pointer structs so cloud provider detection was changed to be based on the secret type as there will be multiple cloud provider properties specified in the incoming request from the CLI

### Checklist


docs/openapi/pipeline.yaml
```diff
@@ -9889,8 +9916,10 @@ components:
                             description: "the secret name of the azure access information"
                             type: string
                             example: "myaccesssecretname"
-                azure:
+                aks:
```
siliconbrain: Is this really for AKS only and not for any distribution running on Azure?

api/bucket.go
```diff
@@ -225,7 +225,19 @@ func CreateBucket(c *gin.Context) {
 		createBucketRequest.SecretId = secret.GenerateSecretIDFromName(createBucketRequest.SecretName)
 	}
 
-	cloudType, err := determineCloudProviderFromRequest(createBucketRequest)
+	retrievedSecret, err := secret.Store.Get(organization.ID, createBucketRequest.SecretId)
+	if err != nil {
+		if strings.Contains(err.Error(), "there's no secret with this id") {
```
siliconbrain: Is there no existing utility method in any of the secret-related packages that does this check? Then there should be!

stoader: I think here you can check for `secret.ErrSecretNotExists` .

docs/openapi/pipeline.yaml
waynz0r: It should be `azure`, but the API currently gives back `aks` and this PR just fixes that in the openapi descriptor. We should address the `aks` -> `azure` transition in another PR.

api/bucket.go
waynz0r: 👍

PR #2119, opened 2019-07-09
dashboardAPI: return clusters regardless of their state

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Return cluster in any state with partial response code.

### Checklist

### DIFF
--- a/internal/dashboard/api.go
+++ b/internal/dashboard/api.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"math"
 	"net/http"
-	"strings"
 
 	"github.com/banzaicloud/pipeline/auth"
 	"github.com/banzaicloud/pipeline/cluster"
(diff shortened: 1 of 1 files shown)


internal/dashboard/api.go
```diff
@@ -89,21 +88,17 @@ func (d *DashboardAPI) GetDashboard(c *gin.Context) {
...
-					logger := d.logger.WithField("clusterId", c.GetID())
-					cluster, partial := d.getClusterDashboardInfo(logger, c, organizationID)
-					if partial {
-						partialResponse = true
-					}
-					clusterResponseChan <- cluster
-				}()
-				i++
-			}
+			go func(commonCluster cluster.CommonCluster) {
```
stoader: Shouldn't we allow filtering by status as well? (e.g. the status to filter by could be passed in as query parameter)

sancyx: for now we said we can show clusters in all statuses, no need to filter

PR #2121, opened 2019-07-10
Integrate federation v2 cross cluster service discovery feature with external DNS

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

- add feature/sync endpoint to be able to recover feature from failure
- integrate federation v2 cross cluster service discovery feature with external DNS
- return 201 & also the error message in case the feature is created but reconcile failed
- delete dns related resource & federated resources when feature is disabled

### Checklist


internal/federation/federation.go
```diff
@@ -103,6 +106,21 @@ func (m *FederationReconciler) init() error {
 	m.Host = m.getHostCluster()
 	m.Members = m.getMemberClusters()
 
+	m.serviceDNSRecordResource = &metav1.APIResource{
+		Group:      "multiclusterdns.kubefed.k8s.io",
```
stoader: I'd make `multiclusterdns.kubefed.k8s.io` const

internal/federation/federation.go
```diff
@@ -103,6 +106,21 @@ func (m *FederationReconciler) init() error {
...
+		Kind:       "ServiceDNSRecord",
+		Version:    "v1alpha1",
```
stoader: I'd make `v1alpha` const

PR #2134, opened 2019-07-18
Error mapping + error handling lib updates, fixes

lpuskas:
- remove deprecated error handling code (emperror)
- map some of the business errors to the appropriate http response codes
- remove typed errors, add "behaviour" to the business error instead
-------

- move feature specific implementation into a separate package in order to delimiot the "framework" and ease the process of adding new features


internal/clusterfeature/clusterfeaturedriver/transport_http.go
```diff
@@ -257,3 +261,16 @@ func isBadRequest(err error) bool {
 
 	return badRequest
 }
+
+func isNotFound(err error) bool {
+	notFound := false
+	errors.UnwrapEach(err, func(err error) bool {
+		if brErr, ok := err.(interface{ NotFound() bool }); ok {
```
siliconbrain: the `br` in `brErr` refers to it being a `BadRequest` error, so here it should be prefixed with `nf` or named using a better, error name independent scheme

lpuskas: thx

PR #2135, opened 2019-07-19
Validate KubeConf secret when added and when used at cluster import

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Validate Kubernetes config secret by trying to query API server version using it.

### Why?

Kubernetes config passed to Pipeline might be syntactically correct while unusable in Pipeline's context for various reasons (e.g. it uses an external auth provider that is not available to Pipeline). The passed Kubernetes config must be verified that Pipeline can use it to interact with Kubernetes cluster references in the config.

### Checklist

### DIFF
--- a/secret/verify/kubernetes.go
+++ b/secret/verify/kubernetes.go
@@ -1,0 +1,55 @@
+// Copyright © 2019 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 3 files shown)


secret/verify/kubernetes.go
```diff
@@ -0,0 +1,60 @@
...
+
+	_, err = clientSet.ServerVersion()
+	if err != nil {
+		return errors.WrapIf(err, "couldn't validate Kubernetes config")
+	}
+
+	return nil
+}
+
+var _ Verifier = (*kubernetesConfigVerifier)(nil)
```
orymate: this is not needed anymore, the use in NewVerifier does the same
```suggestion
```

secret/verify/kubernetes.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	clientSet, err := k8sclient.NewClientFromKubeConfig(kubeConfig)
+	if err != nil {
+		return errors.WrapIf(err, "couldn't get Kubernetes client")
+	}
+
+	_, err = clientSet.ServerVersion()
+	if err != nil {
+		return errors.WrapIf(err, "couldn't validate Kubernetes config")
```
orymate: (you can return here unconditionally)

PR #2157, opened 2019-07-29
Add Stackdriver integration

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR adds Stackdriver integration to Pipeline.

### Why?

This is a temporary solution to help debugging our managed platform.

### Checklist


internal/platform/errorhandler/handler.go
```diff
@@ -0,0 +1,47 @@
...
+	"emperror.dev/emperror"
+	"emperror.dev/errors"
+)
+
+// Handlers collects a number of error handlers into a single one.
+// TODO(sagikazarmark): Use emperror.dev/emperror.Handlers instead.
+type Handlers []emperror.Handler
+
+func (h Handlers) Handle(err error) {
+	if len(h) < 1 {
```
bonifaido: This check is redundant. The next loop won't loop.

internal/platform/errorhandler/handler.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	for _, handler := range h {
+		handler.Handle(err)
+	}
+}
+
+// Close calls Close on the underlying handlers (if there is any closable handler).
+func (h Handlers) Close() error {
+	errs := make([]error, len(h))
```
stoader: if len(h) == 0 you could return with immediately with nil instead of creating a zero-length slice and passing that to errors.Combine()

PR #2164, opened 2019-07-30
Remove unnecessary ipv4 range exclusion in Istio feature

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

We don't need to specifically include k8s clusters' ipv4 ranges to generally allow outside traffic from the Istio mesh, so this PR removes that functionality.

### Checklist


internal/istio/istiofeature/reconcile_remoteistio_cr.go
```diff
@@ -155,9 +149,5 @@ func (m *MeshReconciler) generateRemoteIstioCR(config Config, ipRanges *pkgClust
 		},
 	}
 
-	if config.BypassEgressTraffic {
```
Laci21: Why don't you use something like this here?
```
	if config.BypassEgressTraffic {
		istioConfig.Spec.OutboundTrafficPolicy = v1beta1.OutboundTrafficPolicyConfiguration{
			Mode: "ALLOW_ANY",
		}
	} else {
		istioConfig.Spec.OutboundTrafficPolicy = v1beta1.OutboundTrafficPolicyConfiguration{
			Mode: "REGISTRY_ONLY",
		}
	}
```

waynz0r: The `RemoteIstio` doesn't have those options in the spec.

Laci21: Ah, makes sense then :) Thanks!

Issue #2158, opened 2019-07-29
Convert spotguide sync to Cadence CronWorkflow


sagikazarmark:
https://cadenceworkflow.io/docs/05_cron

PR #2177, opened 2019-08-01
spotguide: scrape shared spotguides in cron workflow

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/banzaicloud/pipeline/issues/2158

### What's in this PR?
Instead of a Go routine shared spotguides are scraped now in a Cadence Cron Workflow.

### Why?
More elegant, and when in running in HA mode - more correct. :)

### Additional context
Currently the first decision task fails in a CronWorkflow, but then it finishes properly.
There is a pending issue in Cadence: uber/cadence#2100

### Checklist

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -209,8 +209,9 @@ func main() {
 		generateCertificatesActivity := pkeworkflow.NewGenerateCertificatesActivity(clusterSecretStore)
 		activity.RegisterWithOptions(generateCertificatesActivity.Execute, activity.RegisterOptions{Name: pkeworkflow.GenerateCertificatesActivityName})
 
-		scrapeSharedSpotguidesWorkflow := spotguide.NewScrapeSharedSpotguidesWorkflow(spotguideManager)
-		workflow.RegisterWithOptions(scrapeSharedSpotguidesWorkflow.Execute, workflow.RegisterOptions{Name: spotguide.ScrapeSharedSpotguidesWorkflowName})
+		scrapeSharedSpotguidesActivity := spotguide.NewScrapeSharedSpotguidesActivity(spotguideManager)
+		workflow.RegisterWithOptions(spotguide.ScrapeSharedSpotguidesWorkflow, workflow.RegisterOptions{Name: spotguide.ScrapeSharedSpotguidesWorkflowName})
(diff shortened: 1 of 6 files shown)


spotguide/sync_spotguides_workflow.go
```diff
@@ -0,0 +1,40 @@
...
+	"go.uber.org/cadence/workflow"
+)
+
+const ScrapeSharedSpotguidesWorkflowName = "scrape-shared-spotguides"
+
+type ScrapeSharedSpotguidesWorkflow struct {
+	manager *SpotguideManager
+}
+
+type ClusterDNSRecordsDeleter interface {
```
sagikazarmark: This looks like garbage

spotguide/spotguide.go
```diff
@@ -198,16 +202,31 @@ func (s *SpotguideManager) isSpotguideReleaseAllowed(release scm.RepositoryRelea
 	return supported && (!prerelease || viper.GetBool(config.SpotguideAllowPrereleases))
 }
 
-func (s *SpotguideManager) ScrapeSharedSpotguides() error {
-	if s.sharedLibraryOrganization == nil {
-		return fmt.Errorf("failed to scrape shared spotguides")
+func (s *SpotguideManager) ScheduleScrapingSharedSpotguides() error {
```
sagikazarmark: Given this is called once from main and doesn't use the manager itself, I don't think it has to be a member of the manager. This could be a plain function, accepting a workflow client. Then the manager also does not have to receive a workflowClient instance in it's constructor.

cmd/worker/main.go
```diff
@@ -173,6 +175,29 @@ func main() {
 		clusterAuthService, err := intClusterAuth.NewDexClusterAuthService(clusterSecretStore)
 		emperror.Panic(errors.Wrap(err, "failed to create DexClusterAuthService"))
 
+		scmProvider := v.GetString("cicd.scm")
```
sagikazarmark: The worker does not use the global configuration values, so you have to define defaults in the `configure` method.

bonifaido: Done.

PR #2182, opened 2019-08-05
Fix non-deterministic map iterations in Cadence workflows

### LINKED ISSUE #2155: Non deterministic code in the UpdateClusterWorklfow
**Describe the bug**
Cadence workflow code must be deterministic to support workflow state recovery through event sourcing. This rules out use of range over map as Go runtime on purpose [randomizes iteration order](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/).

Execute workflow with [sticky execution off](https://github.com/uber-go/cadence-client/blob/103fa1e95de1a892571ccdef71ebca786e4728cb/internal/worker.go#L170).

**Expected behavior**
The code of all workflows (I've looked at UpdateClusterWorklfow only)  should be rewritten to not use range over map.

**Additional context**
Read [Implementation](https://cadenceworkflow.io/docs/03_goclient/02_create_workflows#implementation) section of the Cadence Go Documentation for more potential determinism gotchas.

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2155

### What's in this PR?
Future map structures are converted to slices to ensure iterations are deterministic.

### Additional context

In some cases the implementation itself was affected. Please take that into account when reviewing.

### Checklist

~~- [-] Implementation tested (with at least one cloud provider)~~

### DIFF
--- a/internal/providers/pke/pkeworkflow/update_cluster.go
+++ b/internal/providers/pke/pkeworkflow/update_cluster.go
@@ -83,92 +83,128 @@ func UpdateClusterWorkflow(ctx workflow.Context, input UpdateClusterWorkflowInpu
 	}
 
 	// delete removed nodepools
-	futures := make(map[string]workflow.Future, len(input.NodePoolsToDelete))
-	for _, np := range input.NodePoolsToDelete {
-		if np.Master || !np.Worker {
-			continue
+	{
+		futures := make([]workflow.Future, len(input.NodePoolsToDelete))
+
+		for i, np := range input.NodePoolsToDelete {
+			if np.Master || !np.Worker {
+				continue
+			}
+
+			activityInput := DeletePoolActivityInput{
+				// AWSActivityInput: awsActivityInput,
+				ClusterID: input.ClusterID,
+				Pool:      np,
+			}
+
+			futures[i] = workflow.ExecuteActivity(ctx, DeletePoolActivityName, activityInput)
 		}
 
-		futures[np.Name] = workflow.ExecuteActivity(ctx, DeletePoolActivityName, DeletePoolActivityInput{
-			// AWSActivityInput: awsActivityInput,
-			ClusterID: input.ClusterID,
-			Pool:      np,
-		})
+		errs := make([]error, len(futures))
+		for i, future := range futures {
+			errs[i] = errors.Wrapf(future.Get(ctx, nil), "couldn't delete node pool %q", input.NodePoolsToDelete[i].Name)
+		}
+
+		if err := errors.Combine(errs...); err != nil {
+			return err
+		}
 	}
 
-	for name, future := range futures {
-		err = errors.Append(err, errors.Wrapf(future.Get(ctx, nil), "couldn't delete nodepool %q", name))
-	}
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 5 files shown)


internal/providers/pke/pkeworkflow/update_cluster.go
```diff
@@ -83,92 +83,128 @@ func UpdateClusterWorkflow(ctx workflow.Context, input UpdateClusterWorkflowInpu
...
 		}
 
-		futures[np.Name] = workflow.ExecuteActivity(ctx, DeletePoolActivityName, DeletePoolActivityInput{
-			// AWSActivityInput: awsActivityInput,
-			ClusterID: input.ClusterID,
-			Pool:      np,
-		})
-	}
+		errs := make([]error, len(futures))
+		for i, future := range futures {
```
orymate:
```suggestion
		for i, future := range futures {
		if future != nil {
```

internal/providers/pke/pkeworkflow/update_cluster.go
```diff
@@ -83,92 +83,128 @@ func UpdateClusterWorkflow(ctx workflow.Context, input UpdateClusterWorkflowInpu
...
+			errs[i] = errors.Wrapf(future.Get(ctx, nil), "couldn't delete node pool %q", input.NodePoolsToDelete[i].Name)
```
orymate:
```suggestion
				errs[i] = errors.Wrapf(future.Get(ctx, nil), "couldn't delete node pool %q", input.NodePoolsToDelete[i].Name)
			}
```

internal/providers/pke/pkeworkflow/update_cluster.go
```diff
@@ -83,92 +83,128 @@ func UpdateClusterWorkflow(ctx workflow.Context, input UpdateClusterWorkflowInpu
...
+				ExternalBaseUrlInsecure:   input.PipelineExternalURLInsecure,
+				SSHKeyName:                "pke-ssh-" + input.ClusterName,
+			}
+
+			futures[i] = workflow.ExecuteActivity(ctx, CreateWorkerPoolActivityName, createWorkerPoolActivityInput)
+		}
+
+		errs := make([]error, len(futures))
+		for i, future := range futures {
+			errs[i] = errors.Wrapf(future.Get(ctx, nil), "couldn't create node pool %q", input.NodePoolsToAdd[i].Name)
```
orymate: to be honest, I don't feel this approach better than the append

sagikazarmark: Append dynamically resizes the underlying array, while this approach does not require that.

siliconbrain: Not necessarily. `append` only resizes the slice when its capacity would be exceeded. If the initial capacity is chosen well, no resizing will happen. Also, append will fill the slice without leaving "gaps", while explicit indexed assignment might leave gaps when non-assigning branches are taken (see the nil ptr problem above).

sagikazarmark: In this case not leaving gaps is not an option, because the index is used to access elements in the original slice.

PR #2199, opened 2019-08-08
Support both 'release' and 'app.kubernetes.io/instance' labels to link objects to Helm releases

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Consider beside label 'release' also 'app.kubernetes.io/instance' when trying to determine which helm release a Kubernetes resource belongs to.

### Why?

Most of the helm charts use 'release' label to stamp a Kubernetes resource with the Helm release name. According to  https://helm.sh/docs/chart_best_practices/#standard-labels, the new recommendation is to use 'app.kubernetes.io/instance' thus Pipeline should handle both.

### Checklist

### DIFF
--- a/api/images.go
+++ b/api/images.go
@@ -21,6 +21,7 @@ import (
 
 	apiclient "github.com/banzaicloud/pipeline/client"
 	pkgCommmon "github.com/banzaicloud/pipeline/pkg/common"
+	pkgHelm "github.com/banzaicloud/pipeline/pkg/helm"
 	"github.com/banzaicloud/pipeline/pkg/k8sclient"
 	"github.com/gin-gonic/gin"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


helm/helm.go
```diff
@@ -67,6 +67,11 @@ const SystemNamespace = "kube-system"
 
 const versionAll = "all"
 
+// the label Helm places on Kubernetes objects for differentiating between
+// different instances: https://helm.sh/docs/chart_best_practices/#standard-labels
+const HelmReleaseNameLabelLegacy = "release"
```
sagikazarmark: Can we move this to `pkg/helm` instead? I'd very much like to get rid of this package.

helm/helm.go
```diff
@@ -1070,3 +1075,17 @@ func getChartVersion(v *repo.ChartVersion) (*ChartVersion, error) {
...
+// the legacy label key
+func GetHelmReleaseName(labels map[string]string) string {
+	if labels == nil {
+		return ""
+	}
+
+	if labels[HelmReleaseNameLabel] != "" {
+		return labels[HelmReleaseNameLabel]
+	}
+	return labels[HelmReleaseNameLabelLegacy]
```
sagikazarmark:
```suggestion

	return labels[HelmReleaseNameLabelLegacy]
```

api/images.go
```diff
@@ -82,10 +83,16 @@ func GetDeploymentImages(c *gin.Context) {
 		return
 	}
 
-	selector := fmt.Sprintf("release=%s", release)
+	selector := fmt.Sprintf("%s=%s", helm.HelmReleaseNameLabel, release)
 	log.Infof("Label selector: %s", selector)
 
 	imageList, err := listAllImages(client, selector)
+	if err == nil && len(imageList) == 0 {
```
sagikazarmark: Move this after the error checking

stoader: if this is moved to after the error checking than the error checking must be repeated for the second `listAllImages` call.

sagikazarmark: I don't mind, also, that's an entirely different issue (how errors are handled in API handlers).

"Better clean than clever" This code is unreadable.

sancyx:
👍

PR #2200, opened 2019-08-10
Override klog

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR replaces the global API machinery error handler, which is a klog instance.
It also contains some general improvements for the logger in pipeline's main.

### Why?

Klog tries to write logs to file and ingeniously exits when that fails. When running Pipeline in a scratch container, this is quite an issue.

### DIFF
--- a/config/configuration.go
+++ b/config/configuration.go
@@ -23,9 +23,10 @@ import (
 	"strings"
 	"time"
 
-	"github.com/banzaicloud/pipeline/pkg/helm"
 	"github.com/gin-contrib/cors"
 	"github.com/spf13/viper"
+
+	"github.com/banzaicloud/pipeline/pkg/helm"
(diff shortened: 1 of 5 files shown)


config/configuration.go
```diff
@@ -199,6 +196,10 @@ func init() {
 	viper.AddConfigPath("$PIPELINE_CONFIG_DIR/")
 	viper.SetConfigName("config")
 
+	// Deprecated
```
siliconbrain: should include suggestion about what to use instead

sagikazarmark: Done. See also #2201

PR #2205, opened 2019-08-12
Use middleware on feature endpoints to check cluster and organization

### LINKED ISSUE #2195: Do not allow feature actions in case of wrong organization
**Describe the bug**
I have a GKE cluster with Auto DNS feature. I want to list the feature details (`/pipeline/api/v1/orgs/2/clusters/{clusterID}/features/dns`) i got the right response. After i change the organization (from `colin014`(2) to `banzaicloud`(10)) i still got the right response but the cluser isn't belong to `banzaicloud` organization.

I can do any action (like active & deactivate) on features of cluster

**Expected behavior**
Do not allow any action with feature on clusters which not belongs to the organization. Check the organization with the organization of the cluster.

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2195 

### What's in this PR?
Use middleware on feature endpoints to check cluster and organization. In case of wrong cluster/organization pair the new response:
```json
{
    "type": "about:blank",
    "title": "Not Found",
    "status": 404,
    "detail": "could not get cluster from database: could not find cluster by ID: cluster not found"
}
```

### Checklist


cmd/pipeline/main.go
```diff
@@ -563,16 +564,7 @@ func main() {
...
-				// TODO: move this to a middleware
-				router.Use(func(c *gin.Context) {
-					clusterID, ok := ginutils.UintParam(c, "id")
-					if !ok {
-						return
-					}
-
-					c.Request = c.Request.WithContext(ctxutil.WithClusterID(c.Request.Context(), clusterID))
-				})
+				router.Use(features.NewMiddleware(clusterManager, errorHandler))
```
sagikazarmark: Use this middleware on every cluster endpoint

internal/clusterfeature/features/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		orgID, ok := ginutils.UintParam(c, "orgid")
+		if !ok {
+			return
+		}
+
+		_, err := manager.GetClusterByID(c, orgID, clusterID)
+		if err != nil {
+			errorHandler.Handle(err)
```
sagikazarmark: Only pass the error to the error handler, if it's not a not found error.

internal/clusterfeature/features/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+			problem := problems.NewDetailedProblem(http.StatusNotFound, err.Error())
```
sagikazarmark: The manager might respond with error for various reasons (eg. DB unavailable). Check if the returned error is a not found error (the root cause implements`interface{NotFound() bool}` and it returns true, see other examples in the repo) and return 404 only if that's the case.

colin014:
thx @sagikazarmark i made the requested changes

PR #2208, opened 2019-08-13
Fix UUID resource label

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2193

### What's in this PR?

Fixes and unifies Pipeline UUID resource label name.

### Why?

Cloud providers support resource labeling with different restrictions.

### DIFF
--- a/internal/providers/azure/uuid.go
+++ b/internal/providers/azure/uuid.go
@@ -18,11 +18,9 @@ import (
 	"github.com/banzaicloud/pipeline/internal/global"
 )
 
-const managedByPipelineTag = "io.banzaicloud.pipeline.uuid"
-
 // PipelineTags returns resource tags for Azure based on the pipeline uuid if available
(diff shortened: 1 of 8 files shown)


internal/providers/azure/uuid.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/banzaicloud/pipeline/internal/global"
 )
 
-const managedByPipelineTag = "io.banzaicloud.pipeline.uuid"
+const managedByPipelineTag = "banzaicloud-pipeline-uuid"
```
siliconbrain: if now they are the same, Azure could also use the global UUID tag name

sagikazarmark: True dat

PR #2215, opened 2019-08-14
Initial GitHub Actions config

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

GitHub actions for Pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       mysql:
           image: mysql:5.7
           ports:
-            - 127.0.0.1:3306:3306
+            - 127.0.0.1:13306:3306
           env:
               MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
       postgres:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,132 @@
...
+    runs-on: ubuntu-16.04
+    services:
+      mysql:
+          image: mysql:5.7
+          ports:
+            - 13306:3306
+          env:
+              MYSQL_ALLOW_EMPTY_PASSWORD: "yes"
+      postgres:
+          image: postgres:9.6
```
chrispat: postgres does not define a health check so you have to do that yourself to get the container to work properly.  Here is an example of a couple of different usages https://github.com/actions/example-services/blob/master/.github/workflows/postgres-service.yml

sagikazarmark: Cool, thanks!

PR #2242, opened 2019-08-26
Build docker image with CircleCI

asdwsda:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets |  -

### What's in this PR?
Replace Docker Hub's auto build.

```
master -> banzaicloud/pipeline:latest
tag    -> banzaicloud/pipeline:<tag>
branch -> only build
```

### Why?
To make it faster.

### ToDo

- [x] Setup docker login/password env variables in circleci
- [x] Disable auto build on Docker Hub

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  docker: circleci/docker@0.5.13
+
 jobs:
     build:
         resource_class: large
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -236,7 +239,38 @@ jobs:
...
+                    deploy: false
+                    extra_build_args: '--build-arg GOPROXY=https://proxy.golang.org'
+                    image: banzaicloud/pipeline
+                    tag: $CIRCLE_BRANCH
+                    filters:
+                        branches:
+                            ignore: master
+            -
+                docker/publish:
+                    deploy: false
```
sagikazarmark: Shouldn't this be true instead?

.circleci/config.yml
```diff
@@ -236,7 +239,38 @@ jobs:
...
+                    image: banzaicloud/pipeline
+                    tag: latest
+                    requires:
+                      - build
+                    filters:
+                        branches:
+                            only: master
+            -
+                docker/publish:
+                    deploy: false
```
sagikazarmark: Shouldn't this be true instead?

.circleci/config.yml
asdwsda: Must be true, but it was just the first test build.

orymate:
From @bonifaido: tag the latest release as `latest`, and build `master` as `master`.

.circleci/config.yml
```diff
@@ -235,8 +238,88 @@ jobs:
...
+                    filters:
+                        branches:
+                            ignore: master
+            -
+                docker/publish:
+                    context: dockerhub
+                    deploy: true
+                    extra_build_args: '--build-arg GOPROXY=https://proxy.golang.org'
+                    image: banzaicloud/pipeline
+                    tag: latest
```
sagikazarmark: This should be `master` I guess.

asdwsda: fixed, thanks

PR #2252, opened 2019-08-29
tag cluster resources with "unknown" if uuid not configured

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no

### What's in this PR?
Fall back to tag cloud resources with "unknown" if the pipeline uuid is not configured.

### Why?
If `pipeline.uuid` setting is missing, pipeline tried to tag cloud resources with empty tags. This causes problems at some cloud providers, while the presence of the pipeline tag in itself is valuable too.

### Checklist


internal/global/uuid.go
```diff
@@ -28,3 +28,12 @@ const ManagedByPipelineTag = "banzaicloud-pipeline-uuid"
 func PipelineUUID() string {
 	return viper.GetString(config.PipelineUUID)
 }
+
+// AlwaysPipelineUUID returns PipelineUUID or "unknown" if none set
+func AlwaysPipelineUUID() string {
```
bonifaido: Can't we just set the default in viper to "unknown"?

orymate: i'd like to make clear that this is not an uuid, but a fallback for cases when a non-empty value is mandatory

PR #2258, opened 2019-09-02
dex: use the same db locally as other services

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | `$PIPELINE_DIR/.docker/volumes/dex`
| Related tickets | N/A

### What's in this PR?
Use MySQL by default from local Dex, and added Postgres example.

### Why?
To use and browse the same DB as for other services.

### Additional context

You are most probably using the SQLite backend currently. If you want to use the MySQL or Postgres backend right now, you have to create the `dex` database manually (if you don't want to loose your date), or

```bash
make down
rm config/dex.yml
make up
```

### Checklist

- [x] Implementation tested locally


bonifaido:
Probably I have to put back `restart: on-failure` to Dex to overcome those situations.

PR #2267, opened 2019-09-04
Add k8s roles and bindings

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add some basic roles and rolebindings as initial manifest.


sagikazarmark:
I think we agreed to have one file to avoid ordering problems. Also, the feature's name should be something like "init manifest" or "kube init", since it's not limited to roles.

PR #2269, opened 2019-09-04
Go 1.13

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | supersedes #2232, causes #2271

### What's in this PR?

Upgrades Go version to 1.13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
                     name: Install build dependencies
                     command: |
                       sudo apt-get update
-                      sudo apt-get install -y mysql-client mysql-utilities postgresql-client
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -8,11 +8,10 @@ jobs:
         resource_class: large
         docker:
             -
-                image: circleci/golang:1.12-stretch
+                image: banzaicloud/golang:1.13
```
bonifaido: Why not using the circleci image?

sagikazarmark: We use mysqldiff, which is only available in stretch/bionic, but not in later distros.

PR #2274, opened 2019-09-04
Rewrite organization synchronization

### LINKED ISSUE #2124: Org sync should remove inactive or nonexistent org memberships
**Describe the bug**
Right now not only active but inactive/pending or nonexistent org memberships are also listed after syncing organizations from the github authentication provider. Other providers may also be affected.

To reproduce the issue one has to be an inactive or pending member of an organization that can be achived by sending an invite to the user without accepting it.

**Expected behavior**
I would expect to remove all orgs that the user is not an active member of.

### LINKED ISSUE #2275: Drop GitHub ID column from organizations
Not used anymore

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2124, fixes #2275

### What's in this PR?

This PR rewrites the current organization synchronization processes.

### Why?

With the new authorization feature org membership management will be managed by the OIDC provider.

### Additional context

- It relies on refresh tokens, so chances are it won't work with SAML.

### Checklist


auth/authn.go
```diff
@@ -621,31 +615,7 @@ func (h *banzaiDeregisterHandler) handler(context *auth.Context) {
 
 	db := context.GetDB(context.Request)
 
-	userAdminOrganizations := []UserOrganization{}
-
-	// Query the organizations where the only admin is the current user.
-	sql :=
-		`SELECT * FROM user_organizations WHERE role = ? AND organization_id IN
```
bonifaido: I will really miss this SQL. 😢 🙂

PR #2276, opened 2019-09-05
Introduce Vault feature

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Introduce new Vault secrets webhook feature

### Checklist


cmd/pipeline/main.go
```diff
@@ -543,12 +544,16 @@ func main() {
 
 			// Cluster Feature API
 			{
+				vaultClient, err := commonadapter.NewVaultClient()
```
bonifaido: I think we should use `NewClientFromConfig` instead.

bonifaido: https://github.com/banzaicloud/bank-vaults/blob/master/pkg/sdk/vault/client.go#L109

colin014:
I will open a new PR. In this one the `Build docker image` check cannot build successfully because of the name of the branch.

PR #2318, opened 2019-09-19
Pke ami changes

kozmagabor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
- Bump PKE tool version to 0.4.14
- Update related ami images
- Add EKS ami image query script documentation


docs/developer.md
```diff
@@ -163,6 +163,23 @@ Creating and using EKS clusters requires to you to have the [AWS IAM Authenticat
...
+
+```
+K8S_VERSIONS=(
+  "1.11"
+  "1.12"
+  "1.13"
+  "1.14"
+)
+
+for version in ${K8S_VERSIONS[@]}; do
```
stoader: In the past, this type of query returned also images that were not available for use yet. Our experience was that only those images are available for use that are published here: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html, though thins may have changed since then.

PR #2320, opened 2019-09-19
EKS: delete unused ENIs that were not released by EKS CNI plugin during cluster delete operation

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Search for network interfaces that were created by EKS CNI plugin and are no longer "in-use" and delete them prior starting to delete subnets and vpc at cluster deletion.

### Why?

EKS CNI plugin allocates native AWS VPC IPs addresses for PODs. Behind the scenes what happens is that multiple network interfaces (ENIs) are created and attached to the nodes with multiple pre-allocated IPs. Pods running on the node will get an IP from these.
When pods and nodes are deleted the ENIs of which IPs are not used any more are deleted. However, it may happen that some of these ENIs are left behind (detached state) holding a reference to the subnets they are in. This blocks cluster deletion as subnets can not be deleted until these ENIs are not cleaned up.

### Checklist

### DIFF
--- a/pkg/providers/amazon/ec2/network.go
+++ b/pkg/providers/amazon/ec2/network.go
@@ -15,7 +15,10 @@
 package ec2
 
 import (
+	"fmt"
+
 	"emperror.dev/emperror"
+	"emperror.dev/errors"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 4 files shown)


pkg/providers/amazon/ec2/network.go
```diff
@@ -213,3 +216,80 @@ func (svc *NetworkSvc) GetSubnetsById(subnetIds []string) ([]*ec2.Subnet, error)
...
+
+	if aerr, ok := err.(awserr.Error); ok {
+		switch aerr.Code() {
+		case "InvalidNetworkInterfaceID.NotFound":
+			logger.Info("network interface not found")
+			return nil
+		}
+	}
+
+	return errors.WrapIff(err, "couldn't delete network interface %s", nicId)
```
sagikazarmark: Maybe use `WrapIfWithDetails`?

stoader: I used `WrapIff` here to report back to the user the id of the network interface.

PR #2386, opened 2019-10-16
Add Vault and Monitoring features to capabilities

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| License         | Apache 2.0

### What's in this PR?
Add `Vault` and `Monitoring` features to capabilities

### Why?
To support disabling features and their components

### Checklist

- [x] Implementation tested (with at least one cloud provider)
- [x] Error handling code meets the [guideline](https://github.com/banzaicloud/pipeline/blob/master/docs/error-handling-guide.md)
- [x] Logging code meets the guideline (TODO)

PR #2387, opened 2019-10-16
Add vault and monitoring feature configuration

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2386

### What's in this PR?
Add `Vault` and `Monitoring` features to capabilities

### Why?
To support disabling features and their components

### Checklist


internal/clusterfeature/features/vault/manager.go
```diff
@@ -127,6 +130,13 @@ func (m FeatureManager) ValidateSpec(ctx context.Context, spec clusterfeature.Fe
 		return err
 	}
 
+	if !m.isManagedEnabled && !vaultSpec.CustomVault.Enabled {
+		return clusterfeature.InvalidFeatureSpecError{
+			FeatureName: featureName,
+			Problem:     "Pipeline's Vault configuration is not supported",
```
siliconbrain: I think this is not entirely correct, a little complicated and not very helpful. I'd rephrase it to something like this:
```suggestion
			Problem:     "Pipeline's managed Vault service is not available, configure a custom Vault instance",
```

PR #2391, opened 2019-10-16
Bump anchore-policy-validator version to the latest

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

This PR fixes version of the anchore-policy-validator chart.


pbalogh-sa:
The version of helm charts should be in config or we should use the latest one.

PR #2394, opened 2019-10-17
Add new Slack and PagerDuty secret types

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add 2 new secret types: `slack` and `pagerduty`.

Slack fields
---

| Field name | Description
|--------|---
| apiUrl | Slack url to send alerts

PagerDuty fields
---

| Field name | Description
|--------|---
| integrationKey | The PagerDuty integration key

### Why?
In the `Monitoring` feature we want to use `slack` and `pagerduty` secrets.

### Checklist


pkg/secret/secret.go
```diff
@@ -343,6 +360,19 @@ var DefaultRules = map[string]Meta{
 			{Name: VaultToken, Required: true, Opaque: true, Description: "Token for Vault"},
 		},
 	},
+	SlackSecretType: {
+		Fields: []FieldMeta{
+			{Name: SlackApiUrl, Required: true, Opaque: true, Description: "Slack url to send alerts"},
```
sagikazarmark:
```suggestion
			{Name: SlackApiUrl, Required: true, Opaque: true, Description: "Slack URL to send alerts to"},
```

pkg/secret/secret.go
```diff
@@ -343,6 +360,19 @@ var DefaultRules = map[string]Meta{
...
+			{Name: SlackChannel, Required: true, Opaque: true, Description: "Channel name for the alerts"},
```
sagikazarmark: Is this supposed to be in the secret? IMHO this is rather a feature setting. (Eg. use the same secret, but separate channels for separate clusters)

pkg/secret/secret.go
```diff
@@ -343,6 +360,19 @@ var DefaultRules = map[string]Meta{
...
+		},
+	},
+	PagerDutySecretType: {
+		Fields: []FieldMeta{
+			{Name: PagerDutyUrl, Required: true, Opaque: true, Description: "PagerDuty service endpoint"},
+			{Name: PagerDutyRoutingKey, Required: true, Opaque: true, Description: "The PagerDuty integration key (when using PagerDuty integration type 'Events API v2')."},
```
sagikazarmark: Not sure if this should be in a secret either.

colin014:
I updated the description and the PR on the basis of the discussions

PR #2395, opened 2019-10-17
Cluster feature adjustments

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
1. When listing cluster features, only the features' status is returned
1. When querying an inactive feature, instead of a 404 response, a feature with inactive status is returned

### Why?
1. Listing features took too long because for every feature their output had to be queried.
1. This was the original design. 404 should mean that there is no such supported feature.

### Checklist


internal/clusterfeature/service.go
```diff
@@ -84,18 +73,25 @@ func (s FeatureService) Details(ctx context.Context, clusterID uint, featureName
 
 	// TODO: check cluster ID?
 
-	logger.Debug("retrieving feature from repository")
-	feature, err := s.featureRepository.GetFeature(ctx, clusterID, featureName)
+	logger.Debug("retrieving feature manager")
+	featureManager, err := s.featureManagerRegistry.GetFeatureManager(featureName)
 	if err != nil {
-		const msg = "failed to retrieve feature from repository"
+		const msg = "failed to retrieve feature manager"
```
sagikazarmark: What's the added value of a constant here?

siliconbrain: It's defined once and used by both the debug log and the error wrapping.

sagikazarmark: Yeah...it should be gone. But not in this PR. I don't see how this is useful other than being a noise.

PR #2407, opened 2019-10-22
EKS: APIServerAccessPoints support

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### Checklist

### DIFF
--- a/model/cluster.go
+++ b/model/cluster.go
@@ -194,20 +194,25 @@ type EKSClusterModel struct {
 	ClusterRoleId      string
 	NodeInstanceRoleId string
 
-	LogTypes SQLStringArray `sql:"type:json"`
+	LogTypes EKSLogTypes `sql:"type:json"`
 
-	APIServerAccessPoints SQLStringArray `sql:"type:json"`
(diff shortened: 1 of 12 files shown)


model/cluster.go
```diff
@@ -194,18 +194,20 @@ type EKSClusterModel struct {
 	ClusterRoleId      string
 	NodeInstanceRoleId string
 
-	LogTypes EKSLogTypes `sql:"type:json"`
+	LogTypes SQLStringArray `sql:"type:json"`
```
siliconbrain: Assigning the same type to different things is a bad practice. I'd rather you implement the common type once and create aliases for the different use cases. (Makes refactoring, e.g. extending/changing just one of the types, a whole lot easier.) Also, I think `JSONStringArray` would better describe the type than `SQLStringArray`.
```go
type JSONStringArray []string
type EKSLogTypes = JSONStringArray
type APIServerAccessPoints = JSONStringArray
```

bonifaido: Thanks, fixed, I agree!

PR #2418, opened 2019-10-25
support for proxying security calls sent through the `anchore` path, while old endpoints behaviour not changed 

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    |yes
| API breaks?     |yes
| Deprecations?   |yes

### What's in this PR?
Add cluster uuid to the generated anchore username to avoid conflicts

### Why?
Names generated using clusterID resulted in errors

### Additional context
The "proxy-approach" will be added in a new PR

### Checklist

- [X] User guide and development docs updated (if needed)


internal/security/clusteradapter/cluster_service.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+
+	"emperror.dev/errors"
+
+	"github.com/banzaicloud/pipeline/cluster"
+)
+
+//go:generate mockery -name ClusterGetter -inpkg
+// ClusterGetter restricts the external dependencies for the repository
+type ClusterGetter interface {
```
siliconbrain: This is never used outside of here, I'd pass the `CommonClusterGetter` directly to the cluster service implementation and use it there. That way, you could get rid of `ClusterGetter`, `MakeClusterGetter`, `clusterGetterAdapter`, `ClusterService` and `Cluster`. `NewClusterService` could return an `AnchoreClusterService` which could be renamed to just `ClusterService`.

lpuskas: thx, fixed

PR #2427, opened 2019-10-28
deactivation succeeds when (legacy) resources could not be deleted

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
feature deactivation is considered succeeded when legacy resources (eg.: generated anchore users)
, security scan related namespace labels) can't be removed

### Why?
security scan feature removal failed in the above cases

### Checklist


internal/clusterfeature/features/securityscan/operator.go
```diff
@@ -173,7 +173,11 @@ func (op FeatureOperator) Deactivate(ctx context.Context, clusterID uint, spec c
 	}
 
 	if err := op.namespaceService.RemoveLabels(ctx, clusterID, boundSpec.WebhookConfig.Namespaces, []string{"scan"}); err != nil {
-		return errors.WrapIf(err, "failed to delete namespace labels")
+
+		// if the operation fails due some reason (eg. inexistent namespaces) we notice that and let the deactivation succeed
```
siliconbrain:
```suggestion
		// if the operation fails for some reason (eg. non-existent namespaces) we notice that and let the deactivation succeed
```

PR #2439, opened 2019-10-31
Convert configuration to yaml

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Converts toml config to yaml

### Why?

We use yaml everywhere else, TOML was the reference configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ MGA_VERSION = 0.0.10
 GOLANG_VERSION = 1.13
 
 .PHONY: up
-up: etc/config/dex.yml config/ui/feature-set.json start config/config.toml ## Set up the development environment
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -71,8 +71,8 @@ start: docker-compose.override.yml ## Start docker development environment
 stop: ## Stop docker development environment
 	docker-compose stop
 
-config/config.toml:
-	cp config/config.toml.dist config/config.toml
+config/config.yaml:
+	cp config/config.yaml.dist config/config.yaml
```
orymate:
```suggestion
	test ! -e config/config.toml
	cp config/config.yaml.dist config/config.yaml
```

PR #2474, opened 2019-11-08
Config improvements

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

- Remove old toml config
- Create separate dev config for local development


config/config.yaml.dist
```diff
@@ -54,39 +51,40 @@ database:
...
 #    domain: "pipeline"
-    createNonexistentDomain: true
+#    createNonexistentDomain: false
 #    workflowExecutionRetentionPeriodInDays: 3
 
-cors:
-    allowAllOrigins: false
-    allowOrigins: ["http://localhost:4200"]
+#cors:
+#    allowAllOrigins: true
```
orymate: Please, add a note about the risks of running even a development instance with this option left true.

sagikazarmark: Thanks!

PR #2493, opened 2019-11-13
endpoint for listing namespaces for a cluster

lpuskas:
* The PR contains the implementation of retrieving namespaces for a given cluster (this is required by pipeline clients to facilitate eg.: feature activation)

* the new resource has been added to the open API specification

### DIFF
--- a/apis/pipeline/pipeline.yaml
+++ b/apis/pipeline/pipeline.yaml
@@ -4496,6 +4496,58 @@ paths:
                 500:
                     $ref: '#/components/responses/InternalServerError'
 
+    '/api/v1/orgs/{orgId}/clusters/{id}/namespaces':
+        get:
+            security:
+                -
+                    bearerAuth: []
+            tags:
+                - clusters
+            summary: Lists namespaces for a cluster
+            description: Lists namespaces for a cluster
+            operationId: ListNamespaces
+            parameters:
+                -
+                    name: orgId
+                    in: path
+                    required: true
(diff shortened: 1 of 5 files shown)


apis/pipeline/pipeline.yaml
```diff
@@ -4496,6 +4496,58 @@ paths:
...
+                    required: true
+                    schema:
+                        type: integer
+            responses:
+                '200':
+                    description: Namespaces listed
+                    content:
+                        application/json:
+                            schema:
+                                type: array
```
sagikazarmark: Returning an array directly is generally not a good idea.

lpuskas: thx, fixed.

cmd/pipeline/main.go
```diff
@@ -830,7 +830,8 @@ func main() {
 			cRouter.POST("/nodepools/labels", nplsApi.SetNodepoolLabelSets)
 
 			namespaceAPI := namespace.NewAPI(commonClusterGetter, errorHandler)
-			namespaceAPI.RegisterRoutes(cRouter.Group("/namespaces/:namespace"))
+			cRouter.DELETE("/namespaces/:namespace", namespaceAPI.Delete)
```
siliconbrain: Why not use `RegisterRoutes`?

lpuskas: thx,fixed

PR #2506, opened 2019-11-15
[Monitoring] Set hosts fields only in case of Loki ingress

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Set hosts fields only in case of Loki ingress

### Why?
After this change we can setup properly the ingress path of Loki

### Checklist


internal/clusterfeature/features/logging/operator.go
```diff
@@ -555,3 +556,23 @@ func mergeValuesWithConfig(chartValues interface{}, configValues interface{}) ([
...
+	if (isBasePathSuffix && !isPathPrefix) || (isPathPrefix && !isBasePathSuffix) {
+		return fmt.Sprintf("%s%s", basePath, path)
+	}
+
+	if isBasePathSuffix && isPathPrefix {
+		basePath = strings.TrimSuffix(basePath, "/")
+		path = strings.TrimPrefix(path, "/")
+	}
+
+	return fmt.Sprintf("%s/%s", basePath, path)
```
siliconbrain: `path.Join(basePath, path)`?

colin014: thx

PR #2525, opened 2019-11-20
Relocate set admin rights

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Remove SetupPrivileges posthook and relocate the functionality to legacy cluster create workflow.

### Why?
Cluster setup is the new place for post-create actions.

### Additional context
This functionality is only used by oracle, so some methods have been removed from the CommonCluster interface.


internal/cluster/clusteradapter/client_factory.go
```diff
@@ -27,6 +28,39 @@ import (
...
+}
+
+// FromSecret creates a Kubernetes client for a cluster from a secret.
+func (f ClientFactory) FromSecret(ctx context.Context, secretID string) (kubernetes.Interface, error) {
+	values, err := f.secretStore.GetSecretValues(ctx, secretID)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: better secret parsing?
```
bonifaido: Would be nice to have a common parser for this (not related to this PR), I think we have this code snippet at least 8 different places.

sagikazarmark: Agreed 🙂

sagikazarmark: My idea was to have secret specific access interfaces which expose a secret in a denormalized form.

bonifaido: Great!

PR #2532, opened 2019-11-21
[Logging] Add service labels to output definitions and flow

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add service labels to output definitions and flow

### Why?
In case of service update now we delete all previously created `ClusterOutput`-s and `ClusterFlow`. With this change we can granted to delete only those `ClusterOutput`-s and `ClusterFlow` that we created with Pipeline

### Checklist

### DIFF
--- a/internal/clusterfeature/features/logging/common.go
+++ b/internal/clusterfeature/features/logging/common.go
@@ -50,7 +50,7 @@ const (
 
 	lokiOutputDefinitionName = "loki-output"
 	flowResourceName         = "banzai-logging-flow"
-	resourceLabelKey         = "service"
+	resourceLabelKey         = "banzaicloud.io/service"
 )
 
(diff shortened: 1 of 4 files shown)


internal/clusterfeature/features/logging/common.go
```diff
@@ -50,6 +50,7 @@ const (
 
 	lokiOutputDefinitionName = "loki-output"
 	flowResourceName         = "banzai-logging-flow"
+	resourceLabelKey         = "service"
```
pepov: I would recommend using a qualified key here, something like `banzaicloud.io/service`

colin014: thx @pepov , i updated the key

PR #2540, opened 2019-11-26
Relocate label nodes posthook

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Relocate label nodes posthook to legacy cluster create workflow

### Why?
Posthooks are deprecated

### Additional context
It's moved to legacy cluster create instead of cluster setup, because it's limited to AKS and ACK.

After https://github.com/Azure/AKS/issues/1088 it will be ACK only.

### Checklist

### DIFF
--- a/cluster/create_workflow.go
+++ b/cluster/create_workflow.go
@@ -16,6 +16,8 @@ package cluster
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"go.uber.org/cadence"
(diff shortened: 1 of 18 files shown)


cluster/manager_common_updater.go
```diff
@@ -168,8 +174,72 @@ func (c *commonUpdater) Update(ctx context.Context) error {
...
+		return err
+	}
+
+	log.Debug("get K8S connection")
+	client, err := k8sclient.NewClientFromKubeConfig(kubeConfig)
+	if err != nil {
+		return err
+	}
+
+	nodeNameLister, ok := commonCluster.(nodeNameLister)
```
Ecsy: This should be switched with Kubernetes client creation. Interface check is more lightweight, it should fail fast.

cluster/create_workflow.go
```diff
@@ -265,3 +293,78 @@ func (a SetupPrivilegesActivity) Execute(ctx context.Context, input SetupPrivile
...
+		logger.Debug("labeling nodepool")
+
+		for _, nodeName := range nodes {
+			logger := logger.With("node", nodeName)
+			logger.Debug("labeling node")
+			labels := map[string]string{pkgCommon.LabelKey: poolName}
+
+			tokens := make([]string, 0, len(labels))
+			for k, v := range labels {
+				tokens = append(tokens, "\""+k+"\":\""+v+"\"")
```
Ecsy: Are we sure `v` should not be sanitized first?

sagikazarmark: This code is copied without changes.

PR #2575, opened 2019-12-02
Make calico the default Kubernetes network provider for PKE

stoader:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Make `calico` as the default Kubernetes network provider for PKE.

### Checklist


internal/providers/pke/network.go
```diff
@@ -55,9 +55,7 @@ func (n Network) String() string {
 // NetworkProvider is the schema for the DB.
 type NetworkProvider string
 
-const (
-	NPWeave NetworkProvider = "weave" // Weave network provider.
-)
+var _ driver.Valuer = (*NetworkProvider)(nil)
```
siliconbrain: Should check in a test instead, like https://github.com/banzaicloud/pipeline/blob/d1c66e00714faf22a14525143cc48341fed126b0/internal/providers/pke/config_test.go

PR #2578, opened 2019-12-03
Secret refactors: Part 2

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | Fixes #2549

### What's in this PR?
A second batch of secret related refactors.

### Why?
Move closer to eradicating `src/secret` and global secret store references.

### Checklist


cmd/pipeline/main.go
```diff
@@ -681,7 +681,6 @@ func main() {
 				featureRepository := clusterfeatureadapter.NewGormFeatureRepository(db, logger)
 				clusterGetter := clusterfeatureadapter.MakeClusterGetter(clusterManager)
 				clusterPropertyGetter := dnsadapter.NewClusterPropertyGetter(clusterManager)
-				secretStore := commonadapter.NewSecretStore(secret.Store, commonadapter.OrgIDContextExtractorFunc(auth.GetCurrentOrganizationID))
```
sagikazarmark: I have a feeling that this will break the latest master

siliconbrain: How so?

pkg/secret/password.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return "", errors.WrapIf(err, "failed to generate index")
+		}
+		r := alphabet[idx]
+		b.WriteRune(r)
+	}
+	return b.String(), nil
+}
+
+func (g PasswordGenerator) generateIndex(limit int) (int, error) {
```
orymate: https://golang.org/pkg/crypto/rand/#Int this would be theoretically better because `byte % limit` won't be uniform except for limits of powers of 2

PR #2581, opened 2019-12-04
Add node pool deletion api

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
A new API for deleting node pools.

### Why?
The current cluster update API is error prone from a client point of view, so we decided to add a new node pool API that allows a client to interact with one node pool at a time.

### Additional context
For the time being only EKS is supported.

### Checklist


internal/cluster/clusteradapter/node_pool_store.go
```diff
@@ -0,0 +1,135 @@
...
+		return false, err
+	}
+
+	switch {
+	case c.Cloud == providers.Amazon && c.Distribution == "eks":
+		var eksCluster model.EKSClusterModel
+
+		err := s.db.Where(model.EKSClusterModel{ClusterID: clusterID}).First(&eksCluster).Error
+		if gorm.IsRecordNotFoundError(err) {
+			return false, errors.Wrap(err, "cluster model is inconsistent")
```
stoader: I'd make the error message more explicit. Something like "Cluster not found in database"

internal/cluster/clusteradapter/node_pool_store.go
```diff
@@ -0,0 +1,135 @@
...
+			return false, errors.WrapWithDetails(
+				err, "failed to check if node pool exists",
+				"clusterId", clusterID,
+				"nodePoolName", name,
+			)
+		}
+
+		var eksNodePool model.AmazonNodePoolsModel
+
+		err = s.db.Where(model.AmazonNodePoolsModel{ClusterID: eksCluster.ID}).First(&eksNodePool).Error
```
stoader: I wonder if instead of querying the database for the node pool in a separate call it would be better to:
 ```
s.db.Where(model.EKSClusterModel{ClusterID: clusterID}).Preload("NodePools").First(&eksCluster)
``` and than look for the node pool in `eksCluster.NodePools`

File: internal/cluster/clusteradapter/node_pool_store.go [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return err
+	}
+
+	switch {
+	case c.Cloud == providers.Amazon && c.Distribution == "eks":
+		var eksCluster model.EKSClusterModel
+
+		err := s.db.Where(model.EKSClusterModel{ClusterID: clusterID}).First(&eksCluster).Error
+		if gorm.IsRecordNotFoundError(err) {
```
stoader: Do we need to perform this check? Isn't enough to simply delete the node pool db record?

internal/cluster/clusteradapter/store.go
```diff
@@ -0,0 +1,111 @@
...
+		Cloud:          clusterModel.Cloud,
+		Distribution:   clusterModel.Distribution,
+		Location:       clusterModel.Location,
+		SecretID:       brn.New(clusterModel.OrganizationId, brn.SecretResourceType, clusterModel.SecretId),
+		ConfigSecretID: brn.New(clusterModel.OrganizationId, brn.SecretResourceType, clusterModel.ConfigSecretId),
+	}, nil
+}
+
+func (s Store) findModel(ctx context.Context, id uint) (*model.ClusterModel, error) {
+	clusterModel, err := s.clusters.FindOneByID(0, id)
```
stoader: We may need a `FindOneByID(id)` overload as `FindOneByID(0, id)` is a bit confusing

sagikazarmark: Agreed, but that's a problem for another day.

internal/cluster/clusteradapter/node_pool_store.go
sagikazarmark: Unfortnately the ClusterID in amazon node pools is not **the** cluster ID, but the primary key of the eks_clusters table.

internal/cluster/clusteradapter/node_pool_store.go
sagikazarmark: That indeed sounds better, thanks!

internal/cluster/clusterworkflow/workflow_delete_node_pool.go
```diff
@@ -0,0 +1,93 @@
...
+func DeleteNodePoolWorkflow(ctx workflow.Context, input DeleteNodePoolWorkflowInput) error {
+	ao := workflow.ActivityOptions{
+		ScheduleToStartTimeout: 5 * time.Minute,
+		StartToCloseTimeout:    10 * time.Minute,
+		WaitForCancellation:    true,
+		RetryPolicy: &cadence.RetryPolicy{
+			InitialInterval:          15 * time.Second,
+			BackoffCoefficient:       1.0,
+			MaximumAttempts:          30,
+			NonRetriableErrorReasons: []string{_cadence.ClientErrorReason},
```
stoader: imho `cadenceInternal:Panic` should be excluded from retries as that indicates a code error which usually is not an interim error that goes away after a couple of retries.

internal/cluster/clusteradapter/node_pool_store.go
sagikazarmark: This is an internal error that is supposed to signal to us, that there is a cluster record, but no matching record in the eks clusters table.

PR #2614, opened 2019-12-13
Fix error handling of legacy cluster create workflow

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Set error status when the legacy workflow fails, and don't retry config download 30 times when permission errors occur.

### Checklist

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -196,8 +196,9 @@ func main() {
 		}
 		global.SetDB(db)
 
+		clusterRepo := clusteradapter.NewClusters(db)
 		clusterManager := cluster.NewManager(
-			clusteradapter.NewClusters(db),
+			clusterRepo,
 			nil,
 			nil,
 			nil,
@@ -205,6 +206,7 @@ func main() {
 			nil,
 			logrusLogger,
 			errorHandler,
+			clusteradapter.NewStore(db, clusterRepo),
 		)
 		tokenStore := bauth.NewVaultTokenStore("pipeline")
 		tokenManager := pkgAuth.NewTokenManager(
--- a/src/api/secrets.go
+++ b/src/api/secrets.go
@@ -546,7 +546,9 @@ func IsValidSecretType(secretType string) error {
 func checkClustersBeforeDelete(orgId uint, secretId string) error {
 	// TODO: move these to a struct and create them only once upon application init
 	secretValidator := providers.NewSecretValidator(secret.Store)
-	clusterManager := cluster.NewManager(clusteradapter.NewClusters(global.DB()), secretValidator, cluster.NewNopClusterEvents(), nil, nil, nil, log, errorHandler)
+	clusterRepo := clusteradapter.NewClusters(global.DB())
+	clusterStore := clusteradapter.NewStore(global.DB(), clusterRepo)
+	clusterManager := cluster.NewManager(clusterRepo, secretValidator, cluster.NewNopClusterEvents(), nil, nil, nil, log, errorHandler, clusterStore)
 
 	clusters, err := clusterManager.GetClustersBySecretID(context.Background(), orgId, secretId)
 	if err != nil {
--- a/cmd/pipeline/main.go
+++ b/cmd/pipeline/main.go
@@ -370,7 +370,7 @@ func main() {
 		errorHandler.Handle(errors.WrapIf(err, "Failed to configure Cadence client"))
 	}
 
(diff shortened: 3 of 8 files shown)


cmd/worker/main.go
```diff
@@ -196,15 +196,17 @@ func main() {
 		}
 		global.SetDB(db)
 
+		klasters := clusteradapter.NewClusters(db)
 		clusterManager := cluster.NewManager(
-			clusteradapter.NewClusters(db),
+			klasters,
```
sagikazarmark: Is this intentional?

src/api/secrets.go
```diff
@@ -546,7 +546,9 @@ func IsValidSecretType(secretType string) error {
 func checkClustersBeforeDelete(orgId uint, secretId string) error {
 	// TODO: move these to a struct and create them only once upon application init
 	secretValidator := providers.NewSecretValidator(secret.Store)
-	clusterManager := cluster.NewManager(clusteradapter.NewClusters(global.DB()), secretValidator, cluster.NewNopClusterEvents(), nil, nil, nil, log, errorHandler)
+	klusters := clusteradapter.NewClusters(global.DB())
```
sagikazarmark: ??

cmd/worker/main.go
siliconbrain: `clusters` was already in use and naming is hard

src/api/secrets.go
siliconbrain: `clusters` was already in use and naming is hard

cmd/worker/main.go
sagikazarmark: How about not putting it in a variable then?

src/api/secrets.go
sagikazarmark: How about not putting it in a variable then?

cmd/worker/main.go
siliconbrain: Because the value is used in more than one place?

src/api/secrets.go
siliconbrain: Because the value is used in more than one place?

cmd/worker/main.go
stoader: what about clusteradapter.Store interface extend the clusteradapter.Clusters interface? Or adding a method Clusters() to it which exposes the internal `clusters` field.

src/api/secrets.go
stoader: what about clusteradapter.Store interface extend the clusteradapter.Clusters interface? Or adding a method Clusters() to it which exposes the internal `clusters` field.

cmd/worker/main.go
siliconbrain: The `Store` implementation (`struct`!) could be modified to expose the `Clusters`' methods. This way, a `Store` instance could be used in place of `Clusters` instances. Eventually, `Clusters` methods should be "merged" into `Store`, since they do mostly the same things.

sagikazarmark: > what about clusteradapter.Store interface extend the clusteradapter.Clusters interface? Or adding a method Clusters() to it which exposes the internal clusters field.

That would defeat the whole purpose of this interface. 😕

sagikazarmark: > Because the value is used in more than one place?

This is a thin layer on top of the database. I would rather create two instances and refactor later than use half hungarian, half english variable names.

sagikazarmark:
```suggestion
			clusteradapter.NewClusters(db),
```

cmd/worker/main.go
```diff
@@ -196,15 +196,17 @@ func main() {
 		}
 		global.SetDB(db)
 
+		klasters := clusteradapter.NewClusters(db)
```
sagikazarmark:
```suggestion
```

cmd/worker/main.go
```diff
@@ -196,15 +196,17 @@ func main() {
...
-			clusteradapter.NewClusters(db),
+			klasters,
 			nil,
 			nil,
 			nil,
 			nil,
 			nil,
 			logrusLogger,
 			errorHandler,
+			clusteradapter.NewStore(db, klasters),
```
sagikazarmark:
```suggestion
			clusteradapter.NewStore(db, clusteradapter.NewClusters(db)),
```

PR #2616, opened 2019-12-23
fix issue, when labels are not added to new node pools in case there …

sancyx:
…are no custom labels specified

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
- remove POST nodepool-labels API call as it's not used
- fix issue, no labels are present on a newly added node in case there were no custom labels specified in the request

### Checklist


cmd/pipeline/main.go
```diff
@@ -827,7 +827,6 @@ func main() {
 			cgroupsAPI.AddRoutes(orgs.Group("/:orgid/clustergroups"))
 
 			cRouter.GET("/nodepool-labels", nplsApi.GetNodepoolLabelSets)
-			cRouter.POST("/nodepool-labels", nplsApi.SetNodepoolLabelSets)
```
stoader: Is this not used any more?

internal/providers/amazon/eks/driver/cluster_creator.go
```diff
@@ -190,11 +192,24 @@ func (c *EksClusterCreator) create(ctx context.Context, logger logrus.FieldLogge
 			Labels:           np.Labels,
 		}
 		asgList = append(asgList, asg)
+
+		nodePoolLabels = append(nodePoolLabels, cluster.NodePoolLabels{
```
stoader: The `NodePoolLabels ` name is misleading as it seems that beside Labels it holds node pool related fields as well.

src/api/npls.go
```diff
@@ -115,102 +112,3 @@ func (n *NodepoolManagerAPI) GetNodepoolLabelSets(c *gin.Context) {
 
 	c.JSON(http.StatusOK, response)
 }
-
-func (n *NodepoolManagerAPI) SetNodepoolLabelSets(c *gin.Context) {
```
stoader: Isn't this used any more?

sancyx: it not used by UI as labels are updated in cluster update request

cmd/pipeline/main.go
sancyx: it not used by UI as labels are updated in cluster update request

internal/providers/amazon/eks/driver/cluster_creator.go
sancyx: Actually it holds some fields from node pool like name, instance type & spot price, necessary for cloud info call.

src/cluster/nodelabels.go
```diff
@@ -26,17 +26,25 @@ import (
...
 
 const labelFormatRegexp = "[^-A-Za-z0-9_.]"
 
+type NodePoolLabels struct {
+	Name         string
+	Existing     bool
+	InstanceType string            `json:"instanceType,omitempty"`
+	SpotPrice    string            `json:"spotPrice,omitempty"`
+	Preemptible  bool              `json:"preemptible,omitempty"`
+	Labels       map[string]string `json:"labels,omitempty"`
```
stoader: I'd rename this to something like `CustomLabels` or `ExtraLabels` to make it clear that these are user provided labels while the others are set by Pipeline.

src/cluster/nodelabels.go
```diff
@@ -26,17 +26,25 @@ import (
...
 	"github.com/banzaicloud/pipeline/internal/global"
 	pipelineContext "github.com/banzaicloud/pipeline/internal/platform/context"
-	pkgCluster "github.com/banzaicloud/pipeline/pkg/cluster"
 	"github.com/banzaicloud/pipeline/pkg/common"
 )
 
 const labelFormatRegexp = "[^-A-Za-z0-9_.]"
 
+type NodePoolLabels struct {
+	Name         string
```
stoader: `Name` -> `NodePoolName`

sancyx: renamed

src/cluster/nodelabels.go
sancyx: renamed to CustomLabels

PR #2617, opened 2019-12-28
Node pool create

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Adds node pool create API for EKS.

### Checklist

### Testing

1. Build the latest Banzai CLI from master
2. Create a new EKS cluster
2. Use the following node pool create descriptor:

<details><summary>Node pool create</summary><br><pre>
{
    "name": "pool1",
    "labels": {
        "mykey": "myvalue"
    },
    "autoscaling": {
        "enabled": true,
        "minSize": 3,
        "maxSize": 4
    },
    "size": 3,
    "instanceType": "c5.large",
    "spotPrice": "0.04",
    "image": "OPTIONAL IMAGE ID",
    "subnet": {
        "subnetId": "SUBNETID"
    }
}
</pre></details>

Command: `banzai cluster nodepool create --cluster-name=CLUSTER_NAME -f nodepool.json --name OPTIONAL_POOL_NAME_OVERRIDE `


internal/cluster/clusteradapter/node_pool_validator.go
```diff
@@ -0,0 +1,119 @@
...
+// NewDistributionNodePoolValidator returns a new DistributionNodePoolValidator.
+func NewDistributionNodePoolValidator(db *gorm.DB) DistributionNodePoolValidator {
+	return DistributionNodePoolValidator{
+		db: db,
+	}
+}
+
+func (v DistributionNodePoolValidator) Validate(_ context.Context, c cluster.Cluster, rawNodePool cluster.NewRawNodePool) error {
+	switch {
+	case c.Cloud == providers.Amazon && c.Distribution == "eks":
```
stoader:
```suggestion
	case c.Cloud == providers.Amazon && c.Distribution == cluster.EKS:
```

there are constants defined for distributions in `pkg/cluster/base.go`

internal/cluster/clusterworkflow/activity_create_node_pool.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func (a CreateNodePoolActivity) Execute(ctx context.Context, input CreateNodePoolActivityInput) error {
+	c, err := a.clusters.GetCluster(ctx, input.ClusterID)
+	if err != nil {
+		return cadence.WrapClientError(err)
+	}
+
+	switch {
+	case c.Cloud == providers.Amazon && c.Distribution == "eks":
```
stoader:
```suggestion
	case c.Cloud == providers.Amazon && c.Distribution == cluster.EKS:
```

there are constants defined for distributions in `pkg/cluster/base.go`

internal/cluster/clusterworkflow/activity_create_node_pool_label_set.go
```diff
@@ -0,0 +1,106 @@
...
+	c, err := a.clusters.GetCluster(ctx, input.ClusterID)
+	if err != nil {
+		return cadence.WrapClientError(err)
+	}
+
+	var name string
+	var labels map[string]string
+
+	switch {
+	case c.Cloud == providers.Amazon && c.Distribution == "eks":
```
stoader:
```suggestion
	case c.Cloud == providers.Amazon && c.Distribution == cluster.EKS:
```

internal/cluster/clusteradapter/node_pool_validator.go
sagikazarmark: I don't want to import that package here for now.

internal/cluster/node_pool.go
```diff
@@ -45,33 +123,97 @@ type NodePoolStore interface {
...
+	}
+
+	if exists {
+		return errors.WithStack(NodePoolAlreadyExistsError{
+			ClusterID: clusterID,
+			NodePool:  nodePool.Name,
+		})
+	}
+
+	err = s.clusters.SetStatus(ctx, clusterID, Updating, "creating node pool")
```
stoader: I wonder if this should be invoked right after  `s.checkCluster(cluster)` or just before `s.nodePools.NodePoolExists` to ensure there won't be two create node pool flows running in parallel in case user invokes `add node pool` API for the same node pool  within a very short period of time.

internal/cluster/distribution/eks_node_pool.go
```diff
@@ -0,0 +1,73 @@
...
+package distribution
+
+import (
+	"context"
+
+	"github.com/banzaicloud/pipeline/internal/cluster"
+)
+
+// NewEKSNodePool describes Amazon node groups model of a cluster
+type NewEKSNodePool struct {
```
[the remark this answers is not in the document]
sagikazarmark: It is already added to the descriptor.

internal/cluster/node_pool.go
sagikazarmark: There is no atomicity guaranteed, so even then two *very* close calls could end up adding a new node pool.

Status update should only happen once all checks are passed, otherwise we would have to change the status back. If we want additional guarantees, we should introduce a locking mechanism. But as a first version this should be fine, we don't have such guarantees anywhere else in the system.

internal/cluster/clusterworkflow/activity_create_node_pool_label_set.go
sagikazarmark: I don't want to import that package here for now.

internal/cluster/clusterworkflow/activity_create_node_pool.go
sagikazarmark: I don't want to import that package here for now.

internal/cluster/clusterworkflow/workflow_create_node_pool.go
```diff
@@ -0,0 +1,95 @@
...
+		RetryPolicy: &cadence.RetryPolicy{
+			InitialInterval:          15 * time.Second,
+			BackoffCoefficient:       1.0,
+			MaximumAttempts:          30,
+			NonRetriableErrorReasons: []string{_cadence.ClientErrorReason, "cadenceInternal:Panic"},
+		},
+	}
+	_ctx := ctx
+	ctx = workflow.WithActivityOptions(ctx, ao)
+
```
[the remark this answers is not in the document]
sagikazarmark: I don't think it's necessary, especially because this will change in the near future.

internal/cluster/distribution/eks_node_pool.go
```diff
@@ -48,7 +48,7 @@ func (n NewEKSNodePool) Validate() error {
 			violations = append(violations, "minimum autoscaling size cannot be lower than one")
 		}
 
-		if n.Autoscaling.MaxSize < n.Autoscaling.MinSize {
+		if n.Autoscaling.MaxSize <= n.Autoscaling.MinSize {
```
stoader: If Autoscaling.MaxSize equals Autoscaling.MinSize then the cluster autoscaling loses it's function. IMHO if cluster autoscaler is enabled for a node pool then max size >= min size+1

internal/cluster/clusteradapter/node_pool_processor.go
```diff
@@ -0,0 +1,122 @@
...
+			if len(eksCluster.Subnets) == 0 {
+				return rawNodePool, errors.New("cannot resolve subnet")
+			}
+
+			rawNodePool["subnet"] = map[string]interface{}{
+				"subnetId":         *eksCluster.Subnets[0].SubnetId,
+				"cidr":             *eksCluster.Subnets[0].Cidr,
+				"availabilityZone": *eksCluster.Subnets[0].AvailabilityZone,
+			}
+		}
```
stoader: if `nodePool.Subnet.SubnetId != "" ` than we should validate that the provided subnet id is contained by `eksCluster.Subnets`

internal/cluster/distribution/eks_node_pool.go
sagikazarmark: If the maxSize equals (or is less than) minSize validation fails. In other words, validation only passes when max size >= min size+1

internal/cluster/clusteradapter/node_pool_processor.go
sagikazarmark: That's done in the validator: https://github.com/banzaicloud/pipeline/pull/2617/files#diff-001ac5a7d84bdf4c47843ca75f19cf2dR89

PR #2639, opened 2020-01-13
Patch AWS CNI daemonset with additional tags

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Configure the CNI driver to tag ENIs with Pipeline tags by patching the CNI driver daemonset when creating the cluster.

### Why?
ENIs are not cleaned up by the control plane delete, so we need to clean them up once the cluster is deleted. In order to allow customers to restrict IAM access to resources created/managed by Pipeline, we need to tag them. ENIs are created by the CNI driver, so we need to configure it before creating any ASGs during cluster create.

### Additional context
This feature is not released yet. Will be part of the 1.6.0 CNI driver release:

https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.6.0-rc5

Additional resources:

https://eksworkshop.com/beginner/160_advanced-networking/secondary_cidr/configure-cni/

### Testing

1. Create a new cluster
2. Ensure that the CNI daemonset is patched with the proper env var (see bellow): `kubectl -n kube-system get daemonset -o yaml aws-node`
2. Download the following file: https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.6/config/v1.6/aws-k8s-cni.yaml
3. Patch it with environment variables (see bellow)
4. Run `kubectl apply -f aws-k8s-cni.yaml`
5. Scale up the cluster to create new ENIs

Env var patch:

```yaml
        - name: ADDITIONAL_ENI_TAGS
          value: '{"banzaicloud-pipeline-managed":"true","banzaicloud-pipeline-uuid":"my-instance"}'
```

### Checklist

### DIFF
--- a/internal/providers/amazon/eks/workflow/bootstrap_activity.go
+++ b/internal/providers/amazon/eks/workflow/bootstrap_activity.go
@@ -163,6 +163,19 @@ func (a *BootstrapActivity) Execute(ctx context.Context, input BootstrapActivity
 
 	tags := map[string]string{}
 
+	var envVars []v1.EnvVar
+
+	for _, envVar := range ds.Spec.Template.Spec.Containers[0].Env {
+		if envVar.Name == "ADDITIONAL_ENI_TAGS" {
+			// omit invalid JSONs
+			_ = json.Unmarshal([]byte(envVar.Value), &tags)
+
+			continue
+		}
+
+		envVars = append(envVars, envVar)
+	}
+
 	for _, tag := range amazon.PipelineTags() {
 		tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
 	}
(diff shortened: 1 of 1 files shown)


internal/providers/amazon/eks/workflow/bootstrap_activity.go
```diff
@@ -153,6 +156,32 @@ func (a *BootstrapActivity) Execute(ctx context.Context, input BootstrapActivity
...
+	for _, tag := range amazon.PipelineTags() {
+		tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
+	}
+
+	tagBody, err := json.Marshal(tags)
+	if err != nil {
+		return nil, cadence.NewClientError(err)
+	}
+
+	ds.Spec.Template.Spec.Containers[0].Env = append(ds.Spec.Template.Spec.Containers[0].Env, v1.EnvVar{
```
stoader: I think we first should check if `ADDITIONAL_ENI_TAGS` is already set or not. If `ADDITIONAL_ENI_TAGS` already exists we need to merge the Pipeline tags with the existing values under `ADDITIONAL_ENI_TAGS`.

sagikazarmark: The default value is `{}`. In case of a new cluster I doubt this will have any initial values.

PR #2663, opened 2020-01-24
Optionally enable removing credentials from EKS kubeconfigs

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Adds a configuration flag to remove credentials from EKS kubeconfig exposed through the API.

### Why?
To avoid exposing shared secrets accidentally in case the administrator explicitly wants this.

### Checklist

- [x] Logging code meets the guideline

### DIFF
--- a/src/cluster/eks.go
+++ b/src/cluster/eks.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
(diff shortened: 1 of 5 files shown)


src/cluster/eks.go
```diff
@@ -27,6 +27,8 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
+	"github.com/banzaicloud/pipeline/internal/global"
+
```
siliconbrain: organize imports

PR #2667, opened 2020-01-24
Use k8s runtime client in service mesh feature

waynz0r:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

It replaces generated k8s clientsets in the service mesh feature implementation to the more generic k8s runtime client.

It also bumps backyards version to the latest available.

### Checklist


internal/cmd/config.go
```diff
@@ -503,22 +503,30 @@ func Configure(v *viper.Viper, _ *pflag.FlagSet) {
...
-			"tag":        "1.0.4",
+		"application": map[string]interface{}{
+			"image": map[string]interface{}{
+				"repository": "banzaicloud/backyards",
+				"tag":        "1.1.2",
+			},
+		},
+		"web": map[string]interface{}{
+			"image": map[string]interface{}{
+				"repository": "banzaicloud/backyards",
```
Laci21: Isn't this repo `backyards-web`?
```suggestion
				"repository": "banzaicloud/backyards-web",
```

waynz0r: good catch, i fixed it, it wasn't properly used in the code

PR #2685, opened 2020-01-30
Support ingress configuration

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add support for ingress configuration. There're are4 new config keys:
- `cluster::posthook::ingress::enabled` (boolean)
- `cluster::posthook::ingress::chart` (string)
- `cluster::posthook::ingress::version` (string)
- `cluster::posthook::ingress::values` (string)

If `cluster::posthook::ingress::enabled` config key is `false` pipeline doesn't deploy `pipeline-cluster-ingress`

**How to use**
```
    posthook:
        ingress:
            enabled: true
            chart: "banzaicloud-stable/pipeline-cluster-ingress"
            version: "0.0.8"
            values: |
              traefik:
                service:
                  annotations:
                  colin.io/dummy: "banzai"
```

### Checklist

### To Do

- [x] Viper decode hook


config/config.yaml.dist
```diff
@@ -362,6 +362,11 @@ dex:
 #
 #                # See https://github.com/kubernetes-sigs/kubefed/tree/master/charts/kubefed for details
 #                values: {}
+#
+#    posthook:
+#        ingressconrtroller:
```
sagikazarmark:
```suggestion
#        ingresscontroller:
```

sagikazarmark: The configuration should probably follow configuration structure for other features: chart name and version should be configurable.

sagikazarmark: Or just:

```suggestion
#        ingress:
```

internal/cmd/config.go
```diff
@@ -484,6 +488,22 @@ func Configure(v *viper.Viper, _ *pflag.FlagSet) {
...
+				"generateTLS":    true,
+			},
+		},
+	})
+	if err != nil {
+		// todo (colin): panic??
+		panic(err)
+	}
+	v.SetDefault("cluster::posthook::ingresscontroller::enabled", true)
+	v.SetDefault("cluster::posthook::ingresscontroller::values", defaultIngressControllerValues)
```
sagikazarmark: This should probably become a string as well. Viper does not play well with custom types.

src/cluster/hookfunctions.go
```diff
@@ -93,7 +93,8 @@ func (p *Priority) GetPriority() int {
 
 // BasePostFunction describe a default posthook function
 type BasePostFunction struct {
-	f func(CommonCluster) error
+	f      func(CommonCluster, pkgCluster.PostHookConfig) error
```
sagikazarmark: Instead of this change I would rather add the configuration to the global config (in `internal/global`) and use it in posthooks where necessary.

Posthooks will be decomissioned eventually, so I would keep changes here minimal.

pkg/values/values.go
```diff
@@ -0,0 +1,69 @@
...
+
+func toMap(v string) (map[string]interface{}, error) {
+	var out = make(map[string]interface{})
+	var trimmedStr = strings.TrimSpace(v)
+	err := yaml.Unmarshal([]byte(trimmedStr), &out)
+	if err != nil {
+		return nil, errors.WrapIf(err, "failed to unmarshal values")
+	}
+
+	return out, nil
```
siliconbrain:
```go
var out map[string]interface{}
err := yaml.Unmarshal([]byte(strings.TrimSpace(v)), &out)
return out, errors.WrapIf(err, "failed to unmarshal values")
```

pkg/values/values.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/mitchellh/mapstructure"
+)
+
+type Config map[string]interface{}
+
+func DecodeHook() mapstructure.DecodeHookFunc {
+	return func(a reflect.Type, b reflect.Type, d interface{}) (interface{}, error) {
+
+		// parse duration
+		if a.Kind() == reflect.String && b == reflect.TypeOf(new(time.Duration)).Elem() {
```
sagikazarmark: Why is this necessary?

pkg/values/values.go
```diff
@@ -0,0 +1,59 @@
...
+			return time.ParseDuration(d.(string))
+		}
+
+		// parse values config
+		if a.Kind() == reflect.String && b == reflect.TypeOf(new(Config)).Elem() {
+
+			if data, ok := d.(string); ok {
+				output, err := toMap(data)
```
sagikazarmark: I think this function could be inlined, we won't be reusing it any where and extracting it doesn't make to code easier to understand.

config/config.yaml.dist
```diff
@@ -362,6 +362,15 @@ dex:
...
 #                values: {}
+#
+#    posthook:
+#        ingress:
+#            enabled: true
+#            chart: "banzaicloud-stable/pipeline-cluster-ingress"
+#            version: "0.0.8"
+#
+#            # See https://github.com/helm/charts/tree/master/stable/traefik for details
+#            values: |
```
sagikazarmark: I think we could add the default values here to show the user them.

pkg/values/values.go
colin014: To avoid these type of panics:
```
* cannot parse 'Cluster.DisasterRecovery.Ark.BackupSyncInterval' as int: strconv.ParseInt: parsing "20s": invalid syntax
```

when we want to parse a `string` into a `time.Duration`

PR #2697, opened 2020-02-03
Remove unused endpoints from openapi

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
Removes the following endpoints:
- cluster monitoring update (replaced with integrated service)
- old feature endpoints (replaced with integrated service)

**Please make sure that these endpoints are not used anywhere.**


apis/pipeline/pipeline.yaml
```diff
@@ -3658,112 +3637,6 @@ paths:
                 default:
                     $ref: '#/components/responses/Error'
 
-    /api/v1/orgs/{orgId}/clusters/{id}/features:
```
siliconbrain: I think we still call these from the CLI, I'm not sure about the web UI.

sagikazarmark: :sadface:

PR #2701, opened 2020-02-05
Ingress integrated service MVP

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
The MVP of an integrated service for deploying an ingress controllers to clusters.

### Why?
This integrated service is to supersede the ingress controller post hook for deploying an ingress controller to the cluster.

### Additional info

Service spec format:
```json
{
  "controller": {
    "type": "traefik",
    "config": { // controller specific parameters
      "ssl": {
        "defaultCN": "domain.org",
        "defaultSANList": ["domain.org", "*.domain.org"],
        "defaultIPList": ["1.2.3.4"]
    }
  },
  "ingressClass": "traefik-internal",
  "service": {
    "type": "LoadBalancer", // LoadBalancer | NodePort | ClusterIP
    "annotations": {
      "foo": "bar"
    }
  }
}
```

Relevant capabilities:
* `features.ingress.enabled`: is the service enabled
* `features.ingress.controllers`: list of enabled controller types (e.g. `["traefik", "nginx"]`)

### Checklist


pkg/jsonstructure/copy.go
```diff
@@ -0,0 +1,83 @@
...
+// limitations under the License.
+
+package jsonstructure
+
+import (
+	"fmt"
+
+	"emperror.dev/errors"
+)
+
```
lpuskas: can't the same results be achived by using mapstructure?

siliconbrain: Sadly, no,
```go
src := map[string]interface{}{ ... }
var dst map[string]interface{}
mapstructure.Decode(src, &dst)
```
will only create a shallow copy of `src` into `dst`.

internal/integratedservices/services/ingress/config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type ChartsConfig struct {
+	Traefik TraefikChartConfig
+}
+
+type TraefikChartConfig struct {
+	Chart   string
+	Version string
+	Values  map[string]interface{}
```
sagikazarmark: I guess this should be `values.Config`, or whatever the type is.

internal/integratedservices/services/ingress/ingressadapter/org_domain_service.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+
+	"emperror.dev/errors"
+
+	"github.com/banzaicloud/pipeline/internal/integratedservices/services/ingress"
+	"github.com/banzaicloud/pipeline/src/auth"
+	"github.com/banzaicloud/pipeline/src/dns"
+)
+
+type OrgDomainService struct {
```
sagikazarmark: I would stick to the best effort solution as seen in the posthook. This service will eventually go away/will be rewritten, so I don't think it's worth introducing the dependency. The best-effort solution is about 5 lines. If you want, you can wrap that in a service or something. It should probably be a cluster domain service though.

siliconbrain: How is this different from the implementation in the post-hook other than it's hidden behind a layer of abstraction to enabled easier testing?

sagikazarmark:
```
func (s OrgDomainService) GetOrgDomain(ctx context.Context, orgID uint) (ingress.OrgDomain, error) {
```

It binds the ingress service to a relatively old concept which will eventually be removed.

siliconbrain: The service's current implementation has to get these values somehow. When it won't, this all can be deleted. What's the harm? Would it be better to directly depend on an org repo service to get the organization's values and the base domain config value? Those concepts could become irrelevant just as easily.

PR #2704, opened 2020-02-06
replace decode hook for custom wrapper type

siliconbrain:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?
A replacement for the `mapstructure` decode hook for the custom `type Config map[string]interface{}` wrapper type.

### Why?
The new hook makes it possible to define YAML/JSON object values in the config as strings that will be unmarshalled to `map[string]interface{}` typed fields without changing the code to use a custom type.

### DIFF
--- a/internal/global/config.go
+++ b/internal/global/config.go
@@ -20,7 +20,6 @@ import (
 	"time"
 
 	"github.com/banzaicloud/pipeline/internal/platform/log"
-	"github.com/banzaicloud/pipeline/pkg/values"
 )
 
 // Config is a global config instance.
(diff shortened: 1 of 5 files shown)


sagikazarmark:
👎 

I prefer using a specific type, because this way every string that's supposed to be unmarshaled int o a map will be assumed to be yaml/json, even if that's a mistake.

Explicit type sends a signal to the hook to only parse the config, if it's supposed to be a map according to the type.

siliconbrain:
_I'll be back..._

PR #2751, opened 2020-02-24
fix nodepool API returning 500 errors when PKE clusters are not ready

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
When cluster is not ready, like there's no k8s secret yet, node pool labels are not available as they are stored on cluster. They are only fetched in case the cluster isReady, which is not implemented yet on PKE clusters, resulting in errors codes, instead of 206 StatusPartialContent codes.

### Checklist

### DIFF
--- a/internal/providers/azure/pke/driver/commoncluster/cluster.go
+++ b/internal/providers/azure/pke/driver/commoncluster/cluster.go
@@ -235,6 +235,10 @@ func (a *AzurePkeCluster) GetStatus() (*pkgCluster.GetClusterStatusResponse, err
 }
 
 func (a *AzurePkeCluster) IsReady() (bool, error) {
+	// cluster is not ready in case there's no config secret yet
+	if len(a.GetConfigSecretId()) == 0 {
+		return false, nil
+	}
 	return true, nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/providers/azure/pke/driver/commoncluster/cluster.go
```diff
@@ -235,6 +235,10 @@ func (a *AzurePkeCluster) GetStatus() (*pkgCluster.GetClusterStatusResponse, err
 }
 
 func (a *AzurePkeCluster) IsReady() (bool, error) {
+	// cluster is not ready in case there's no config secret yet
+	if len(a.GetConfigSecretId()) == 0 {
```
siliconbrain:
```suggestion
	if a.GetConfigSecretId() == "" {
```

src/cluster/ec2_pke.go
```diff
@@ -673,7 +673,10 @@ func (c *EC2ClusterPKE) GetStatus() (*pkgCluster.GetClusterStatusResponse, error
 
 // IsReady checks if the cluster is running according to the cloud provider.
 func (c *EC2ClusterPKE) IsReady() (bool, error) {
-	// TODO: is this a correct implementation?
+	// cluster is not ready in case there's no config secret yet
+	if len(c.GetConfigSecretId()) == 0 {
```
siliconbrain:
```suggestion
	if c.GetConfigSecretId() == "" {
```

PR #2785, opened 2020-03-03
helm environment service implementation

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no
| Related tickets | 
| License         | Apache 2.0

### What's in this PR?
Add helm environment related operations to the new helm service

### Why?
The legacy functionality has to be ported (temporarily) to the new helm service implementation 

### Additional context
- the integration needs to be backed off before merge

Issue #2748, opened 2020-02-24
hpa-operator causes operator-sdk based operators to fail


khernyo:
**Describe the bug**
Operators built using operator-framework/operator-sdk from helm charts don't work on pipeline created k8s clusters.

**Steps to reproduce the issue:**

1. create a k8s cluster using pipeline (any of EKS 1.14, PKE 1.15 or PKE 1.17, possibly others)
1. install sample helm based memcached-operator by following the instructions at https://github.com/operator-framework/operator-sdk-samples/tree/master/helm/memcached-operator
1. observe that no memcached pods are created

Note: the go based memcached-operator is not affected.

**Expected behavior**
memcached-operator creates three memcached pods

**Additional context**
pipeline 0.39.0-dev.6 (61a9abf62)

Error log from memcached-operator:
```
{
  "level": "error",
  "ts": 1582549445.626265,
  "logger": "controller-runtime.controller",
  "msg": "Reconciler error",
  "controller": "memcached-controller",
  "request": "helm-memcached/example-memcached",
  "error": "failed to install release: could not get apiVersions from Kubernetes: unable to retrieve the complete list of server APIs: custom.metrics.k8s.io/v1beta1: Got empty response for: custom.metrics.k8s.io/v1beta1",
[...]

PR #2792, opened 2020-03-04
hpa: use latest hpa chart, fix Prometheus url

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2785 2748

### What's in this PR?
- Use latest hpa-operator & kube-metrics-adapter chart, which is basedon latest controller runtime, custom metrics endpoint is no longer used
- Fix Prometheus url setup for kube-metrics-adapter 

### Why?
Some helm based operators may fail, in case custom metrics endopoint exists, but doesn't return any items, as described in: https://github.com/banzaicloud/pipeline/issues/2748

### Checklist

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -654,7 +654,7 @@ ssl:
 	v.SetDefault("cluster::ingress::cert::path", "config/certs")
 
 	v.SetDefault("cluster::autoscale::namespace", "")
-	v.SetDefault("cluster::autoscale::hpa::prometheus::serviceName", "monitor-prometheus-server")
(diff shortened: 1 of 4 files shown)


internal/cmd/config.go
```diff
@@ -694,7 +694,7 @@ ssl:
 	})
 
 	v.SetDefault("cluster::autoscale::charts::hpaOperator::chart", "banzaicloud-stable/hpa-operator")
-	v.SetDefault("cluster::autoscale::charts::hpaOperator::version", "0.1.0")
+	v.SetDefault("cluster::autoscale::charts::hpaOperator::version", "0.2.2")
```
sagikazarmark: Please update `config.yaml.dist`

PR #2820, opened 2020-03-19
separate platform and org bound helm environments

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | yes

### What's in this PR?
- helm env service rewritten to differentiate between org bound platform helm envs
- default (builtin) repositories installed in the platform helm env (user deployments need to add repositories in order to be able to install charts)

### Additional context
 - builtin repos are added to the org env, they're not persisted !!!

### Checklist


internal/helm/envresolver.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return HelmEnv{
+		home:     path.Join(h.helmHomes, orgName, helmPostFix),
+		platform: false,
+	}, nil
+}
+
+func (h helmEnvResolver) ResolvePlatformEnv(ctx context.Context) (HelmEnv, error) {
+	return HelmEnv{
+		home:     path.Join(h.helmHomes, PlatformHelmHome, helmPostFix),
```
siliconbrain: Are we sure, that organizations cannot be called "pipeline"?

internal/helm/service.go
```diff
@@ -91,6 +96,14 @@ type Store interface {
...
 }
 
+type Releaser interface {
+	// TODO list / define all operations related to releases
+}
+
+type Chartmanager interface {
+	// TODO list /define all operations related to charts
+}
+
```
siliconbrain: These two are not used (yet?).

internal/helm/service.go
```diff
@@ -291,3 +348,23 @@ func (s service) repoExists(ctx context.Context, orgID uint, repository Reposito
 
 	return true, nil
 }
+
+// mergeDefaults adds the defaults to the list of repositories if already not added
```
siliconbrain:
```suggestion
// mergeDefaults adds the defaults to the list of repositories if not already added
```

src/helm/install.go
```diff
@@ -51,6 +52,10 @@ func GenerateHelmRepoEnv(orgName string) (env helmEnv.EnvSettings) {
 	return
 }
 
+func GeneratePlatformHelmRepoEnv() (env helmEnv.EnvSettings) {
```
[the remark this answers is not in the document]
lpuskas: thanks, fixed

internal/helm/envresolver.go
lpuskas: i'll rename it to _platform_ instead; eventually this can come from config

internal/helm/service.go
lpuskas: yes, "tyet". -removed them

PR #2829, opened 2020-03-24
Fix potential panic in anchore client

sagikazarmark:
Backport #2828 to 0.40.x


sagikazarmark:
No need to backport, not relevant to this version.

PR #2832, opened 2020-03-26
chart: extend tls for worker intercommunication

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?

- removal of the hardcoded pipeline CN
- pipeline TLS is always enabled from the chart
- pass CA and client certificate to the worker as well
- split CA and the actual certificate in the Secret
- make cert consumption cert-manager friendly

### Additional context

The Helm generated certificate can be used as a client and server certificate as well, so it works fine for default Pipeline and Worker intercommunication. These certificates can be sourced from the outside as well if defined.

### Checklist


.gitignore
```diff
@@ -35,3 +35,7 @@ test.txt
 !/.idea/modules.xml
 !/.idea/runConfigurations/
 !/.idea/scopes/
+
+# Helm
+*.tgz
+*.lock
```
sagikazarmark: Since this is chart specific, could go into a gitignore file in the charts directory.

charts/pipeline/templates/deployment-worker.yaml
```diff
@@ -81,6 +81,15 @@ spec:
             secretKeyRef:
               name: {{ include "pipeline.fullname" . }}
               key: githubToken
+        # Pipeline service and TLS
```
sagikazarmark: I don't think the builtin worker needs these settings.

charts/pipeline/templates/secrets.yaml
```diff
@@ -16,25 +16,36 @@ data:
...
     heritage: "{{ .Release.Service }}"
 data:
-  ca.crt: {{ $ca.Cert | b64enc }}
-  ca.key: {{ $ca.Key | b64enc }}
-  server.crt: {{ $cert.Cert | b64enc }}
-  server.key: {{ $cert.Key | b64enc }}
+  tls.crt: {{ $cert.Cert | b64enc }}
+  tls.key: {{ $cert.Key | b64enc }}
+---
+apiVersion: v1
```
sagikazarmark: Cert manager places CA in the same secret, under `ca.crt` key (if we are talking about cert manager friendly setup). Also, I'm not sure from a consumer point of view we need them in separate secrets.

We only need the CA key to actually sign certs, but that will rarely happen inside the cluster (unless we use the self-signed signer of cert manager, but even then the CA cert is placed in the final secret as well).

cmd/pipeline/config.go
```diff
@@ -175,6 +176,7 @@ func configure(v *viper.Viper, p *pflag.FlagSet) {
 	v.SetDefault("pipeline::caCertFile", "")
 	v.SetDefault("pipeline::certFile", "")
 	v.SetDefault("pipeline::keyFile", "")
+	v.SetDefault("pipeline::commonName", "pipeline")
```
sagikazarmark: What's the benefit of setting the common name? IMHO we probably don't even have to validate the common name. Matching the CA is more than enough, isn't it?

.gitignore
bonifaido: Fixed.

charts/pipeline/templates/secrets.yaml
bonifaido: Fixed.

charts/pipeline/templates/deployment-worker.yaml
bonifaido: Currently not, but it works as a reference for the future worker. We can remove this for now. (The tls secret was mounted here also before this PR).

cmd/pipeline/config.go
bonifaido: Fixed.

charts/pipeline/templates/deployment-worker.yaml
bonifaido: Removed.

PR #2834, opened 2020-03-31
[Vsphere] Add force delete check to error handling

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add force delete check to every error case in case of Vsphere cluster delete

### Why?
To delete Vsphere clusters in case of errors too

### Checklist


internal/providers/vsphere/pke/workflow/delete_cluster.go
```diff
@@ -96,9 +96,13 @@ func DeleteClusterWorkflow(ctx workflow.Context, input DeleteClusterWorkflowInpu
 
 			err := workflow.ExecuteActivity(ctx, DeleteNodeActivityName, activityInput).Get(ctx, nil)
 			if err != nil {
-				e := errors.WrapIff(err, "deleting node %q", node.Name)
```
orymate: i'd log this wrapped error (or add the node name to the log)

PR #2840, opened 2020-04-06
AKS: remove Storage Contributor Role assignment to VM's

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
As we no longer install PVC operator there's no need to assign Storage Contributor Role to VM's.

### Why?
Some times users creating AKS cluster run into role assigning permission problems.

### Checklist


src/cluster/aks.go
```diff
@@ -290,132 +282,6 @@ func makeInfrastructureResourceGroupName(resourceGroupName, clusterName, locatio
...
-	if err := assignRoleToVMs(c.log, cc.GetRoleAssignmentsClient(), virtualMachines, role, scope); err != nil {
-		return errors.WrapIf(err, fmt.Sprintf("failed to assign %s role to VMs", StorageAccountContributor))
-	}
-
-	c.log.Info("Role assign succeeded")
-
-	return nil
-}
-
-func enableVMManagedServiceIdentity(log logrus.FieldLogger, client *pkgAzure.VirtualMachinesClient, resourceGroupName string, vm *compute.VirtualMachine) (*compute.VirtualMachine, error) {
```
stoader, line 349: We may want to keep the assignment of identity to VMs as that allows us granting access to cloud resources by adding roles to the identity. An improvement could be done by assigning the identity to the VM at creation time (the same way as we do for VMSSes in for PKE)

PR #2848, opened 2020-04-15
add go templates to cloudformation templates

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Enable Go templating of CF templates.

### Why?

Some parts can't be configured otherwise.

### Checklist

### DIFF
--- a/config/config.yaml.dist
+++ b/config/config.yaml.dist
@@ -40,6 +40,8 @@ pipeline:
         url: ""
 #        insecure: false
 
+#   enterprise: false
+
 database:
     # dialect: "postgres"
     dialect: "mysql"
--- a/internal/cluster/distribution/eks/eksprovider/workflow/templates.go
+++ b/internal/cluster/distribution/eks/eksprovider/workflow/templates.go
@@ -15,10 +15,12 @@
 package workflow
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"text/template"
 
 	"emperror.dev/errors"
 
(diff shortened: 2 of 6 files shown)


config/config.yaml.dist
```diff
@@ -479,3 +479,6 @@ cicd:
 #secret:
 #    tls:
 #        defaultValidity: 8760h # 1 year
+
+enterprise:
+    enabled: false
```
sagikazarmark, line 484: Default config, so it should be commented. Also, default value has to be added to Viper.

```suggestion
#enterprise:
#    enabled: false
```

Furthermore, I think something like this would be better:

```yaml
pipeline:
    enterprise: false
````

File: internal/cluster/distribution/eks/eksprovider/workflow/templates.go [hunk 2]
```diff
@@ -60,7 +62,18 @@ func getEksCloudFormationTemplate(name string) (string, error) {
...
 	}
 
-	return string(content), nil
+	t, err := template.New("cf").Parse(string(content))
+	if err != nil {
+		return "", errors.WrapIf(err, fmt.Sprintf("failed to parse CloudFormation template content from %s", templatePath))
+	}
+
+	buffer := bytes.NewBuffer(nil)
+	err = t.Execute(buffer, global.Config)
```
sagikazarmark, line 71: Ideally the enterprise flag should not appear directly in templates and code. The template data could be something like:

```
data := map[string]interface{}{
    "UpdatePolicyEnabled": !config.Pipeline.Enterprise,
}
```

internal/cmd/config.go
```diff
@@ -86,6 +86,10 @@ type Config struct {
 		}
 	}
 
+	Enterprise struct {
```
sagikazarmark, line 89: I feel like this would fit the Pipeline config better.

bonifaido:
Fixed accordingly.

cmd/worker/eks.go
```diff
@@ -46,6 +46,8 @@ func registerEKSWorkflows(secretStore eksworkflow.SecretStore, clusterManager *a
 		return errors.WrapIf(err, "failed to get CloudFormation template for node pools")
 	}
 
+	println(nodePoolTemplate)
+
```
sagikazarmark, line 50:
```suggestion
```

config/config.dev.yaml
```diff
@@ -14,7 +14,7 @@ pipeline:
     external:
         # Base URL where the end users can reach this pipeline instance
         url: "https://example.com/pipeline"
-
+
```
sagikazarmark, line 17:
```suggestion

```

bonifaido:
Fixed, thanks!

PR #2857, opened 2020-04-17
Remove CICD

sagikazarmark:
Remove CICD integration components


cmd/pipeline/main.go
```diff
@@ -629,7 +623,7 @@ func main() {
 		dcGroup.GET("", dashboardAPI.GetClusterDashboard)
 	}
 
-	scmTokenStore := auth.NewSCMTokenStore(tokenStore, config.CICD.Enabled)
+	scmTokenStore := auth.NewSCMTokenStore(tokenStore)
```
bonifaido, line 626: I think scmTokenStore is not needed anymore.

sagikazarmark: It's required by the user update API which cannot be disabled on the UI at the moment. I will submit the removal of SCM tokens in a separate PR.

bonifaido: cool!

PR #2883, opened 2020-04-24
Expose scheme field of proxy settings in case of PKE

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Expose scheme field of proxy settings in case of PKE

### Why?
HTTP/S proxy requested in the cluster creation requests (all PKE providers) assume that http proxy is served over http, while https is served over https. This case is rare, so scheme should be requested as well.

### Checklist


apis/pipeline/pipeline.yaml
```diff
@@ -4200,6 +4200,9 @@ components:
                 secretId:
                     description: "ID of the secret containing the username and password for the proxy"
                     type: string
+                scheme:
+                    description: "Scheme of the proxy"
+                    type: string
```
siliconbrain, line 4205: Shouldn't this be an enum? https://swagger.io/docs/specification/data-models/enums/

PR #2889, opened 2020-04-27
Helm3 support for the Backyards integrated service

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | yes

### What's in this PR?
Helm3 support for the Backyards integration by replacing existing helm2 calls with the generic implementation.
Fixes for the helm3 implementation:
- add wait if configured
- explicit timeouts for release operations
- create namespace if it does not exist

### Why?
To be able to switch to helm3,  and still support helm2 in the meantime.

### DIFF
--- a/internal/helm/helmadapter/releaser_service.go
+++ b/internal/helm/helmadapter/releaser_service.go
@@ -16,8 +16,10 @@ package helmadapter
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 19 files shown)


internal/helm/helmadapter/releaser_service.go
```diff
@@ -124,6 +129,33 @@ func (r releaser) Install(_ context.Context, helmEnv helm.HelmEnv, kubeConfig he
 		}
 	}
 
+	clientSet, err := k8sclient.NewClientFromKubeConfigWithTimeout(kubeConfig, time.Second*10)
```
lpuskas, line 132: I'd add a new collaborator for performing namespace related part (ensure namespace)

pepov: I can do it, but I don't see much added value in it right now. This is already the lowest level layer in the stack and the helm operations are also working with pure kubeconfig here. Unit tests would benefit from a dependency that can be mocked out, but I think we are better off with integration tests and from that point of view it doesn't matter whether it's a dependency or inline.
I can however move this to a separate function call for more clarity if that is a good compromise.

PR #2896, opened 2020-04-29
Process log interfaces V2

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets |

### What's in this PR?
Process log interfaces used by third party components


pkg/sdk/cadence/lib/pipeline/process_log.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// StartActivity records a new activity of a process.
+	StartActivity(ctx workflow.Context, typ string) Activity
+
+	// RecordEvent records a new event that happened at a single point in time during the execution of a process.
+	RecordEvent(ctx workflow.Context, typ string, err error)
+}
+
+// Activity is a short lived part of a Process.
+type Activity interface {
```
bonifaido, line 40: It might be worth adding a "resource/object" setter method to this interface as well.

sagikazarmark: What would it set that's not already set by in the process by the resource ID (that would be a BRN in this case)

bonifaido: I was thinking about node name.

sagikazarmark: That's more like payload to me which I would like to handle in a separate PR.

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
```diff
@@ -42,20 +54,22 @@ type UpdateNodePoolWorkflowInput struct {
...
 	}
 
 	ctx = workflow.WithActivityOptions(ctx, activityOptions)
 
-	proc := process.Start(
-		workflow.WithStartToCloseTimeout(ctx, 10*time.Minute),
-		input.OrganizationID,
-		fmt.Sprint(input.ClusterID),
-	)
+	clusterID := brn.New(input.OrganizationID, brn.ClusterResourceType, fmt.Sprint(input.ClusterID))
```
sagikazarmark, line 68: Two ideas here:

- provide a BRN to the workflow in the first place?
- use UID instead of the numeric ID?

bonifaido: it is not strictly numeric (that is why I choose string for the resource id)

this should be something that is user friendly on the other end

PR #2897, opened 2020-04-29
Add cadence AWS lib

sagikazarmark:
.


sagikazarmark:
We are going to do this a bit differently.

PR #2903, opened 2020-05-04
ClusterSetup workflow enhanced not to install Tiller in case helm v3 is switched on 

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
* Remove tiller related activities from the cluster setup workflow
* Amendments to the Releaser implementation to handle all namespaces
* Decoupled delete all releases call from the helm version (affects cluster deletion)

### Why?
No tiller is needed for helm 3

### DIFF
--- a/internal/cmd/helm.go
+++ b/internal/cmd/helm.go
@@ -90,14 +90,21 @@ func CreateUnifiedHelmReleaser(
 }
 
 // CreateReleaseDeleter creates a new (helm2 or helm3 specific) deleter instance based on the provided argunments
-func CreateReleaseDeleter(helmConfig helm.Config, logger helm.Logger) helm.ReleaseDeleter {
-	orgService := helmadapter.NewOrgService(logger)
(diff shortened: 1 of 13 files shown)


internal/cmd/helm.go
```diff
@@ -88,3 +88,16 @@ func CreateUnifiedHelmReleaser(
 
 	return helmadapter.NewUnifiedHelm3Releaser(service, logger), service
 }
+
+// CreateReleaseDeleter creates a new (helm2 or helm3 specific) deleter instance based on the provided argunments
+func CreateReleaseDeleter(helmConfig helm.Config, logger helm.Logger) helm.ReleaseDeleter {
+	orgService := helmadapter.NewOrgService(logger)
+	if helmConfig.V3 {
+		releaser := helmadapter.NewReleaser(logger)
+		// TODO ensuring envresolver
```
pepov, line 97: why is this a todo?

PR #2911, opened 2020-05-06
Add helm version to capabilities API

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Add helm version to capabilities API

### Why?
CLI should be able to query the currently used Helm version.


pepov:
Would it make sense to add a comment in the go.mod so that we don't forget to update in the caps config?

PR #2916, opened 2020-05-12
List helm charts operation made backwards compatible

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The helm chartlist response is made similar to the legacy (helm2)

### Why?
Each helm chart version was displayed on the UI (broken backwards compatibility)

### DIFF
--- a/internal/helm/helmadapter/helm3_env_service.go
+++ b/internal/helm/helmadapter/helm3_env_service.go
@@ -224,10 +224,12 @@ func (h helm3EnvService) ListCharts(ctx context.Context, helmEnv helm.HelmEnv, f
 			Name:   filter.RepoFilter(),
 			Charts: make([]interface{}, 0, 0),
 		}
+
+		versions := make([]interface{}, 0, len(chartVersions))
 		for _, chartVersion := range chartVersions {
-			// omg!
-			chartEntry.Charts = append(chartEntry.Charts, []interface{}{chartVersion})
+			chartEntry.Charts = append(versions, chartVersion)
 		}
+		chartEntry.Charts = append(chartEntry.Charts, versions)
 	}
 
 	if len(chartEntry.Charts) > 0 {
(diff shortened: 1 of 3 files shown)


internal/helm/helmadapter/helm3_env_service.go
```diff
@@ -224,10 +224,12 @@ func (h helm3EnvService) ListCharts(ctx context.Context, helmEnv helm.HelmEnv, f
 			Name:   filter.RepoFilter(),
 			Charts: make([]interface{}, 0, 0),
 		}
+
+		versions := make([]interface{}, 0, len(chartVersions))
 		for _, chartVersion := range chartVersions {
-			// omg!
-			chartEntry.Charts = append(chartEntry.Charts, []interface{}{chartVersion})
+			chartEntry.Charts = append(versions, chartVersion)
```
sagikazarmark, line 230: This doesn't seem right. Shouldn't it be `versions = append(versions, chartVersion)` ?

internal/helm/helmadapter/helm3_env_service.go
```diff
@@ -224,10 +224,12 @@ func (h helm3EnvService) ListCharts(ctx context.Context, helmEnv helm.HelmEnv, f
...
 		}
+		chartEntry.Charts = append(chartEntry.Charts, versions)
```
sagikazarmark, line 232: Not sure about this either, but this could be correct. Just double checking that it's not intended to be

```
chartEntry.Charts = append(chartEntry.Charts, versions...)
```

internal/helm/helmadapter/helm3_env_service.go
lpuskas: thanks!

internal/helm/helmadapter/helm3_env_service.go
lpuskas: this is correct in this form here ... thx

PR #2944, opened 2020-05-21
Add security scan information to the release list response

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
This PR implements decoration of release list with security scan information

### Why?
This information was provided in the earlier versions as well


internal/helm/service.go
```diff
@@ -537,7 +540,7 @@ func (s service) UpgradeRelease(ctx context.Context, organizationID uint, cluste
 	return nil
 }
 
-func (s service) ListCharts(ctx context.Context, organizationID uint, filter ChartFilter, options Options) (charts ChartList, err error) {
+func (s service) ListCharts(ctx context.Context, organizationID uint, filter ChartFilter, _ Options) (charts ChartList, err error) {
```
pepov, line 540: Not sure where these changes come from but doesn't seem like it's necessary

internal/helm/service.go
```diff
@@ -551,7 +554,7 @@ func (s service) ListCharts(ctx context.Context, organizationID uint, filter Cha
 	return chartList, nil
 }
 
-func (s service) GetChart(ctx context.Context, organizationID uint, chartFilter ChartFilter, options Options) (chartDetails ChartDetails, err error) {
+func (s service) GetChart(ctx context.Context, organizationID uint, chartFilter ChartFilter, _options Options) (chartDetails ChartDetails, err error) {
```
pepov, line 554: Not sure where these changes come from but doesn't seem like it's necessary

internal/helm/service.go
lpuskas: this is linter a linter fix

internal/helm/service.go
lpuskas: linter fix

PR #2953, opened 2020-05-25
Use preinstalled PKE binary

### LINKED ISSUE #2938: Do not install PKE if it's already present in the image
PKE is always downloaded in the user data script, regardless if it's already installed.

Check if PKE is already installed and only download it if it's not.

**Additional context**
When using a custom image, I want to use the PKE version in the image, not the one provided by the user.

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2938

### What's in this PR?
Use preinstalled PKE binary in the user data script.

### Additional context
It might make sense to also validate if the binary is executable. In this case the condition would look like:

```bash
if ! [ -x "$(command -v pke)" ]; then
    # ...
fi
```

### DIFF
--- a/templates/pke/master.cf.yaml
+++ b/templates/pke/master.cf.yaml
@@ -64,9 +64,11 @@ Resources:
 
             hostnamectl set-hostname $(curl http://169.254.169.254/latest/meta-data/hostname)
 
-            curl -v https://banzaicloud.com/downloads/pke/pke-${PkeVersion} -o /usr/local/bin/pke
-            chmod +x /usr/local/bin/pke
-            export PATH=$PATH:/usr/local/bin/
+            if ! command -v pke > /dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


templates/pke/master.cf.yaml
```diff
@@ -66,7 +66,7 @@ Resources:
 
             export PATH=$PATH:/usr/local/bin/
             if ! command -v pke > /dev/null 2>&1; then
-                curl -v https://banzaicloud.com/downloads/pke/pke-${PkeVersion} -o /usr/local/bin/pke
+                until curl -v https://banzaicloud.com/downloads/pke/pke-{{ .PKEVersion }} -o /usr/local/bin/pke; do sleep 10; done
```
orymate, line 69:
```suggestion
                until curl -v https://banzaicloud.com/downloads/pke/pke-${PkeVersion} -o /usr/local/bin/pke; do sleep 10; done
```

PR #2955, opened 2020-05-26
EKS: update node pool CF template to latest EKS node group template

### LINKED ISSUE #2948: Upgrade EKS node pool CF template
Download the latest node pool CF template:

```
wget https://amazon-eks.s3.us-west-2.amazonaws.com/cloudformation/2020-05-08/amazon-eks-nodegroup.yaml
```

Apply the necessary changes from our templates.

(We should stay close to the official templates for easier upgrades in the future)

This is change is required by the latest EKS AMIs according to the documentation:
https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2948 

### What's in this PR?
Update EKS CloudFormation templates to align with latest EKS node group template: https://amazon-eks.s3.us-west-2.amazonaws.com/cloudformation/2020-05-08/amazon-eks-nodegroup.yaml

### Checklist

### DIFF
--- a/templates/eks/amazon-eks-nodepool-cf.yaml
+++ b/templates/eks/amazon-eks-nodepool-cf.yaml
@@ -2,173 +2,6 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: 'Amazon EKS - Node Group'
 
-Parameters:
-
-  KeyName:
-    Description: The EC2 Key Pair to allow SSH access to the instances
-    Type: String
-
-  NodeImageId:
-    Type: AWS::EC2::Image::Id
-    Description: AMI id for the node instances.
-
-  NodeInstanceType:
-    Description: EC2 instance type for the node instances
-    Type: String
-    Default: t2.medium
-    AllowedValues:
-    - t2.small
-    - t2.medium
-    - t2.large
-    - t2.xlarge
-    - t2.2xlarge
-    - t3.nano
-    - t3.micro
-    - t3.small
-    - t3.medium
-    - t3.large
-    - t3.xlarge
-    - t3.2xlarge
-    - m3.medium
-    - m3.large
-    - m3.xlarge
-    - m3.2xlarge
-    - m4.large
-    - m4.xlarge
-    - m4.2xlarge
-    - m4.4xlarge
-    - m4.10xlarge
-    - m5.large
-    - m5.xlarge
-    - m5.2xlarge
-    - m5.4xlarge
-    - m5.12xlarge
-    - m5.24xlarge
-    - c4.large
-    - c4.xlarge
-    - c4.2xlarge
-    - c4.4xlarge
-    - c4.8xlarge
-    - c5.large
-    - c5.xlarge
-    - c5.2xlarge
-    - c5.4xlarge
-    - c5.9xlarge
-    - c5.18xlarge
-    - i3.large
-    - i3.xlarge
-    - i3.2xlarge
-    - i3.4xlarge
-    - i3.8xlarge
-    - i3.16xlarge
-    - r3.xlarge
-    - r3.2xlarge
-    - r3.4xlarge
-    - r3.8xlarge
-    - r4.large
-    - r4.xlarge
-    - r4.2xlarge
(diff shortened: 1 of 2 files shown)


templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -2,16 +2,92 @@
...
+
+  NodeAutoScalingGroupMinSize:
+      Type: Number
+      Description: Minimum size of Node Group ASG.
+      Default: 1
+
+  NodeAutoScalingGroupMaxBatchSize:
+      Type: Number
+      Description: Minimum size of Node Group ASG.
+      Default: 2
```
sagikazarmark, line 72: This is a custom parameter related to the upgrade process, so it does not have to be in the middle of the autoscaling settings. (Custom params could be at the end of the list?)

internal/cluster/distribution/eks/defaults.go
```diff
@@ -113,6 +113,15 @@ var defaultImageMap = []struct {
 			"us-west-2":      "ami-000a48e69e7695a4a", // US West (Oregon).
 		},
 	},
+	{
+		constraintForVersion("1.16"),
+		map[string]string{
+			// Kubernetes Version 1.16
+			"eu-north-1": "ami-030f9071e0d32b989", // EU (Stockholm).
+			"eu-west-3":  "ami-0170bd57fa0c260da", // EU (Paris).
+			"us-east-2":  "ami-0edc51bc2f03c9dc2", // US East (Ohio).
```
sagikazarmark, line 122: @pbalogh-sa can help with the full list, but ideally in a separate PR.

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -288,7 +288,7 @@ func (eks *UpdateClusterAmazonEKS) Validate() error {
 
 // isValidVersion validates the given K8S version
 func isValidVersion(version string) (bool, error) {
-	constraint, err := semver.NewConstraint(">= 1.13, < 1.16")
+	constraint, err := semver.NewConstraint("> 1.13, <= 1.16")
```
sagikazarmark, line 291: Removing 1.13 should be in a separate PR IMHO.

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -274,7 +274,7 @@ Resources:
         !Ref Subnets
       Tags:
       - Key: Name
-        Value: !Sub "${ClusterName}-${NodeGroupName}"
+        Value: !Sub "${ClusterName}-${NodeGroupName}-Node"
```
sagikazarmark, line 277: Is this going to be backward compatible?

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: In latest CF template parameter are alphabetically ordered, that's why I ordered the same way to be easily comparable. We may put custom params at the end in same order.

internal/cluster/distribution/eks/defaults.go
sancyx: The plan was to have single pr

pbalogh-sa: The list is full now.

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: reordered params to have additional params at the end

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: It's not in terms of, that asg's made with previous CF template will have different tags, however it doesn't seem to effect anything, at least nodes are joining with this new tag as well to EKS master, regardless of k8s version.

PR #2958, opened 2020-05-26
made the helm release api (install and upgrade) backwards compatible

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no|
| Deprecations?   | no|

### What's in this PR?
Release operations in the helm service modified to return the same json structures as the pervious implementation

### Why?
Not to break the API so that other components (eg.: UI, CLI) work properly

### Checklist


internal/helm/helmdriver/transport_http.go
```diff
@@ -212,6 +236,30 @@ func decodeUpgradeReleaseHTTPRequest(_ context.Context, r *http.Request) (interf
...
+
+	// backwards compatibility!
+	oldResponse := struct {
+		ReleaseName string                 `json:"releaseName"`
+		Notes       string                 `json:"notes"`
+		Resources   []helm.ReleaseResource `json:"resources"`
+	}{
+		ReleaseName: resp.Release.ReleaseName,
+		Notes:       resp.Release.ReleaseInfo.Notes,
+		//Resources:   resp.Release.ReleaseResources,
```
pepov, line 257: is this intentional? I mean we return this in the install response, but not here

lpuskas: yes, to be similar to the. old implementation

PR #2959, opened 2020-05-27
Fix the monitoring integrated service broken activation path

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Fixes the activation path of the monitoring integrated service, as the migration path broke it, since if there is no existing release, then it fails on being unable to get the release.

### DIFF
--- a/internal/integratedservices/services/monitoring/operator.go
+++ b/internal/integratedservices/services/monitoring/operator.go
@@ -25,15 +25,15 @@ import (
 	"k8s.io/api/storage/v1beta1"
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/banzaicloud/pipeline/pkg/k8sclient"
-
 	"github.com/banzaicloud/pipeline/internal/common"
+	"github.com/banzaicloud/pipeline/internal/helm"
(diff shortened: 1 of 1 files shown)


internal/integratedservices/services/monitoring/operator.go
```diff
@@ -335,24 +336,26 @@ func (op IntegratedServiceOperator) installPrometheusOperator(
...
 		release, err := op.helmService.GetDeployment(ctx, cluster.GetID(), prometheusOperatorReleaseName, op.config.Namespace)
 		if err != nil {
-			return err
-		}
-
-		k8sClientFactory := func() (kubernetes.Interface, error) {
-			kubeConfig, err := cluster.GetK8sConfig()
-			if err != nil {
-				return nil, err
+			if !helm.ErrReleaseNotFound(err) {
```
sagikazarmark, line 339: When there is only one additional condition, it's nicer to avoid the arrow pattern and merge the condition with the original branch like this:

```
if err != nil && condition(err) {

} else if err != nil { // if necessary

}
```

Issue #2843, opened 2020-04-08
Add NodePool list API


bonifaido:
**Is your feature request related to a problem? Please describe.**
The new node pool API does not support listing which is annoying when you try to update node pools.

**Describe the solution you'd like to see**
Add a node pool list operation to the node pool API. The list should contain regular node pool parameters (see node pool create/update and cluster details APIs). 

**Additional context**
In case of EKS, get details from CloudFormation and Kubernetes.

For optimal performance, use list APIs and aggregate node pool information later.

PR #2965, opened 2020-06-02
feat(API,EKS,nodepool): added GET list endpoint to node pool API

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/pipeline/issues/2843

### What's in this PR?

`GET api/v1/orgs/:orgId/clusters/:clusterId/nodepools` node pool listing endpoint implementation for EKS clusters.

Example:
```golang
[
    {
        "name": "pool1",
        "size": 2,
        "labels": {
            "node.banzaicloud.io/cpu": "1",
            "node.banzaicloud.io/instanceTypeCategory": "General_purpose",
            "node.banzaicloud.io/memory": "2",
            "node.banzaicloud.io/networkPerfCategory": "low",
            "node.banzaicloud.io/ondemand": "true",
            "nodepool.banzaicloud.io/name": "pool1"
        },
        "autoscaling": {
            "enabled": true,
            "minSize": 1,
            "maxSize": 2
        },
        "instanceType": "t2.small",
        "image": "ami-03d9393d97f5959fe"
    },
    {
        "name": "pool2",
        "size": 1,
        "labels": {
            "node.banzaicloud.io/cpu": "1",
            "node.banzaicloud.io/instanceTypeCategory": "General_purpose",
            "node.banzaicloud.io/memory": "2",
            "node.banzaicloud.io/networkPerfCategory": "low",
            "node.banzaicloud.io/ondemand": "false",
            "nodepool.banzaicloud.io/name": "pool2"
        },
        "autoscaling": {
            "enabled": true,
            "minSize": 1,
            "maxSize": 2
        },
        "instanceType": "t2.small",
        "image": "ami-0644e90665b26316b",
        "spotPrice": "0.03"
    }
]
```

### Why?

From the linked issue:  `The new node pool API does not support listing which is annoying when you try to update node pools.`
Resource management is much more convenient when resources can be read/listed at the same management API where they are mutated.

### Checklist


internal/cluster/clusterdriver/transport_http.go
```diff
@@ -42,6 +42,13 @@ func RegisterHTTPHandlers(endpoints Endpoints, router *mux.Router, options ...ki
 		options...,
 	))
 
+	router.Methods(http.MethodGet).Path("/nodepools").Handler(kithttp.NewServer(
+		endpoints.ListNodePools,
+		decodeListNodePoolsHTTPRequest,
+		kitxhttp.ErrorResponseEncoder(kitxhttp.StatusCodeResponseEncoder(http.StatusAccepted), errorEncoder),
```
sagikazarmark, line 48: Shouldn't there be a response encoder as well? This will encode the endpoint response which is probably not the best option. Also, it should return OK, not Accepted.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
 func generateNodePoolStackName(clusterName string, poolName string) string {
 	return "pipeline-eks-nodepool-" + clusterName + "-" + poolName
 }
+
+// ListNodePools is for listing node pools from CloudsetFormation and NodePoolLabelSets
+func (n nodePoolManager) ListNodePools(
+	ctx context.Context,
+	c cluster.Cluster,
+	st eks.SecretStore,
+	dcf kubernetes.DynamicClientFactory,
```
sagikazarmark, line 111: These should be dependencies of the manager struct, not parameters of it.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+) ([]eks.NodePool, error) {
+	// CloudsetFormation
```
sagikazarmark, line 113: Cloudformation? But I don't think this comment is necessary 🙂

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+		stackMap[*stack.StackName] = stack
+	}
+
+	// NodePoolLabelSets
+	clusterClient, err := dcf.FromSecret(ctx, c.SecretID.String())
+	if err != nil {
+		return nil, err
+	}
+
+	manager := npls.NewManager(clusterClient, global.Config.Cluster.Namespace)
```
sagikazarmark, line 146: Do not use global config unless absolutely necessary. In this case, you can either create a factory for the npls manager, or inject the namespace as a dependency in the node pool manager, but at the very least, leave a TODO to not use global config (it's there for legacy reasons)

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+	labelSets, err := manager.GetAll()
+	if err != nil {
+		return nil, err
+	}
+
+	// Aggregate NodePoolLabelSets and CloudsetFormation Stacks
+	var nodePools []eks.NodePool
+	for labelSetName, labelSet := range labelSets {
+		stack := stackMap[labelSetName]
```
sagikazarmark, line 155: This could be nil if the label name cannot be found in the map which would result in a panic on the next two lines. Maybe check if the value exists in the map: `stack := stackMap[labelSetName]`

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+	describeStacksInput := cloudformation.DescribeStacksInput{}
+	stacksOutput, err := cfClient.DescribeStacks(&describeStacksInput)
+	if err != nil {
+		return nil, err
+	}
+
+	var relevantStacks []*cloudformation.Stack
+	for _, stack := range stacksOutput.Stacks {
+		// TODO: Later a better filtering method will be needed
+		if strings.HasPrefix(*stack.StackName, "pipeline-eks-nodepool-"+c.Name) {
```
sagikazarmark, line 130: You could use `aws.StringValue`instead of dereferencing, it's safer.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+			relevantStacks = append(relevantStacks, stack)
+		}
+	}
+
+	stackMap := map[string]*cloudformation.Stack{}
+	for _, stack := range relevantStacks {
+		stackMap[*stack.StackName] = stack
```
sagikazarmark, line 137: See my comment above. Also, stack names do not match node pool names.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+	}
+
+	// Aggregate NodePoolLabelSets and CloudsetFormation Stacks
+	var nodePools []eks.NodePool
+	for labelSetName, labelSet := range labelSets {
+		stack := stackMap[labelSetName]
+
+		parameterMap := map[string]string{}
+		for _, parameter := range stack.Parameters {
+			parameterMap[*parameter.ParameterKey] = *parameter.ParameterValue
```
sagikazarmark, line 159: same thing about dereferencing

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+		}
+
+		AutoscalingEnabled, err := strconv.ParseBool(parameterMap["ClusterAutoscalerEnabled"])
```
sagikazarmark, line 162: In the common go coding style, local variables are camel cased

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+		NodeAutoScalingInitSize, err := strconv.Atoi(parameterMap["NodeAutoScalingInitSize"])
+		if err != nil {
+			return nil, err
+		}
+
+		nodePool := eks.NodePool{
+			Name:   labelSetName,
+			Labels: labelSet,
+			Size:   NodeAutoScalingInitSize,
+			Autoscaling: struct {
```
sagikazarmark, line 186: maybe extract this into a struct.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+	}
+
+	manager := npls.NewManager(clusterClient, global.Config.Cluster.Namespace)
+	labelSets, err := manager.GetAll()
+	if err != nil {
+		return nil, err
+	}
+
+	// Aggregate NodePoolLabelSets and CloudsetFormation Stacks
+	var nodePools []eks.NodePool
```
sagikazarmark, line 153: You can preallocate this slice since you know its size in advance.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +102,103 @@ func (n nodePoolManager) UpdateNodePool(
...
+
+// ListNodePools is for listing node pools from CloudsetFormation and NodePoolLabelSets
+func (n nodePoolManager) ListNodePools(
+	ctx context.Context,
+	c cluster.Cluster,
+	st eks.SecretStore,
+	dcf kubernetes.DynamicClientFactory,
+) ([]eks.NodePool, error) {
+	// CloudsetFormation
+	sessionFactory := workflow.NewAWSSessionFactory(st)
```
sagikazarmark, line 114: The session factory should be a dependency of this struct.

internal/cluster/distribution/eks/service.go
```diff
@@ -61,29 +66,48 @@ type NodePoolUpdateDrainOptions struct {
 	PodSelector string `mapstructure:"podSelector"`
 }
 
+type SecretStore interface {
+	Get(orgnaizationID uint, secretID string) (*secret.SecretItemResponse, error)
+	GetByName(orgnaizationID uint, secretID string) (*secret.SecretItemResponse, error)
+	Store(organizationID uint, request *secret.CreateSecretRequest) (string, error)
+	Delete(organizationID uint, secretID string) error
+	Update(organizationID uint, secretID string, request *secret.CreateSecretRequest) error
+}
```
sagikazarmark, line 75: Do you need all of these methods here? In Go, interfaces are owned by consumers so that they can define what they need.

Furthermore, if you inject the aws session factory into the manager, you don't need the secret store at all.

internal/cluster/distribution/eks/service.go
```diff
@@ -61,29 +66,48 @@ type NodePoolUpdateDrainOptions struct {
...
+
 // NewService returns a new Service instance.
 func NewService(
 	genericClusters cluster.Store,
 	nodePools NodePoolStore,
 	nodePoolManager NodePoolManager,
+	secretStore SecretStore,
+	dynamicClientFactory kubernetes.DynamicClientFactory,
```
sagikazarmark, line 83: These could be injected into the manager.

internal/cluster/distribution/eks/service.go
```diff
@@ -61,29 +66,48 @@ type NodePoolUpdateDrainOptions struct {
...
 
+// +testify:mock:testOnly=true
+
 // NodePoolManager is responsible for managing node pools.
 type NodePoolManager interface {
 	// UpdateNodePool updates an existing node pool in a cluster.
 	UpdateNodePool(ctx context.Context, c cluster.Cluster, nodePoolName string, nodePoolUpdate NodePoolUpdate) (string, error)
+
+	// ListNodePools lists all node pools in a cluster
+	ListNodePools(ctx context.Context, c cluster.Cluster, st SecretStore, dcf kubernetes.DynamicClientFactory) ([]NodePool, error)
```
sagikazarmark, line 110: As mentioned above, these parameters should be dependencies.

internal/cluster/service.go
```diff
@@ -201,6 +201,9 @@ type Service interface {
 
 	// DeleteNodePool deletes a node pool from a cluster.
 	DeleteNodePool(ctx context.Context, clusterID uint, name string) (deleted bool, err error)
+
+	// List NodePools
+	ListNodePools(ctx context.Context, clusterID uint) ([]NodePool, error)
```
sagikazarmark, line 206: Not sure if you can return this NodePool type here, I'm afraid we will have to use `map[string]interface{}`

internal/cluster/clusterdriver/transport_http.go
```diff
@@ -273,3 +296,20 @@ func encodeDeleteNodePoolHTTPResponse(_ context.Context, w http.ResponseWriter,
 
 	return nil
 }
+
+func encodeListNodePoolsHTTPResponse(ctx context.Context, w http.ResponseWriter, response interface{}) error {
+	resp := response.(ListNodePoolsResponse)
+
+	apiResp := pipeline.ListNodePoolsResponse{}
```
sagikazarmark, line 303:
```suggestion
	var apiResp pipeline.ListNodePoolsResponse
```

internal/cluster/clusterdriver/transport_http.go
```diff
@@ -273,3 +296,20 @@ func encodeDeleteNodePoolHTTPResponse(_ context.Context, w http.ResponseWriter,
...
+	for _, nodePool := range resp.NodePoolList {
+		result := pipeline.NodePoolListItem{}
```
sagikazarmark, line 305:
```suggestion
		var result pipeline.NodePoolListItem
```

apis/pipeline/pipeline.yaml
```diff
@@ -4515,6 +4532,46 @@ components:
...
                     type: string
+        
+        ListNodePoolsResponse:
+            type: object
+            properties:
+                nodepoolList:
+                    type: array
+                    description: Nodepool list
+                    items:
+                        $ref: '#/components/schemas/NodePoolListItem'
```
sagikazarmark, line 4543: I believe we currently return a plain list of items in list-like requests.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -16,27 +16,48 @@ package eksadapter
...
 import (
 	"context"
+	"strconv"
+	"strings"
 	"time"
 
 	"emperror.dev/errors"
 	"go.uber.org/cadence/client"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
```
sagikazarmark, line 26: Please use the v1 SDK.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -16,27 +16,48 @@ package eksadapter
...
 	"github.com/banzaicloud/pipeline/internal/cluster/distribution/eks"
+	"github.com/banzaicloud/pipeline/internal/cluster/distribution/eks/eksprovider/workflow"
 	"github.com/banzaicloud/pipeline/internal/cluster/distribution/eks/eksworkflow"
+	"github.com/banzaicloud/pipeline/internal/global"
+	"github.com/banzaicloud/pipeline/internal/kubernetes"
+	"github.com/banzaicloud/pipeline/pkg/kubernetes/custom/npls"
 )
 
+// CF stack name prefix
+const StackNamePrefix = "pipeline-eks-nodepool-"
```
sagikazarmark, line 39: Does this have to be exported? Everything should be unexported by default, unless needed somewhere else.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -16,27 +16,48 @@ package eksadapter
...
+
 type nodePoolManager struct {
-	workflowClient client.Client
-	enterprise     bool
+	workflowClient       client.Client
+	enterprise           bool
+	dynamicClientFactory kubernetes.DynamicClientFactory
+	aWSSessionFactory    workflow.AWSSessionFactory
```
sagikazarmark, line 45:
```suggestion
	awsSessionFactory    workflow.AWSSessionFactory
```

sagikazarmark: I'd copy the interface instead of importing it.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +115,95 @@ func (n nodePoolManager) UpdateNodePool(
 func generateNodePoolStackName(clusterName string, poolName string) string {
 	return "pipeline-eks-nodepool-" + clusterName + "-" + poolName
 }
+
+// ListNodePools is for listing node pools from CloudsetFormation and NodePoolLabelSets
+func (n nodePoolManager) ListNodePools(
+	ctx context.Context,
+	c cluster.Cluster,
+) ([]eks.NodePool, error) {
```
sagikazarmark, line 123: Looks like this could fit into a single line.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -94,3 +115,95 @@ func (n nodePoolManager) UpdateNodePool(
 func generateNodePoolStackName(clusterName string, poolName string) string {
 	return "pipeline-eks-nodepool-" + clusterName + "-" + poolName
 }
+
+// ListNodePools is for listing node pools from CloudsetFormation and NodePoolLabelSets
```
sagikazarmark, line 119:
```suggestion
// ListNodePools lists node pools.
```

internal/cluster/distribution/eks/service.go
```diff
@@ -26,6 +26,9 @@ type Service interface {
 	//
 	// This method accepts a partial body representation.
 	UpdateNodePool(ctx context.Context, clusterID uint, nodePoolName string, nodePoolUpdate NodePoolUpdate) (string, error)
+
+	// List NodePools
```
sagikazarmark, line 30:
```suggestion
	// ListNodePools lists node pools.
```

internal/cluster/distribution/eks/service.go
```diff
@@ -106,3 +132,38 @@ func (s service) UpdateNodePool(
 
 	return s.nodePoolManager.UpdateNodePool(ctx, c, nodePoolName, nodePoolUpdate)
 }
+
+// ListNodePools lists all node pools in a cluster
```
sagikazarmark, line 136:
```suggestion
// ListNodePools lists all node pools in a cluster.
```

internal/cluster/distribution/eks/service.go
```diff
@@ -106,3 +132,38 @@ func (s service) UpdateNodePool(
...
+func (s service) ListNodePools(
```
sagikazarmark, line 137: Maybe this could fit into one line.

internal/cluster/service.go
```diff
@@ -201,6 +201,9 @@ type Service interface {
 
 	// DeleteNodePool deletes a node pool from a cluster.
 	DeleteNodePool(ctx context.Context, clusterID uint, name string) (deleted bool, err error)
+
+	// ListNodePools list nodepools from cluster
```
sagikazarmark, line 205:
```suggestion
	// ListNodePools list nodepools from cluster.
```

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
sagikazarmark: Any chance we could write some unit tests for this? Using kube fake client and some sort of mock for the cf client? Unit test would be a great start, functional tests would be harder now.

apis/pipeline/pipeline.yaml
pregnor: Fixed in 2f0b6933c42b9be148fd286d10ef06ea928453ef.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
pregnor: Fixed in 0805a891374b5af9ef3e2af8b4628bccb4284d65 and b7680606f1d23f944d37cfb70a9140b2845814a1 respectively.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
pregnor: Fixed in b222234fa4df76dfdef041d7367dbae0a048f662

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
pregnor: Fixed in 1427f4df2964a33261074d8ebcc49a6ecf91b105.

apis/pipeline/pipeline.yaml
```diff
@@ -3553,6 +3553,25 @@ paths:
...
+        get:
+            operationId: ListNodePools
+            summary: List node pools
+            security:
+                - bearerAuth: []
+            tags:
+                - clusters
+            responses:
+                200:
+                    description: Node pools listed
```
sagikazarmark, line 3565:
```suggestion
                    description: Node pool list
```

apis/pipeline/pipeline.yaml
```diff
@@ -3553,6 +3553,25 @@ paths:
...
+                    content:
+                        application/json:
+                            schema:
+                                type: array
+                                items:
+                                    $ref: '#/components/schemas/ListNodePoolsResponse'
```
sagikazarmark, line 3571: I think extracting this as `ListNodePoolsResponse` would be more correct.

apis/pipeline/pipeline.yaml
```diff
@@ -4516,6 +4535,37 @@ components:
                     description: Node pool update process ID.
                     type: string
 
+        ListNodePoolsResponse:
```
sagikazarmark, line 4538: This should probably be some generic NodePool object (that could be used for a potential GET as well). This is related for the comment above.

internal/cluster/clusteradapter/distribution_eks.go
```diff
@@ -58,3 +58,18 @@ func (s eksService) UpdateNodePool(ctx context.Context, clusterID uint, nodePool
...
 }
+
+// ListNodePools lists node pools from a cluster.
+func (s eksService) ListNodePools(ctx context.Context, clusterID uint) (nodePoolList cluster.RawNodePoolList, err error) {
+	nodePools, err := s.service.ListNodePools(ctx, clusterID)
+	if err != nil {
+		return nil, errors.WrapWithDetails(err, "listing node pools through EKS service failed", "clusterID", clusterID)
+	}
+
+	response := make(map[string]interface{}, len(nodePools))
```
sagikazarmark, line 69: Using a map makes the order inconsistent. Can we avoid it? What's the reason behind using a map here (it's not immediately obvious).

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
```diff
@@ -93,5 +120,141 @@ func (n nodePoolManager) UpdateNodePool(
...
+	}
+
+	awsClient, err := n.awsFactory.New(c.OrganizationID, c.SecretID.ResourceID, c.Location)
+	if err != nil {
+		return nil, errors.WrapWithDetails(err, "creating aws factory failed", "cluster", c)
+	}
+
+	cfClient := n.cloudFormationFactory.New(awsClient, nil)
+	describeStacksInput := cloudformation.DescribeStacksInput{}
+	nodePools := make([]eks.NodePool, len(nodePoolNames))
```
sagikazarmark, line 149: I think it's nicer to define the slice with 0 len and the appropriate capacity and use `append`, that way you don't have to track the index. Since the capacity will never run out, appending has no performance penalty of reslicing.

internal/cluster/distribution/eks/eksadapter/node_pool_store.go
```diff
@@ -62,3 +62,25 @@ func (s nodePoolStore) CreateNodePool(
...
+	err = s.db.Find(&nodePoolModels,
+		map[string]interface{}{
+			"cluster_id": clusterID,
+		},
+	).Error
+	if err != nil {
+		return nil, errors.WrapWithDetails(err, "fetching node pools from database failed", "clusterID", clusterID)
+	}
+
+	nodePoolNames = make([]string, len(nodePoolModels))
```
sagikazarmark, line 80: Same thing about slice appending.

apis/pipeline/pipeline.yaml
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b

apis/pipeline/pipeline.yaml
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b

apis/pipeline/pipeline.yaml
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b.

There was already a [`NodePool`](https://github.com/banzaicloud/pipeline/blob/f640984ef5dee7ae5770cc55c9ccc9741c0e760b/apis/pipeline/pipeline.yaml#L4452) object defined which is a `oneOf` [`EksNodePool`](https://github.com/banzaicloud/pipeline/blob/f640984ef5dee7ae5770cc55c9ccc9741c0e760b/apis/pipeline/pipeline.yaml#L4497) and possibly other provider specific node pool objects in the future which can share a generic [`GenericNodePool`](https://github.com/banzaicloud/pipeline/blob/f640984ef5dee7ae5770cc55c9ccc9741c0e760b/apis/pipeline/pipeline.yaml#L4456) base holding name, size and labels.

This was already present in the pre-feature code so I used those.

Missed the opportunity while I took over the feature code as I supposed the existing feature code implementation would have been based on such foundations.
Thanks for the note, it feels much better this way.

internal/cluster/clusteradapter/distribution_eks.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b.

No reason behind, this was an oversight from me while taking over the feature code.

internal/cluster/distribution/eks/eksadapter/node_pool_manager.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b.

internal/cluster/distribution/eks/eksadapter/node_pool_store.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/2ad4cafb067c442a1cfbf1b9cd4071e298b269c8..f640984ef5dee7ae5770cc55c9ccc9741c0e760b.

Side-note: also fixed the 
```
database.pipeline.amazon_node_pools.cluster_id != database.pipeline.clusters.id database.pipeline.amazon_node_pools.cluster_id == database.pipeline.amazon_eks_clusters.id (in case of EKS clusters)
```
issue here at the node pool name listing to retrieve the correct cluster node pool models from the database.

I noticed I left a log line in the reivew fix force-with-lease push, so I fixed that here: https://github.com/banzaicloud/pipeline/compare/f640984ef5dee7ae5770cc55c9ccc9741c0e760b..9a918581fe55ba11f2ab3761d29d9f97eee75ed8.

pregnor:
Added more unit tests to the API Cluster Node Pool List code flow in the following commits:
https://github.com/banzaicloud/pipeline/pull/2965/commits/68261a777ee232510f4b82da7e54ac4c095c884a (new content is in the `internal/cluster/distribution/eks/eksadapter/node_pool_store_test.go` file and corresponding mock generation directives and generated mocks)
https://github.com/banzaicloud/pipeline/pull/2965/commits/f2b4ea59b4a55995bdbcdd3274a50e012f8ddd2f (the entire content of the commit is new)

internal/cluster/clusteradapter/distribution_eks_test.go
```diff
@@ -0,0 +1,149 @@
...
+		expectedNodePools cluster.RawNodePoolList
+		isExpectingError  bool
+	}
+	type testCaseType struct {
+		caseName              string
+		constructionArguments constructionArgumentType
+		functionCallArguments functionCallArgumentType
+		functionCallResult    functionCallResultType
+		setupMockFunction     func(constructionArgumentType, functionCallArgumentType)
+	}
```
sagikazarmark, line 46: While I understand what these structs are used for, I find it harder to read code like this.

For example: the testCaseType is never reused so it could be defined as an inline type in the test case list definition.

I guess this is a question of taste as well, but I usually prefer just plain local variables for the rest.

That being said, I'd be open to hearing more arguments while this style/practice is better from a structural point of view.

For now, it's okay.

pregnor: I have refined my approach in the [latest rebase](https://github.com/banzaicloud/pipeline/compare/e19e85be0dba082a8c755a58d53a4dfb4270fc43..f00f068d4e08f9e383f9d21c852421e0053ef3e2).

I kept the `constructionArgumentType` and `functionCallArgumentType` and eliminated/inlined the rest of the test types.
The reason behind these is the repeated use in the test case and the setup function so some syntactic code repetition can be omitted this way which IMO improves readability as only 2 parameters are passed into the `setupMockFunction` instead of arbitrary number of parameters - often 5+ - and also I like the origin/purpose encapsulation of this approach when reading the test code.

Let me know what you think about this mixed approach and whether I should push further and remove those two test type definitions as well.

sagikazarmark: Looks good to me. I'm all for trying new approaches. If it indeed improves the readability/shortens the bootstrap code, let's do it!

PR #2969, opened 2020-06-05
EKS: remove default image param from node pool CF

### LINKED ISSUE #2968: Remove EKS image fallback from node pool CF template
The fallback image parameter requires ssm permission which seems unnecessary, given we always provide an image ID.

Remove it from the template.

We could grant the necessary permissions, but as said above, we probably don't use this fallback anyway.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2968 

### What's in this PR?
Remove default image param from node pool CloudFormation template since it's unnecessary because Pipeline always sets default image in case it's not provided while it requires additional access rights for user.

### Checklist

### DIFF
--- a/templates/eks/amazon-eks-nodepool-cf.yaml
+++ b/templates/eks/amazon-eks-nodepool-cf.yaml
@@ -79,13 +79,7 @@ Parameters:
 
   NodeImageId:
       Type: String
-      Default: ""
       Description: (Optional) Specify your own custom image ID. This value overrides any AWS Systems Manager Parameter Store value specified above.
-
-  NodeImageIdSSMParam:
-      Type: "AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>"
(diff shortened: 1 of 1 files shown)


templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -79,14 +79,8 @@ Parameters:
 
   NodeImageId:
       Type: String
-      Default: ""
       Description: (Optional) Specify your own custom image ID. This value overrides any AWS Systems Manager Parameter Store value specified above.
```
sagikazarmark, line 82:
```suggestion
      Description: Specify your own custom image ID. This value overrides any AWS Systems Manager Parameter Store value specified above.
```

PR #2979, opened 2020-06-11
AWS PKE Subnet create / delete rework

### LINKED ISSUE #2931: Random AZ used when creating a subnet automatically
**Describe the bug**
When creating a cluster without manual network setup, instead of the chosen AZ we use a random one (actually mostly the first one, which is usually **a**), leading to all kinds of issues (like not available instances in an AZ)

1. Create a new cluster
1. Select a zone other than *a*

**Expected behavior**
The cluster (node pools) should be created in the chosen zone.

**Additional context**
This looks like a tech debt (unfinished work), but is classified as a bug, because causes all kinds of issues.

tarokkk:
| Q               | A
| --------------- | ---
| Bug fix?        | kindof
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2931

### What's in this PR?
This PR move subnet creation to dedicated activities instead of using the first subnet of the VPC.

How it worked before?
When you created a PKE cluster there were 2 options:
- Built-in subnet creation - this **always** used the **first** availability zone to create subnet (usually xx-yyyyy-1a)
- Advanced networking with preconfigured subnets (no subnet creation)

How this PR change this?

When you don't specify subnets in the node pools Pipeline will grab all specified Availability Zone from the given node pools and create a subnet for each. These subnets are paired based on the availability zone requirements for node pools. It has a "dummy" CIDR allocation logic.

```
	for zone, _ := range availabilityZoneSet {
		availabilityZoneMap[zone] = fmt.Sprintf("192.168.%d.0/24", id*16)
	}
```

Limitations
- Single Master PKE will only use the first subnet as it is a single EC2 instance not an ASG

### Checklist


internal/providers/pke/pkeworkflow/create_cluster.go
```diff
@@ -148,6 +166,45 @@ func (w CreateClusterWorkflow) Execute(ctx workflow.Context, input CreateCluster
...
+	for _, np := range nodePools {
+		// We only look AZ when no subnet is set
+		if len(np.Subnets) < 1 {
+			for _, az := range np.AvailabilityZones {
+				availabilityZoneSet[az] = true
+			}
+		}
+	}
+	// Create AZ and Address map
+	availabilityZoneMap := make(map[string]string)
```
stoader, line 202:
```suggestion
	availabilityZoneMap := make(map[string]string, len(availabilityZoneSet)
```

internal/providers/pke/pkeworkflow/create_cluster.go
```diff
@@ -148,6 +166,45 @@ func (w CreateClusterWorkflow) Execute(ctx workflow.Context, input CreateCluster
...
+	id := 0
+	for zone, _ := range availabilityZoneSet {
+		availabilityZoneMap[zone] = fmt.Sprintf("192.168.%d.0/24", id*16)
```
stoader, line 205:
```suggestion
		availabilityZoneMap[zone] = fmt.Sprintf("192.168.%d.0/24", id*16)
		id++
```

internal/providers/pke/pkeworkflow/create_cluster.go
```diff
@@ -148,6 +166,45 @@ func (w CreateClusterWorkflow) Execute(ctx workflow.Context, input CreateCluster
...
+		if np.Master {
+			master = np
+			if len(np.AvailabilityZones) <= 0 || np.AvailabilityZones[0] == "" {
+				return errors.Errorf("missing availability zone for nodepool %q", np.Name)
+			}
+			break
+		}
+	}
+	// Collect relevant AZs from NodePools without subnets
+	availabilityZoneSet := make(map[string]bool)
```
stoader, line 192: I believe `map[string]struct{}` uses less memory

internal/providers/pke/pkeworkflow/create_cluster.go
```diff
@@ -148,6 +166,45 @@ func (w CreateClusterWorkflow) Execute(ctx workflow.Context, input CreateCluster
...
+		}
+	}
+
+	// Collect all AZs
+	var master NodePool
+	for _, np := range nodePools {
+		if np.Master {
+			master = np
+			if len(np.AvailabilityZones) <= 0 || np.AvailabilityZones[0] == "" {
+				return errors.Errorf("missing availability zone for nodepool %q", np.Name)
```
stoader, line 186: You could use `errors.NewWithDetails` instead of `errors.Errorf`

internal/providers/pke/pkeworkflow/create_subnet_activity.go
```diff
@@ -0,0 +1,170 @@
...
+
+		_, err := cloudformationClient.CreateStack(createStackInput)
+		if err != nil {
+			return nil, errors.WrapIf(err, "create stack failed")
+		}
+
+		describeStacksInput := &cloudformation.DescribeStacksInput{StackName: stackName}
+		err = eksWorkflow.WaitUntilStackCreateCompleteWithContext(cloudformationClient, ctx, describeStacksInput)
+
+		if err != nil {
```
stoader, line 132: In case of error you can collect the stack events that error out with `NewAwsStackFailure` and return that.

internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
```diff
@@ -51,7 +52,7 @@ type CreateWorkerPoolActivityInput struct {
 	Pool                      NodePool
 	VPCID                     string
 	VPCDefaultSecurityGroupID string
-	SubnetID                  string
+	SubnetID                  []string
```
sagikazarmark, line 55:
```suggestion
	SubnetIDs                  []string
```

?

internal/providers/pke/pkeworkflow/delete_subnet_activity.go
```diff
@@ -0,0 +1,119 @@
...
+
+	clusterName := c.GetName()
+	stackName := "pke-subnet-" + clusterName + "-" + input.AvailabilityZone
+	stackInput := &cloudformation.DeleteStackInput{
+		StackName: &stackName,
+	}
+
+	_, err = cfClient.DeleteStack(stackInput)
+	if err, ok := err.(awserr.Error); ok {
+		switch err.Code() {
```
bonifaido, line 69: This `switch` case seems to be redundant.

internal/providers/pke/pkeworkflow/create_subnet_activity.go
```diff
@@ -0,0 +1,170 @@
...
+		"availabilityZone", input.AvailabilityZone,
+		"secret", input.SecretID,
+	)
+
+	client, err := a.awsClientFactory.New(input.OrganizationID, input.SecretID, input.Region)
+	if err = errors.WrapIf(err, "failed to create AWS session"); err != nil {
+		return nil, err
+	}
+
+	buf, err := ioutil.ReadFile("templates/pke/subnet.cf.yaml")
```
bonifaido, line 85: Probably we need to generalize `getEksCloudFormationTemplate` and reuse it here, see: https://github.com/banzaicloud/pipeline/blob/master/internal/cluster/distribution/eks/eksprovider/workflow/templates.go#L37

internal/providers/pke/pkeworkflow/create_cluster.go
```diff
@@ -198,26 +255,37 @@ func (w CreateClusterWorkflow) Execute(ctx workflow.Context, input CreateCluster
...
+
+				AWSActivityInput: awsActivityInput,
+				ClusterID:        input.ClusterID,
+				ClusterName:      input.ClusterName,
+				VpcID:            vpcOutput["VpcId"],
+				RouteTableID:     vpcOutput["RouteTableId"],
+				Cidr:             ip,
+				AvailabilityZone: zone,
+			}
+			ctx := workflow.WithActivityOptions(ctx, aoWithHeartbeat)
```
bonifaido, line 274: `CreateSubnetActivityName` is started with a Cadence `HeartBeatTimeout` but there is no heart beating in that activity.

internal/cluster/distribution/cloud_formation.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+	"net/url"
+	"text/template"
+
+	"github.com/banzaicloud/pipeline/internal/global"
+
+	"emperror.dev/errors"
+)
+
+// getEksCloudFormationTemplate returns CloudFormation template with given name
```
stoader, line 30:
```suggestion
// GetCloudFormationTemplate returns CloudFormation template with given name
```

orymate:
I've tested, fixed, and tested the creation, update and deletion of ha and single master clusters in new and existing vpc/subnets successfully.

PR #2991, opened 2020-07-03
Rewrite audit log

### LINKED ISSUE #2986: Rewrite Audit log

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2986

### What's in this PR?
Rewrite the audit log middleware addressing a number of different issues with it, as well as adding the possibility to send audit logs to the standard output.

New feature: dump audit events to standard output (disabled by default). The amount of details about the request can be controlled by the verbosity setting (see the code for details).

### Create a cluster

**Steps:**
1. Create a new cluster

**Expected output:**
- The cluster JSON should be in the database
- Based on verbosity, certain details should be included about the request

### Create a PKE cluster

**Steps:**
1. Create a new PKE cluster
1. Wait for at least the masters to join

**Expected output:**
- The ready request should not have body in the database

### Create a secret

**Steps:**
1. Create a new secret
1. Get/Download the secret
1. Update the secret

**Expected output:**
- Secret requests should have no body at all in the database


internal/platform/gin/auditlog/auditlogdriver/log.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if d.config.Verbosity >= 4 {
+		data["http.responseTime"] = entry.HTTP.ResponseTime
+		data["http.responseSize"] = entry.HTTP.ResponseSize
+		data["http.requestBody"] = entry.HTTP.RequestBody
+
+		if len(entry.HTTP.Errors) > 0 {
+			data["http.errors"] = entry.HTTP.Errors
+		}
+	}
```
pepov, line 77: I would consider 1,2,3 as simply metadata (and would add response time and reponse size as well) and enable it with a single flag. Then use the verbose flag to enable request (1) and response (2) body as well.

cmd/pipeline/main.go
```diff
@@ -587,10 +586,34 @@ func main() {
...
+			driver = append(driver, auditlogdriver.NewDatabaseDriver(db))
+		}
+
+		engine.Use(auditlog.Middleware(
+			driver,
+			auditlog.WithUserIDExtractor(auth.GetCurrentUserID),
+			auditlog.WithSensitivePaths([]*regexp.Regexp{
+				regexp.MustCompile("/auth/dex(?:/.*)?"),
+				regexp.MustCompile("/.*/api/v1/orgs/.*/secrets(?:/.*)?"),
+				regexp.MustCompile("/.*/api/v1/orgs/.*/clusters/.*/pke/ready"),
```
pepov, line 611: making these configurable by the user (with these sane defaults)?

config/config.yaml.dist
```diff
@@ -67,6 +67,17 @@ cadence:
...
+#    enabled: true
+#
+#    driver:
+#        log:
+#            enabled: false
+#            verbosity: 1
+#
+#        database:
+#            enabled: true
+
```
pregnor, line 80: **Question**: would it be worth considering to add a config section to the developer configuration (`config.dev.yaml`) with auditlog.driver.log.enabled set to true? (audit logging by default for evaluation purposes)

cmd/pipeline/config.go
```diff
@@ -172,8 +190,4 @@ func configure(v *viper.Viper, p *pflag.FlagSet) {
 
 	v.SetDefault("spotmetrics::enabled", false)
 	v.SetDefault("spotmetrics::collectionInterval", 30*time.Second)
-
-	v.SetDefault("audit::enabled", true)
-	v.SetDefault("audit::headers", []string{"secretId"})
```
pregnor, line 177: **Note**: this configuration option seems to be completely removed from the new code. This is just an FYI notice to make sure this decision was deliberate and has no problematic consequences - I don't see any, but my domain knowledge is heavily lacking to judge it properly.

internal/platform/gin/auditlog/auditlogdriver/db_test.go
```diff
@@ -0,0 +1,95 @@
...
+	t.Cleanup(func() {
+		_ = db.Close()
+	})
+
+	return db
+}
+
+func TestDatabaseDriver(t *testing.T) {
+	entry := auditlog.Entry{
+		Time:          time.Date(1984, time.April, 4, 0, 0, 0, 0, time.UTC),
```
pregnor, line 46: **Question**: just curious, whose birth date is this? 🙂

internal/platform/gin/auditlog/auditlogdriver/model.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2018 Banzai Cloud
+// Copyright © 2020 Banzai Cloud
```
pregnor, line 1: **Note**: changing copyright notices to current year can be done en masse with a simple script, I have some previous experiences with it, let me know if you would like to see a PR doing just that.

internal/platform/gin/auditlog/middleware.go
```diff
@@ -0,0 +1,188 @@
...
+
+		// Consider making this configurable if you need to log unauthorized requests,
+		// but keep in mind that in case of a public installation it's a potential DoS attack vector.
+		if c.Writer.Status() == http.StatusUnauthorized {
+			return
+		}
+
+		entry.HTTP.StatusCode = c.Writer.Status()
+		entry.HTTP.ResponseSize = c.Writer.Size()
+		entry.HTTP.ResponseTime = int(options.clock.Since(entry.Time).Nanoseconds() / 1000 / 1000) // ms
```
pregnor, line 162: **Question**: what is the reason behind using `time.Duration{}.Nanoseconds() / 1000 / 1000` manual division instead of `time.Duration{}.Milliseconds()`?

cmd/pipeline/main.go
sagikazarmark: There are certain URLs that have to be filtered (body not logged). But I don't see a use case for the user specifying these URLs. Do you have one in mind?

internal/platform/gin/auditlog/auditlogdriver/log.go
sagikazarmark: Response body is not recorded. For development, we never needed it. For auditing, I don't see why it's relevant: the state is there, we are mostly concerned about the request part.

I would consider adding the response body as a separate proposal. Without that, if I understand correctly, you propose moving response time and size to level 3. Is that correct?

cmd/pipeline/main.go
```diff
@@ -544,9 +544,8 @@ func main() {
 	router.Use(ocmux.Middleware())
 
 	// These two paths can contain sensitive information, so it is advised not to log them out.
-	skipPaths := config.Audit.SkipPaths
 	engine.Use(correlationid.Middleware())
-	engine.Use(ginlog.Middleware(logrusLogger, skipPaths...))
+	engine.Use(ginlog.Middleware(logrusLogger, []string{"/auth/dex/callback", "/pipeline/api"}...))
```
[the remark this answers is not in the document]
sagikazarmark: The skip path configuration has been abused before this change: although it belonged to the audit log, we used it for filtering the access log as well. They are not the same use case though, in fact, I don't think we need to skip any paths in the access log middleware....it's access log after all. It doesn't log the request body. Also, they filter different paths now.

But changing that is out of scope for this PR, so for now I replaced it with the default value of the configuration. (Feel free to address that in a separate issue/PR though)

The reason why you see two places with that change is that we have an "internal API" that's not really used anymore (technically hollowtrees relies on it, but it's essentially dead), it was badly constructed in the first place, which is why it instantiates similar middleware. Addressing this is also out of scope for this PR in my opinion. (And I don't really care if it diverges from the original one, since it doesn't even serve those paths, so it's essentially non-functional)

I also think that exposing the skipped paths as a configuration was a mistake. I don't see any use case for the *user* to specify which paths should not have the request body recorded. It's rather internal to the application.

config/config.yaml.dist
sagikazarmark: In the past we mostly used the "audit log" feature to check request bodies in the database. I don't think it's a good idea to log request bodies (to the stdout), so from a developer perspective, the database will always be the goto source.

The "regular" logs are already cluttered with lots of stuff, this would rather increase the noise in my opinion.

internal/platform/gin/auditlog/auditlogdriver/db_test.go
sagikazarmark: Hint:

> To the past, or to the future. To an age when thought is free. From the Age of Big Brother, from the Age of the Thought Police, from a dead man - greetings!

😉

cmd/pipeline/config.go
sagikazarmark: We don't use that header anymore, as such, whitelisting it doesn't make any sense. But you are right, this removes the header logging functionality.

internal/platform/gin/auditlog/auditlogdriver/model.go
sagikazarmark: This was done by the IDE automatically (I actually copied the content to a new file which includes the header with the new year)

internal/platform/gin/auditlog/middleware.go
sagikazarmark: Dunno, copy-pasta. Will change it.

cmd/pipeline/main.go
pepov: I had this in mind: having an issue where the outcome is to instantly add more urls to this list, or simply to fix one of them, where otherwise you would need a new build instead of a reconfiguration.

internal/platform/gin/auditlog/auditlogdriver/log.go
pepov: sounds like a deal, but shouldn't we make verbosity 3 by default?

pepov: however from auditing point of view, why wouldn't response body also be helpful, i.e. to see what information had been leaked in a specific case?

PR #2995, opened 2020-07-04
Fix amazon secret default region

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Use the configured default region for amazon secret verification

### Why?
Depending on the account, the hard-coded default region might not be available everywhere.

### DIFF
--- a/cmd/pipeline/main.go
+++ b/cmd/pipeline/main.go
@@ -274,6 +274,7 @@ func main() {
 	secretStore := secretadapter.NewVaultStore(vaultClient, "secret")
 	pkeSecreter := pkesecret.NewPkeSecreter(vaultClient, commonLogger)
 	secretTypes := types.NewDefaultTypeList(types.DefaultTypeListConfig{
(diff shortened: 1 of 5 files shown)


sagikazarmark:
> Maybe the default config (config.yaml.dist) should be adjusted accordingly?

@siliconbrain  What do you mean? It was just a refactor of the structs, but the config is unchanged.

siliconbrain:
> > Maybe the default config (config.yaml.dist) should be adjusted accordingly?
> 
> @siliconbrain What do you mean? It was just a refactor of the structs, but the config is unchanged.

Yeah, but the field is now required instead of being optional with a default value. Maybe we should bring attention to this in the config.

sagikazarmark:
I don't think we do that anywhere in the config. Since there is a default value, the only time users will notice that it became required is when they deliberately set it to an empty string. (And it should already have been required, so this is rather a bug than a new feature).

PR #3011, opened 2020-07-14
chore(integratedservices): regenerated mock file, added ApplyDeploymentV3

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Run `make generate` and committed the resulting diff.

### Why?

While I was working on another issue, I had to regenerate some mock code and I noticed a totally unrelated generated change which is missing from the master branch.

### Additional context

What is our stance on adding a CI/Action step which would check whether the code generation make directive on the current commit's code would result in a non-zero diff (and fail if it is a non-zero diff)?

### Checklist

- [X] Implementation tested (with at least ~one cloud provider~ tests were run locally)

### DIFF
--- a/internal/integratedservices/services/zz_generated.mock.go
+++ b/internal/integratedservices/services/zz_generated.mock.go
@@ -6,7 +6,8 @@ package services
 
 import (
 	"context"
-	"github.com/banzaicloud/pipeline/pkg/helm"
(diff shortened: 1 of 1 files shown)


pepov:
👍

PR #3022, opened 2020-07-19
Regenerate dex proto

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Regenerate dex proto

### Why?
It was different from the upstream file


apis/dex/api.proto
```diff
@@ -1,9 +1,8 @@
 syntax = "proto3";
+option java_package = "com.coreos.dex.api";
```
bonifaido, line 2: I think this option is irrelevant for us and can be removed entirely.

sagikazarmark: I think it's required by certain proto lint rules. In any case, it should be solved in the upstream repo.

(BTW this file will go away once we tag a new api version in the dex repo. Then we can depend on the api package directly)

PR #3049, opened 2020-07-26
Better fallback image selection logic

### LINKED ISSUE #3025: Implement image selector service for AWS PKE
As per our decision, implement an image selector service based on the following interface:

```go
// ImageSelector chooses an image based on a set of criteria.
type ImageSelector interface {
    SelectImage(ctx context.Context, criteria ImageSelectionCriteria) (string, error)
}

// ImageSelectionCriteria contains all parameters for selecting an image.
type ImageSelectionCriteria struct {
    Region            string
    InstanceType      string
    PKEVersion        string
    KubernetesVersion string
    OperatingSystem   string
    ContainerRuntime  string
}
```

The image selector should select a default image based on the provided criteria.

Additional rules:
- if the InstanceType is a GPU instance, try to find a GPU image (see #3050)
- If there is no GPU image, try a regular image
- if there are no matching images, return an error

The image selector should try looking for images in the following sources:

- cloudinfo
- user config (see #3019)
- hard-coded defaults

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3025

### What's in this PR?
This PR implements an extensible image selector service for PKE on AWS.

### Why?
We need a variety of new options for image selection. As a first step, this PR introduces a new way for image selection.


internal/cluster/distribution/pke/pkeaws/image_selector.go
```diff
@@ -0,0 +1,133 @@
...
+		errorHandler: errorHandler,
+	}
+}
+
+func (s *ImageSelectorChain) getLogger() Logger {
+	s.loggerOnce.Do(func() {
+		if s.logger == nil {
+			s.logger = NoopLogger{}
+		}
+	})
```
pregnor, line 85: **Question**: what is the pro for initializing a NoopLogger here as opposed to the constructor (by checking the function argument for `nil`) where no synchronization would be required because only the constructor itself has reference to the newly created object?

(Same for https://github.com/banzaicloud/pipeline/pull/3049/files#diff-93108070aa9a2926c7ddda88735708c4R95)

sagikazarmark: It's possible to use the struct without the constructor, if you don't want to specify a logger and an error handler.

sagikazarmark: I even considered using public fields for the logger and the error handler, but I settled with public fields and a constructor for now. We can always change it (in a backward compatible manner) later.

internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (s CloudinfoImageSelector) SelectImage(ctx context.Context, criteria pkeaws.ImageSelectionCriteria) (string, error) {
+	if s.client == nil {
+		return "", errors.New("cloudinfo: client not configured")
+	}
+
+	kubeVersion, err := semver.NewVersion(criteria.KubernetesVersion)
+	if err != nil {
+		return "", errors.WrapWithDetails(err, "cloudinfo", "kubernetesVersion", criteria.KubernetesVersion)
```
orymate, line 48:
```suggestion
		return "", errors.WrapWithDetails(err, "invalid Kubernetes version requested", "kubernetesVersion", criteria.KubernetesVersion)
```

internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
```diff
@@ -0,0 +1,84 @@
...
+		return "", errors.WrapWithDetails(
+			pkeaws.ImageNotFoundError, "cloudinfo",
+			"cloudProvider", cloudProvider,
+			"service", serviceName,
+			"region", criteria.Region,
+			"getImagesOpts", opts,
+		)
+	}
+
+	return images[0].Name, nil
```
orymate, line 83: do we expect cloudinfo to return the images in a specific order?

internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return "", errors.WrapIfWithDetails(
+			err, "cloudinfo",
+			"cloudProvider", cloudProvider,
+			"service", serviceName,
+			"region", criteria.Region,
+			"getImagesOpts", opts,
+		)
+	}
+	if len(images) <= 0 {
```
orymate, line 73:
```suggestion
	if len(images) == 0 {
```

internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
sagikazarmark: We expect cloudinfo to return one image, but it doesn't matter at the moment. This PR is about refactoring the current logic into the new structure. Later PRs can improve the behavior.

PR #3052, opened 2020-07-27
Cleanup helm2

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Cleanup helm2 from the codebase

### Why?
Helm 3 migration is complete

### Additional context
Many thanks to @pepov and @lpuskas who did the heavy lifting and from whom I stole the glory of removing the dead code.


pepov:
can we get rid of `global.config.Helm` from `internal/global/config.go`?

sagikazarmark:
As far as I can tell, it's still in use. For example in `internal/helm/testing/setup.go:78`: `HelmHome` function

PR #3057, opened 2020-07-27
Integrate new image selector into PKE AWS create workflow

### LINKED ISSUE #3030: Replace current image selection logic with image selector service
Use the new image selector service instead of the current image selection logic

Select images on a node pool basis

Requires #3025

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3030

### What's in this PR?
Integrate new image selector into PKE AWS create workflow

### Why?
This is an intermediate step to better support GPU nodes.

### Checklist

### DIFF
--- a/internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
+++ b/internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
@@ -84,5 +84,18 @@ func (s CloudinfoImageSelector) SelectImage(ctx context.Context, criteria pkeaws
 		)
 	}
 
+	// As a result of a bug in cloudinfo,
+	// the returned item might be empty
+	// See https://github.com/banzaicloud/cloudinfo/pull/356
+	if images[0].Name == "" {
+		return "", errors.WithDetails(
+			errors.WithStack(pkeaws.ImageNotFoundError),
+			"cloudProvider", cloudProvider,
+			"service", serviceName,
+			"region", criteria.Region,
+			"getImagesOpts", opts,
+		)
+	}
+
 	return images[0].Name, nil
 }
--- a/cmd/worker/aws.go
+++ b/cmd/worker/aws.go
@@ -18,16 +18,16 @@ import (
 	"go.uber.org/cadence/activity"
 	"go.uber.org/cadence/workflow"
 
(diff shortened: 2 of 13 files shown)


internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
```diff
@@ -84,5 +84,18 @@ func (s CloudinfoImageSelector) SelectImage(ctx context.Context, criteria pkeaws
...
+	if images[0].Name == "" {
+		return "", errors.WithDetails(
+			errors.WithStack(pkeaws.ImageNotFoundError),
+			"cloudProvider", cloudProvider,
+			"service", serviceName,
+			"region", criteria.Region,
+			"getImagesOpts", opts,
+		)
+	}
+
```
pregnor, line 99: **Optional**: we **MAY** consider adding the condition (`images[0].Name == ""`) to the previous `if` statement as both branches are supposed to return the same values and that way code duplication can be avoided.

I think the temporal state of this case and corresponding external module comment is more explicit/visible this way which is a plus, but I don't like the duplicated branch.

I guess it is opinionated, so it's totally up for you to choose, I just wanted to remark it if it was not deliberately done this way.

sagikazarmark: I deliberately added a separate branch, because it's easier to clean up once the bug is resolved.

internal/cluster/distribution/pke/pkeaws/pkeawsadapter/image_selector.go
```diff
@@ -84,5 +84,18 @@ func (s CloudinfoImageSelector) SelectImage(ctx context.Context, criteria pkeaws
 		)
 	}
 
+	// As a result of a bug in cloudinfo,
+	// the returned item might be empty
+	// See https://github.com/banzaicloud/cloudinfo/pull/356
```
orymate, line 89: I wouldn't handle this case here, at least remove this check once we are sure that the bug is fixed.

sagikazarmark: That's the plan, but I didn't have much of a choice if I wanted to let this PR in.

PR #3059, opened 2020-07-27
Integrate new image selector into Amazon EKS create workflow

### LINKED ISSUE #3029: Replace current image selection logic with image selector service
Use the new image selector service instead of the current image selection logic

Select images on a node pool basis

Requires #3026

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3029

### What's in this PR?
Integrate new image selector into Amazon EKS create workflow

### Why?
This is an intermediate step to better support GPU nodes.

### Checklist


internal/global/globaleks/image_selector.go
```diff
@@ -12,29 +12,32 @@
...
-	_, err = GetDefaultImageID(region, nonExistingVersion)
-	require.Error(t, err)
+// SetImageSelector configures an ImageSelector instance.
+func SetImageSelector(is eks.ImageSelector) {
+	imageSelectorMu.Lock()
+	defer imageSelectorMu.Unlock()
 
-	_, err = GetDefaultImageID(region, "INVALID")
-	require.Error(t, err)
+	imageSelector = is
```
pregnor, line 42: **Question**: what is the reason behind only allowing 1 image selector in this legacy part instead of an image selector chain?

PR #3060, opened 2020-07-28
Fix create nodepool validation

### LINKED ISSUE #3017: Node pool size is not validated properly
**Describe the bug**
It's possible to create a node pool (using the new API) that results in desired count out of the range of min..max in the ASG.

Create a node pool with invalid desired count.

**Expected behavior**
Pipeline should validate requests and should not allow such scenario.

**Additional context**
CF error:

```
Desired capacity:0 must be between the specified min size:1 and max size:9 (Service: AmazonAutoScaling; Status Code: 400; Error Code: ValidationError)
```

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3017 

### What's in this PR?
* Allow zero-sized autoscaled node pools (like at the time of cluster creation)
* Check if desired count is between the min and max size

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/node_pool.go
+++ b/internal/cluster/distribution/eks/node_pool.go
@@ -52,6 +52,14 @@ func (n NewNodePool) Validate() error {
 		if n.Autoscaling.MaxSize <= n.Autoscaling.MinSize {
 			violations = append(violations, "maximum autoscaling size cannot be lower than the minimum")
 		}
+
+		if n.Size < n.Autoscaling.MinSize {
+			violations = append(violations, "desired size cannot be lower than the autoscaling minimum")
+		}
+
+		if n.Size > n.Autoscaling.MaxSize {
+			violations = append(violations, "desired size cannot be higher than the autoscaling maximum")
+		}
 	} else if n.Size < 1 {
 		violations = append(violations, "size cannot be lower than one")
 	}
--- a/internal/cluster/distribution/eks/node_pool.go
+++ b/internal/cluster/distribution/eks/node_pool.go
@@ -45,8 +45,8 @@ func (n NewNodePool) Validate() error {
 	var violations []string
 
 	if n.Autoscaling.Enabled {
-		if n.Autoscaling.MinSize < 1 {
-			violations = append(violations, "minimum autoscaling size cannot be lower than one")
+		if n.Autoscaling.MinSize < 0 {
+			violations = append(violations, "minimum autoscaling size cannot be lower than zero")
 		}
 
 		if n.Autoscaling.MaxSize <= n.Autoscaling.MinSize {
@@ -54,11 +54,11 @@ func (n NewNodePool) Validate() error {
 		}
 
 		if n.Size < n.Autoscaling.MinSize {
-			violations = append(violations, "desired size cannot be lower than the autoscaling minimum")
+			violations = append(violations, "node pool size cannot be lower than the autoscaling minimum size")
 		}
 
 		if n.Size > n.Autoscaling.MaxSize {
-			violations = append(violations, "desired size cannot be higher than the autoscaling maximum")
+			violations = append(violations, "node pool size cannot be higher than the autoscaling maximum size")
 		}
 	} else if n.Size < 1 {
 		violations = append(violations, "size cannot be lower than one")
--- a/internal/cluster/distribution/eks/node_pool_test.go
+++ b/internal/cluster/distribution/eks/node_pool_test.go
@@ -39,6 +39,17 @@ func TestNodePoolSizeValidation(t *testing.T) {
 		assert.IsType(t, cluster.ValidationError{}, errors.Cause(err))
 	})
 
+	t.Run("PoolMinSizeNegative", func(t *testing.T) {
+		pool := base
+		pool.Size = 1
+		pool.Autoscaling.Enabled = true
+		pool.Autoscaling.MinSize = -1
+		pool.Autoscaling.MaxSize = 2
+
+		err := pool.Validate()
+		assert.IsType(t, cluster.ValidationError{}, errors.Cause(err))
+	})
+
 	t.Run("PoolSizeUnderMin", func(t *testing.T) {
 		pool := base
 		pool.Size = 0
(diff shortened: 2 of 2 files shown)


internal/cluster/distribution/eks/node_pool.go
```diff
@@ -45,13 +45,21 @@ func (n NewNodePool) Validate() error {
 	var violations []string
 
 	if n.Autoscaling.Enabled {
-		if n.Autoscaling.MinSize < 1 {
-			violations = append(violations, "minimum autoscaling size cannot be lower than one")
+		if n.Autoscaling.MinSize < 0 {
+			violations = append(violations, "minimum autoscaling size cannot be negative")
```
sagikazarmark, line 49:
```suggestion
			violations = append(violations, "minimum autoscaling size cannot be lower than zero")
```

internal/cluster/distribution/eks/node_pool.go
```diff
@@ -45,13 +45,21 @@ func (n NewNodePool) Validate() error {
...
 		}
 
 		if n.Autoscaling.MaxSize <= n.Autoscaling.MinSize {
 			violations = append(violations, "maximum autoscaling size cannot be lower than the minimum")
 		}
+
+		if n.Size < n.Autoscaling.MinSize {
+			violations = append(violations, "desired size cannot be lower than the autoscaling minimum")
```
sagikazarmark, line 57:
```suggestion
			violations = append(violations, "node pool size cannot be lower than the autoscaling minimum size")
```

internal/cluster/distribution/eks/node_pool.go
```diff
@@ -45,13 +45,21 @@ func (n NewNodePool) Validate() error {
...
+		}
+
+		if n.Size > n.Autoscaling.MaxSize {
+			violations = append(violations, "desired size cannot be higher than the autoscaling maximum")
```
sagikazarmark, line 61:
```suggestion
			violations = append(violations, "node pool size cannot be higher than the autoscaling maximum size")
```

internal/cluster/distribution/eks/node_pool_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	t.Run("PoolValidRange", func(t *testing.T) {
+		pool := base
+		pool.Size = 1
+		pool.Autoscaling.Enabled = true
+		pool.Autoscaling.MinSize = 1
+		pool.Autoscaling.MaxSize = 2
+
+		err := pool.Validate()
+		assert.Nil(t, err)
```
sagikazarmark, line 94:
```suggestion
		assert.NoError(t, err)
```

internal/cluster/distribution/eks/node_pool_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	t.Run("PoolZeroAutoscaling", func(t *testing.T) {
+		pool := base
+		pool.Size = 0
+		pool.Autoscaling.Enabled = true
+		pool.Autoscaling.MinSize = 0
+		pool.Autoscaling.MaxSize = 2
+
+		err := pool.Validate()
+		assert.Nil(t, err)
```
sagikazarmark, line 105:
```suggestion
		assert.NoError(t, err)
```

internal/cluster/distribution/eks/node_pool.go
```diff
@@ -45,13 +45,21 @@ func (n NewNodePool) Validate() error {
 	var violations []string
 
 	if n.Autoscaling.Enabled {
-		if n.Autoscaling.MinSize < 1 {
-			violations = append(violations, "minimum autoscaling size cannot be lower than one")
+		if n.Autoscaling.MinSize < 0 {
+			violations = append(violations, "minimum autoscaling size cannot be negative")
 		}
 
 		if n.Autoscaling.MaxSize <= n.Autoscaling.MinSize {
```
pregnor, line 52: **Note**: kind of out of scope, but while we are here: this condition disallows creating fixed size (`minSize == maxSize`) autoscaled node pools. It's worth checking whether this is deliberate. As long as we don't allow changing node pool size constraints it is OK IMO, but when that design changes this needs to be revised.
Also the current wording of the error doesn't include the case I just talked about which might be confusing for the reader.
(It surely confused me whether the condition or the message reflects the true intent and whichever makes the other one incorrect.)

internal/cluster/distribution/eks/node_pool_test.go
```diff
@@ -0,0 +1,116 @@
...
+	})
+
+	t.Run("PoolValidNoAS", func(t *testing.T) {
+		pool := base
+		pool.Size = 1
+		pool.Autoscaling.Enabled = false
+
+		err := pool.Validate()
+		assert.Nil(t, err)
```
sagikazarmark, line 114:
```suggestion
		assert.NoError(t, err)
```

internal/cluster/distribution/eks/node_pool_test.go
```diff
@@ -0,0 +1,116 @@
...
+		pool.Size = 0
+		pool.Autoscaling.Enabled = true
+		pool.Autoscaling.MinSize = 0
+		pool.Autoscaling.MaxSize = 2
+
+		err := pool.Validate()
+		assert.Nil(t, err)
+	})
+
+	t.Run("PoolValidNoAS", func(t *testing.T) {
```
sagikazarmark, line 108:
```suggestion
	t.Run("PoolValidNoAutoscaling", func(t *testing.T) {
```

internal/cluster/distribution/eks/node_pool_test.go
```diff
@@ -0,0 +1,116 @@
...
+	t.Run("PoolZeroSize", func(t *testing.T) {
+		pool := base
+		pool.Size = 0
+		pool.Autoscaling.Enabled = false
+
+		err := pool.Validate()
+		assert.IsType(t, cluster.ValidationError{}, errors.Cause(err))
+	})
+
+	t.Run("PoolSizeNegative", func(t *testing.T) {
```
pregnor, line 42: **Recommended**: renaming this case **SHOULD** be considered for clarity purposes, the size is positive here, only the minimum size is negative.
```suggestion
	t.Run("PoolMinSizeNegative", func(t *testing.T) {
```

internal/cluster/distribution/eks/node_pool.go
sagikazarmark: AFAIK the min and max size cannot be the same, the max size needs to be at least larger than the min by one.

Note that we don't actually use the autoscaling capabilities of ASGs, so this is purely an upper limit for the cluster autoscaler. Fixed size node pool in our case means an ASG with a fix desired size, disabled autoscaling policies, but the ASG still requires minimum and maximum sizes.

pregnor: I see, thanks for the clarification.

orymate: if autoscaling is not enabled, we don't use the min/max values from this struct, the asg is created with min=size and max=size+1 instead. so if you don't want autoscaling, you should set enabled=false and you can leave the min/max sizes zero. i don't know about any technical constraint from aws because of which we'd have to ensure that max is more than min, but in case of enabled autoscaling it makes sense to require this here

internal/cluster/distribution/eks/node_pool.go
orymate: (why?)

PR #3064, opened 2020-07-28
Automatically add go.mod to plz-out

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Regular go tools look for go files in all directories.
This causes issues in case of please, because it actually stores
Go files in its plz-out directory.

The solution is to make it a dummy module,
because Go's package filtering (eg. ./...) ignores submodules.

### DIFF
--- a/BUILD.plz
+++ b/BUILD.plz
@@ -5,6 +5,22 @@ github_repo(
 )
 
 genrule(
+    name = "go.mod",
+    outs = ["go.mod"],
+    cmd = [
+        """cat > $OUTS << EOF
+module plz-out
+
+go 1.14
+EOF
+      """
+    ],
+    labels = [
+        "link:plz-out/"
+    ]
+)
+
+genrule(
     name = "docker-compose.override.yml",
     srcs = ["docker-compose.override.yml.dist"],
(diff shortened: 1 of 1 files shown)


BUILD.plz
```diff
@@ -4,6 +4,22 @@ github_repo(
...
 )
 
+genrule(
+    name = "go.mod",
+    outs = ["go.mod"],
+    cmd = [
+        """cat > $OUTS << EOF
+module plz-out
+
+go 1.14
```
pregnor, line 14: **Question**: if this is just a dummy module, can we consider deriving a version here from a source instead of hard-coding one to avoid maintaining this version? 
Such maintenance is usually unnecessary for a dummy module, but can still cause confusion and edge case issues when a version go out of support or a non-backward compatible version is promoted to default.

sagikazarmark: Well, it doesn't matter since this go.mod file won't ever be used, but for the sake of consistency: done.

PR #3076, opened 2020-07-30
fix(EKS): removed node pool create failure residue

### LINKED ISSUE #3031: Failed node pool creates leave behind non-deleteable resources
**Describe the bug**
When a node pool create fails, the cloudformation stack stucks in an error state and it cannot be deleted with node pool delete. As a result, the same node pool name cannot be reused.

1. Create a new node pool with an AMI from a different region
1. Let it fail.
1. Try to delete the node pool

**Expected behavior**
Deleting a failed node pool should succeed.

**Additional context**
We have no information about how deleting a failed node pool fails.

In the future, we should probably clean up the failed resources. We should decide whether we wait until the process log becomes GA or go ahead and implement the clean up logic. The important thing is to get the error.

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3031

### What's in this PR?

1. Unwrapped Cadence errors at cluster node pool creation/deletion.
2. Removed all residue resources when an AWS/EKS node pool creation fails, including deleting failed-to-create CloudFormation stacks.

### Why?

There were a couple minor issues such as 

- error details not being displayed and 

and one major issue namely 

- when the CF node pool / ASG stack creation failed, a non-describable (but list-able) failed stack was left behind which couldn't be deleted on the pipeline API due to shortcutting the request response due to no node pool entry in the database and also couldn't be recreated because the stack was already existing.

Solution to the minor problems were trivial.
3 different solutions were considered for the major issue: (in the order of consideration)
1. **Recording the failed-to-create node pools in the database as "known" node pool in order to execute deletion on delete request.**
This proved to be unfeasible as the database modification is part of the activity, but the workflow retry logic would try to rerun the activity which would cause a clash when trying to write the node pool to the DB for the second time. 
- I considered checking whether the DB and activity objects are identical and skipping the DB write in that case, but it would have been tedious and difficult to maintain because of two similarly structured but different types which are used directly, not through an interface.
- Also considered using the workflow ID - which remains the same throughout same workflow's activity retries - to identify retries, but didn't find any good ways to persist it throughout activity retries - for example as part of a pool description in the database it felt forced/out-of-place workaround and not a real fix. **After a discussion there is supposed to be an available activity attempt indicator, so I'm trying to pursue this solution.**
- Also considered moving the database mutation out of the node pool creation activity in order to prevent workflow retry issues while writing failed-to-create node pools to the DB on workflow level, but I didn't like that approach from a high level perspective, the DB write should be part of the current create node pool activity and refactoring that single activity into smaller steps felt like a bigger refactor which seemed out of scope for the current bug.

2. **Checking failed-to-create stack existence from AWS instead of the database and executing a delete on such stacks.**
This option would have been feasible, but it would have meant coupling node pool creation and deletion steps together loosely by requiring a list+delete action for all resources mutated during creation which I consider harder to maintain compared to option 3. Also this basically would have relegated failure cleanup to a manual request.

3. **Making the node pool create activity atomic and restoring the pre-node-pool-creation state of the cluster in case of a failure including the CloudFormation stacks.**
This proved to be the easiest to implement solution right now because only the node pool / ASG stack creation step mutates any resources which meant a single deletion step in the error handling branch resolved the issue and eliminated all failure residue. From what I understood (CMIIW) this is also the long-term goal we are working towards - making infrastructure resource mutation atomic seems a good approach to me. This solution also requires the least amount of manual intervention in case of a node pool creation issue while preserves and conveys all failure information as a cluster status warning.

See the referenced issue for more details.

### Additional context

All 2 solutions, but especially the 1 minor ones (error unwrapping) **SHOULD** be considered to be utilized for all other use-cases in the code-base where similar issues may arise.
I decided not to include these changes into this PR because they are not part of the strict scope, but I'm planning to implement those changes after finishing this PR.

The following PRs implemented those minor fixes/improvements:
Error detail propagation: https://github.com/banzaicloud/pipeline/pull/3106
Unique AWS client request token: https://github.com/banzaicloud/pipeline/pull/3107

### Checklist

- [X] Implementation tested (with at least one cloud provider) [AWS]

### To Do

- ~Write unit tests for the node pool creation/deletion workflows.~ Note: out of scope for the current PR due to time constraints and some obstacles to be solved for activity unit testing.
- [X] Annotate the stack delete errors returned with the intention (like wrap with "deleting failed stack")
- [X] Reimplement this way: https://github.com/banzaicloud/pipeline/issues/3031#issuecomment-669140908

### DIFF
--- a/internal/cluster/clusterworkflow/activity_create_node_pool.go
+++ b/internal/cluster/clusterworkflow/activity_create_node_pool.go
@@ -67,6 +67,8 @@ type CreateNodePoolActivityInput struct {
 }
 
 func (a CreateNodePoolActivity) Execute(ctx context.Context, input CreateNodePoolActivityInput) error {
+	activityInformation := activity.GetInfo(ctx)
+
 	c, err := a.clusters.GetCluster(ctx, input.ClusterID)
 	if err != nil {
 		return cadence.WrapClientError(err)
@@ -137,6 +139,13 @@ func (a CreateNodePoolActivity) Execute(ctx context.Context, input CreateNodePoo
 			AWSClientRequestTokenBase: pkgAmazon.NewNormalizedClientRequestToken(activity.GetInfo(ctx).WorkflowExecution.ID),
 		}
 
+		if activityInformation.Attempt == 0 {
+			err = a.eksNodePools.CreateNodePool(ctx, eksCluster.ID, input.UserID, nodePool)
+			if err != nil {
+				return err
+			}
+		}
+
 		var vpcActivityOutput *eksworkflow.GetVpcConfigActivityOutput
 		{
 			input := eksworkflow.GetVpcConfigActivityInput{
@@ -195,12 +204,6 @@ func (a CreateNodePoolActivity) Execute(ctx context.Context, input CreateNodePoo
 		if err != nil {
 			return cadence.WrapClientError(err)
 		}
-
-		err = a.eksNodePools.CreateNodePool(ctx, eksCluster.ID, input.UserID, nodePool)
-		if err != nil {
-			return err
-		}
-
 	default:
 		return cadence.WrapClientError(errors.WithStack(cluster.NotSupportedDistributionError{
 			ID:           c.ID,
--- a/internal/cluster/clusterworkflow/workflow_create_node_pool.go
+++ b/internal/cluster/clusterworkflow/workflow_create_node_pool.go
@@ -21,7 +21,7 @@ import (
 	"go.uber.org/cadence/workflow"
 
 	"github.com/banzaicloud/pipeline/internal/cluster"
-	_cadence "github.com/banzaicloud/pipeline/pkg/cadence"
+	pkgCadence "github.com/banzaicloud/pipeline/pkg/cadence"
 )
 
 const CreateNodePoolWorkflowName = "create-node-pool"
@@ -41,7 +41,7 @@ func CreateNodePoolWorkflow(ctx workflow.Context, input CreateNodePoolWorkflowIn
 			InitialInterval:          15 * time.Second,
 			BackoffCoefficient:       1.0,
 			MaximumAttempts:          30,
-			NonRetriableErrorReasons: []string{_cadence.ClientErrorReason, "cadenceInternal:Panic"},
+			NonRetriableErrorReasons: []string{pkgCadence.ClientErrorReason, "cadenceInternal:Panic"},
 		},
 	}
 	_ctx := ctx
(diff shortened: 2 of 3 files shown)


internal/cluster/clusterworkflow/activity_create_node_pool.go
```diff
@@ -191,6 +195,15 @@ func (a CreateNodePoolActivity) Execute(ctx context.Context, input CreateNodePoo
 
 		_, err = eksworkflow.NewCreateAsgActivity(a.awsSessionFactory, nodePoolTemplate).Execute(ctx, subinput)
 		if err != nil {
+			input := eksworkflow.DeleteStackActivityInput{
```
orymate, line 198: Did you consider handling this branch in the workflow, and deleting the stack in a different activity? It may be more robust, for example because of the retries and the delays between the creation and the deletion requests (and also the result of the workflow is easier to understand in case of failures). Otherwise, please annotate the stack delete errors returned with the intention (like wrap with "deleting failed stack")

pregnor: I was thinking about it. I could argue for and against both, whichever you would prefer is fine for me.

I guess I dropped it in favor of my original solution because on the level of the workflow if the `CreateNodePoolActivity` activity fails then we **MUST** execute the inverse operation of each mutating subactivity potentially executed in the mentioned main activity - or identify the error and branch from there to undo the previous actually executed mutating subactivities.
Which would loosely couple the activity and workflow code/logic which is IMO a fragile approach from maintenance perspective because of the distance inbetween/scope encapsulating them.
It was a higher precedence for me to avoid this - details on this can be requested.

Also thought about either modifying the `DeleteNodePoolActivity` or defining an `UndoCreateNodePoolActivity` activity which is basically a delete activity with try/no-fast-fail subactivities (so errors doesn't immediately terminate the function) but it would have required more time to think about so I wanted to start the discussion by opening the PR with my original approach.
It has its ups and downs, I would somewhat like that approach if we can pay attention to it on a repository scale and handle activities accordingly from a point in time - which could not be fulfilled on my last project unless I personally made sure of it so I'm a bit more careful this time.
It also means that the create and delete[+undo] activities are needed to be kept "in sync" as each other's inverse operation, similarly the workflow+activity combination discussed earlier.

I guess I wanted to narrow the scope down as much as possible.

On the other hand my approach can also end up badly if another mutating subactivity is introduced and the error handling logic is not updated - but I guess a function's scope is still better than a package's/workflow-activity pair's.
I'm going to think about better ways to do this[1], I might have hastily decided on the first approach, but as a discussion driver I achieved my goal.
[1]: one quick idea is to define a local cleanup function which can work on such a try/no-fast-fail basis, and wire the error handling through that, it seems to have the ups and minimize the downs, but this is just a McDonalds idea popping into my head while I were typing the response.

pregnor: I will definitely incorporate the error changes you mentioned, adding it to the todo list

pregnor:
Note: https://github.com/banzaicloud/pipeline/issues/3031#issuecomment-669140908

PR #3079, opened 2020-07-30
Refactor AWS auth config

### LINKED ISSUE #3058: Refactor AWS auth config support
Use ConfigMap structure instead of a string field

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | yes
| Deprecations?   | no
| Related tickets | fixes #3058 

### What's in this PR?
Use ConfigMap structure instead of a string field in case of `authConfigMap`

### Why?
To make API more user friendly

### Checklist


.gen/pipeline/api/openapi.yaml
```diff
@@ -18840,6 +18840,33 @@ components:
       - maxSize
       - minSize
       type: object
+    AWSAuthConfigMap:
+      properties:
+        data:
+          $ref: '#/components/schemas/AWSAuthConfigMap_data'
+      type: object
+    AWSAuthConfigMapUsers:
```
bonifaido, line 18848: I think these `mapRoles` and `mapUsers` (and`mapAccounts`, which is missing actually) should be mere strings. They are defined as strings in the original YAML format (Yaml strings).

pregnor: While they are indeed defined as multi-line strings on [AWS side](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html), it is not easy to define these on the REST API interface in the JSON, because the new lines are needed to be escaped as `\n`, which makes it much harder to specify it properly - with correct YAML-like multi-line string content which I believe is somewhat used manually as well (besides being used programatically).

I leave the trade-off decision for you (all), I just wanted to make sure it is a conscious decision.

sagikazarmark: FYI: this is the generated descriptor, not the original one.

apis/pipeline/pipeline.yaml
```diff
@@ -4494,6 +4494,45 @@ components:
                     type: integer
                     example: 5
 
+        AWSAuthConfigMap:
+            type: object
+            properties:
+                data:
```
sagikazarmark, line 4500: I don't think we need the `data` level here.

.gen/pipeline/api/openapi.yaml
bonifaido: Ah okay, so we offer the API in a structured way, in that way it makes sense to leave it like that.

Then the only thing left, please add mapAccounts as well.

apis/pipeline/pipeline.yaml
```diff
@@ -4494,6 +4494,45 @@ components:
                     type: integer
                     example: 5
 
+        AWSAuthConfigMap:
```
sagikazarmark, line 4497: How about one of the following (or similar):

- `AWSAuth`
- `AWSAuthConfig`
- `AWSClusterAuth`
- `AWSClusterAuthConfig`

apis/pipeline/pipeline.yaml
```diff
@@ -4665,7 +4704,7 @@ components:
                         - nodePools
                     properties:
                         authConfigMap:
```
sagikazarmark, line 4706: This could also be something like `authConfig` or `clusterAuth` or `clusterAuthConfig`

.gen/pipeline/api/openapi.yaml
colin014: thx guys, I added `mapAccounts` field to the structure

apis/pipeline/pipeline.yaml
colin014: removed `data` level in cbd6f9bd0

apis/pipeline/pipeline.yaml
colin014: I changed it to `AWSAuthConfig`

apis/pipeline/pipeline.yaml
colin014: I also changed this `authConfig`

apis/pipeline/pipeline.yaml
```diff
@@ -4494,6 +4494,46 @@ components:
...
+                mapUsers:
+                    type: array
+                    items:
+                        $ref: '#/components/schemas/AWSAuthConfigMapUsers'
+                mapAccounts:
+                    type: array
+                    items:
+                        type: string
+
+        AWSAuthConfigMapUsers:
```
sagikazarmark, line 4513: Please also rename this one (without `Map`)

apis/pipeline/pipeline.yaml
```diff
@@ -4494,6 +4494,46 @@ components:
...
+                group:
+                    type: array
+                    items:
+                        type: string
+                username:
+                    type: string
+                userarn:
+                    type: string
+
+        AWSAuthConfigMapRoles:
```
sagikazarmark, line 4525: Please also rename this one (without `Map`)

apis/pipeline/pipeline.yaml
```diff
@@ -4494,6 +4494,46 @@ components:
                     type: integer
                     example: 5
 
+        AWSAuthConfig:
```
sagikazarmark, line 4497: Further thinking about it: maybe this should be called `EKSAuthConfig` since it's actually EKS specific. WDYT @colin014 @bonifaido @pregnor

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -42,9 +44,68 @@ type CreateClusterEKS struct {
...
+data:
+ mapRoles: |
+  %s
+ mapUsers: |
+  %s
+ mapAccounts: |
+  %s`, mrIndent, muIndent, maIndent), nil
+}
+
+type AuthConfigMapData struct {
```
sagikazarmark, line 90: Please also rename this one (without `Map`). Maybe just call it `AuthConfig`?

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -42,9 +44,68 @@ type CreateClusterEKS struct {
...
+	MapRoles    []MapRoles `json:"mapRoles,omitempty" yaml:"mapRoles,omitempty"`
+	MapUsers    []MapUsers `json:"mapUsers,omitempty" yaml:"mapUsers,omitempty"`
+	MapAccounts []string   `json:"mapAccounts,omitempty" yaml:"mapAccounts,omitempty"`
+}
+
+type BaseMapFields struct {
```
sagikazarmark, line 96:
```suggestion
type AuthConfigBaseFields struct {
```

??

src/cluster/eks.go
```diff
@@ -61,6 +61,10 @@ func CreateEKSClusterFromRequest(request *pkgCluster.CreateClusterRequest, orgId
 
 	modelNodePools := createNodePoolsFromRequest(request.Properties.CreateClusterEKS.NodePools, userId)
 
+	configMapString, err := request.Properties.CreateClusterEKS.AuthConfig.ConvertToString()
```
sagikazarmark, line 64:
```suggestion
	authConfigMap, err := request.Properties.CreateClusterEKS.AuthConfig.ConvertToString()
```

apis/pipeline/pipeline.yaml
pregnor: I agree it is used in AWS/EKS context only and not other AWS contexts such as PKE, so it may slightly improve clarity to call it EKS.

PR #3082, opened 2020-08-02
Use go_library compat rules

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Use official go_* compatible rules.

### Why?
As we move forward with please, it would be nice to remain compatible with the official tools.


sagikazarmark:
Unfortunately, no. The package concept that's supposed to speed up the builds requires a BUILD file in every package. I don't think it's that much of a problem though, since these files can be generated (actually, they are).

sagikazarmark:
Sadly, this made things worse. We will have to find another way to move forward with Please. I'll open a separate PR for the rule compat part though.

Issue #3024, opened 2020-07-20
Improve container runtime options


sagikazarmark:
**Is your feature request related to a problem? Please describe.**
Currently, GPU based workloads require Docker container runtime (and the relevant drivers installed, for which we provide custom images). Unfortunately, this forces me to use Docker for my entire cluster, whereas I could use Docker for the GPU workers, and use containerd for everything else. (Whether it's a good idea or not is an entirely different question, but it seems to be working)

**Describe the solution you'd like to see**
I'd like to have better control over the used container runtime.

One option is customizing a (cluster-)global default on a node pool basis.

Another option is having an "auto" value that chooses the best available option based on image availability (eg. docker for GPU nodes, containerd for everything else)

**Additional context**
This is closely related to the image selection.

PR #3092, opened 2020-08-04
Add GPU image selector for PKE

### LINKED ISSUE #3050: Add GPU support to the image selection logic for PKE AWS
When I create a cluster from the UI using GPU instance types, I cannot utilizes the GPUs because the selected machine image does not have support for it.

If I choose an instance type with GPU, try to find an image that has GPU support, fallback to "regular" images otherwise.

**Additional context**
Integrate the new image selection logic into the interface in #3025

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3050

### What's in this PR?
Select GPU images for GPU instance types

### Why?
To support GPU workloads better

### DIFF
--- a/internal/cluster/distribution/pke/pkeaws/image_selector_gpu.go
+++ b/internal/cluster/distribution/pke/pkeaws/image_selector_gpu.go
@@ -1,0 +1,44 @@
+// Copyright © 2020 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pkeaws
+
+import (
+	"context"
+	"strings"
+
+	"emperror.dev/errors"
+)
+
+type gpuImageSelector struct {
+	imageSelector ImageSelector
+}
+
+// NewGPUImageSelector returns a new GPU image selector.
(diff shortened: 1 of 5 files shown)


internal/cluster/distribution/pke/pkeaws/image_selector_gpu.go
```diff
@@ -0,0 +1,44 @@
...
+		return "", errors.WithStack(ImageNotFoundError)
+	}
+
+	return s.imageSelector.SelectImage(ctx, criteria)
+}
+
+func isGPUInstance(instanceType string) bool {
+	return strings.HasPrefix(instanceType, "p2.") || strings.HasPrefix(instanceType, "p3.") ||
+		strings.HasPrefix(instanceType, "g3.") || strings.HasPrefix(instanceType, "g4.")
+}
```
pregnor, line 44: **Question**: wouldn't it be a good idea to export this function and then use it at [`create_aws_worker_pool_activity.go:100`](https://github.com/banzaicloud/pipeline/pull/3092/files#diff-bdb953b7e804358ba84ce7b188c30d74R100) as this package is already imported there?

Similar situation at [`ec2_pke.go:894`](https://github.com/banzaicloud/pipeline/pull/3092/files#diff-29cd556cd4ed92cc68414cf495df9b58R894), even though there this package is not imported.
(In case we want to avoid importing this package there, I can start thinking of finding a better place for the exported `IsGPUInstance()` function. In general I would think of some interface above the concrete providers' level and individual implementations at and below that point.)

Because I think maintaining multiple copies of the same code implies a mental burden and I'm afraid of human errors in case we have to modify the logic - which I admit is unlikely in the near future, but is still a possibility and if reusing code here doesn't have a serious cost which I don't see, I would argue why not.

sagikazarmark: I was trying to implement this code the worst possible way, so we will have an incentive in fixing it. 😄 

This is a temporary solution which will be replaced once #3024 is done.

PR #3093, opened 2020-08-04
Plz wollemi

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets |

### What's in this PR?
Plz changes
Introduction of wollemi

### Why?
Somewhere along the way the current please setup became slower. This is an attempt to make it faster again.


apis/pipeline/BUILD
```diff
@@ -2,14 +2,14 @@ subinclude("///pleasings2//openapi")
...
-    spec = "pipeline.yaml",
-    generator = "go-server",
     out = ".gen/pipeline",
     additional_properties = {
         "packageName": "pipeline",
         "withGoCodegenComment": "true",
         "sourceFolder": "pipeline",
     },
+    generator = "go-server",
+    spec = "pipeline.yaml",
```
pregnor, line 12: Is this generated?
If not, why is the generator put here and not above the `name` field?

sagikazarmark: It's auto formatted by wollemi using the [buildifier](https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md) tool

pregnor: I'm a bit sad personally they didn't go for the lexicographical order, so close yet so far.

PR #3107, opened 2020-08-07
fix(AWS): added unique client request token

### LINKED ISSUE #3016: Creating a node pool with a previously used name fails
**Describe the bug**
When creating a node pool in a cluster (using the new node pool API/CLI tool) with a name that was previously used for another (but deleted) node pool, the CF create request fails.

1. Create a node pool with the name `poolX`
2. Delete the node pool
3. Create a node pool again with the name `poolX`

**Expected behavior**
The node pool should be created successfully.

**Additional context**

CF error:

```
 could not create 'pipeline-eks-nodepool-poolX' CF stack: TokenAlreadyExistsException: The client request token 481bee4e-fd1a-49bd-a2dc-7b85cb1cd163-eks-create-asg has already been used for this stack. Retry the request with a different client request token status code: 400
```

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3016 

### What's in this PR?

Replaced cluster UUID, generated UUID and empty AWS client request tokens with Cadence workflow ID.

### Why?

To provide a client request token that is unique among different "requests", but are the same between activity retries so AWS can handle same-stack requests in an idempotent fashion. ActivityID is not used, because it would be different for the same same-stack operation between activity retries.

### Checklist

### DIFF
--- a/internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
+++ b/internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
@@ -29,6 +29,7 @@ import (
 	cloudformation2 "github.com/banzaicloud/pipeline/internal/cloudformation"
 	"github.com/banzaicloud/pipeline/internal/cluster/distribution/pke/pkeaws"
 	"github.com/banzaicloud/pipeline/internal/providers/amazon"
+	pkgAmazon "github.com/banzaicloud/pipeline/pkg/providers/amazon"
 	pkgCloudformation "github.com/banzaicloud/pipeline/pkg/providers/amazon/cloudformation"
 )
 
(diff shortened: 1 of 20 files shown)


internal/providers/pke/pkeworkflow/create_aws_worker_pool_activity.go
```diff
@@ -122,9 +122,9 @@ func (a *CreateWorkerPoolActivity) Execute(ctx context.Context, input CreateWork
 	}
 
 	stackInput := &cloudformation.CreateStackInput{
-		StackName:    aws.String(stackName),
-		TemplateBody: aws.String(template),
-		// ClientRequestToken: aws.String(string(activity.GetInfo(ctx).ActivityID)),
+		StackName:          aws.String(stackName),
+		TemplateBody:       aws.String(template),
+		ClientRequestToken: aws.String(activity.GetInfo(ctx).WorkflowExecution.ID),
```
pregnor, line 127: Note: I explicitly highlight this change, because this isn't just a client token request (CRT) value replacement, but also an introduction of using CRT in this context.
I assumed the original value was commented out because activity ID was not a reliable CRT source as it changed between activity retries.

internal/providers/pke/pkeworkflow/delete_aws_pool_activity.go
```diff
@@ -71,8 +72,8 @@ func (a *DeletePoolActivity) Execute(ctx context.Context, input DeletePoolActivi
 	}
 
 	stackInput := &cloudformation.DeleteStackInput{
-		StackName: aws.String(stackName),
-		// ClientRequestToken: aws.String(string(activity.GetInfo(ctx).ActivityID)),
+		StackName:          aws.String(stackName),
+		ClientRequestToken: aws.String(activity.GetInfo(ctx).WorkflowExecution.ID),
```
pregnor, line 76: Note: I explicitly highlight this change, because this isn't just a client token request (CRT) value replacement, but also an introduction of using CRT in this context.
I assumed the original value was commented out because activity ID was not a reliable CRT source as it changed between activity retries.

PR #3114, opened 2020-08-12
WIP: explore alternative ways to diff DB schemas

### LINKED ISSUE #2271: Use official golang docker image in CI
Currently we use mysqldiff to check the consistency between migration scripts and gorm automigration.

Unfortunately, Ubuntu Bionic and Debian Stretch were the last distributions with a package containing mysqldiff.

Since Go 1.13 images use newer distributions, we temporarily had to switch to a custom image based on bionic.

For the long term, we need to find another way to run those tests and get rid of the custom image.

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #2271 

### What's in this PR?

TBA

### Why?

TBA

### Additional context

TBA

### Checklist

### To Do

- [ ] Column order differences break diff.
- [X] Comment differences break diff.
- [X] Constraint name differences break diff.
- [ ] Constraint order differences break diff.
- [ ] Key order differences break diff.
- [X] Table option differences break diff.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,11 @@ jobs:
             -
                 run:
                     name: Test MySQL database schema diff
-                    command: mysqldiff --skip-table-options --server1=root:@127.0.0.1:3306 --server2=root:@127.0.0.1:3306 pipeline_migrations:pipeline_automigrate
+                    command: |
+                        go run ./cmd/mysqldiff/main.go \
+                            --databases pipeline_migrations,pipeline_automigrate \
+                            --host 127.0.0.1 --password "" --port 3306 --user root \
+                            --options ignore-comments,ignore-table-options
 
             -
                 run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -133,7 +133,37 @@ jobs:
...
+                            --host 127.0.0.1 \
+                            --password= \
+                            --port 3306 \
+                            --user "root" \
+                            \
+                            --no-data \
+                            --result-file "pipeline_automigrate.sql" \
+                            --verbose \
+                            pipeline_automigrate
+                        diff \
```
sagikazarmark, line 157: This won't skip table options

pregnor: I'm in the experiment phase and there is a flag for that in mysqldump for exactly that, I just didn't want to start with enabling skips we might not need.

pregnor: Fixed in https://github.com/banzaicloud/pipeline/compare/eadf636cf60cfb95524e7d6500832b54c0743096..cff5a2ae422cd4afb400a4d4feffe80a7dfe9f6e.

pregnor:
Won't fix for now.

PR #3117, opened 2020-08-14
feat(EKS): added envelope encryption

### LINKED ISSUE #3088: Add envelope encryption support to EKS
As a user concerned about security, I want etcd to be encrypted.

On EKS, envelope encryption provides that feature, but it can only be enabled/disabled when the cluster is launched.

Make this feature available when launching an EKS cluster.

In case of managed distributions, there aren't really any alternatives.

**Additional context**
https://aws.amazon.com/about-aws/whats-new/2020/03/amazon-eks-adds-envelope-encryption-for-secrets-with-aws-kms/

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | closes #3088 

### What's in this PR?

Added `encryptionConfigurations` property to EKS cluster creation in order to propagate it to the EKS cluster creation request for implementing envelope encryption.

### Why?

Our users would like to use envelope encryption to protect their resources which requires an encryption key to be specified on cluster creation and needs to be propagated to the EKS cluster creation request.

### Additional context

Documentation PR: https://github.com/banzaicloud/banzaicloud.github.io/pull/1731

The underlying EKS API encryption configuration is represented the following way on the SDK:
```json
{
    "encryptionConfig": [ // The encryption configuration for the cluster.
        { 
            "provider": {  // AWS Key Management Service (AWS KMS) customer master key (CMK). Either the ARN or the alias can be used. 
                "keyArn": "arn:aws:kms:<REGION>:000000000000:key/00000000-0000-0000-0000-000000000000"
            },
            "resources": [ // Specifies the resources to be encrypted. The only supported value is "secrets".
                "secrets"
            ]
        }
    ]
}
```

Meaning the `secrets` resources are going to be encrypted with the customer master key `arn:aws:kms:<REGION>:000000000000:key/00000000-0000-0000-0000-000000000000`.

Alternatives considered:
1. Original EKS `encryptionConfig` definition: I found it tedious (meaning unnecessary hierarchy levels) and unintuitive while also wanted to avoid exposing underlying SDK types.
2. Single `encryptionConfigurationKey` string value which would transfer the CMK and automatically apply to all resources: I found it not robust enough while predicting support for more resource kinds in the future which would mean a possible requirement of handling multiple encryption keys for different resource kinds and cause difficulties to support without API breakage.
3. Map of CMKs and corresponding resource list to encrypt: difficulty to support alternative provider structure implementations.

### Checklist

- [X] User guide and development docs updated (if needed): https://github.com/banzaicloud/banzaicloud.github.io/pull/1731

### To Do

- ~Restrict `EKSEncryptionConfiguration` resource values to available values.~ (Decided not to implement, because it would make the data representation more fragile in case of possible foreseeable changes such as other resource types being introduced.)
- [X] Check whether encryption configuration validation can be done with the criteria listed in the issue, preferably close to the cluster creation code.


pregnor:
Tested on EKS for happy path with
```json
```

pregnor:
Tested on EKS for sad path (cluster and encryption key region mismatch) with:
```json
{
  "name": "pregnor-local-200815-pipeline-3088-4",
  "location": "eu-west-1",
  "cloud": "amazon",
  "secretName": "aws-pregnor",
  "properties": {
    "eks": {
      "version": "1.17.9",
      "encryptionConfig": [
        {
          "provider": {
            "keyArn": "arn:aws:kms:eu-central-1:161831738826:key/02a2ba49-42ce-487f-b006-34c64f4b760e"
          },
          "resources": [
            "secrets"
          ]
        }
      ],
      "nodePools": {
        "pool1": {
          "spotPrice": "0.01",
          "count": 1,
          "minCount": 1,
          "maxCount": 2,
          "autoscaling": true,
          "instanceType": "t2.small"
        }
      }
    }
  }
}
```
resulting in error and cluster status message `failed to create EKS master: InvalidParameterException: The keyArn for encryptionConfig is scoped to invalid aws region. { RespMetadata: { StatusCode: 400, RequestID: "6a4663c1-2999-45bf-ad50-54bc36abbb7e" }, ClusterName: "pregnor-local-200815-pipeline-3088-4", Message_: "The keyArn for encryptionConfig is scoped to invalid aws region." } `.

sagikazarmark:
Validating the request early on is probably better as the EKS cluster is created async.

pregnor:
@sagikazarmark 
> Validating the request early on is probably better as the EKS cluster is created async.

Fixed in https://github.com/banzaicloud/pipeline/compare/22cea0c7f9c2143807db8a7bded25cac31626817..dd8e4302c71baf2bd60101d646dbe0a5ac3e17fd.

database/migrations/mysql/1597442135_add_eks_encryption_config_column.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `amazon_eks_clusters` ADD COLUMN `encryption_config` JSON;
```
sagikazarmark, line 1: Question: Do we have to save these details in the database? Is this something we want to return in details?

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -38,9 +39,10 @@ type CreateClusterEKS struct {
 	// Subnets for EKS master and worker nodes. All worker nodes will be launched in the same subnet
 	// (the first subnet in the list - which may not coincide with first subnet in the cluster create request payload as
 	// the deserialization may change the order) unless a subnet is specified for the workers that belong to a node pool at node pool level.
-	Subnets  []*Subnet  `json:"subnets,omitempty" yaml:"subnets,omitempty"`
-	IAM      ClusterIAM `json:"iam,omitempty" yaml:"iam,omitempty"`
-	LogTypes []string   `json:"logTypes,omitempty" yaml:"logTypes,omitempty"`
+	Subnets          []*Subnet               `json:"subnets,omitempty" yaml:"subnets,omitempty"`
+	IAM              ClusterIAM              `json:"iam,omitempty" yaml:"iam,omitempty"`
+	EncryptionConfig []*eks.EncryptionConfig `json:"encryptionConfig,omitempty" yaml:"encryptionConfig,omitempty"`
```
sagikazarmark, line 44: Does it have to be a pointer? (TBH I have no idea why Subnet is a pointer).

internal/cluster/distribution/eks/eksprovider/driver/cluster_creator.go
```diff
@@ -315,6 +320,16 @@ func (c *EksClusterCreator) validate(r *pkgCluster.CreateClusterRequest, logger
 		return errors.WrapIf(err, "failed to create AWS session")
 	}
 
+	// Validate encryption configuration.
+	if modelCluster.EncryptionConfig != nil {
```
sagikazarmark, line 324: `len > 0` sounds more appropriate, since this could be a non-nil, 0 length slice, couldn't it?

sagikazarmark:
Looks like we don't just formally validate the input, but also check if the key is available. Is that correct?

When I meant validation, I rather meant formal validation, but I'm not saying checking if the key exists is a bad idea, but we have to be aware of the consequences:

- It will make the create cluster request "slower"
- We need to document any additional permission requirements

I'm okay with merging it as is, but we need a strategy for these things in the future. IMHO formal data validation during the request should be enough, any additional validation (eg. key exists) should be done asynchronously (the key could be deleted after the request is submitted for instance).

@pregnor can you formulate an ADR for this validation problem? (I can provide more details for that offline)

database/migrations/mysql/1597442135_add_eks_encryption_config_column.up.sql
pregnor: I guess it's not really a use-case, no.
I took the auth config as an example that's why I included it in the database.
I'm gonna try and move the encryption config out of the EKS cluster model then into the EKS cluster directly.

internal/cluster/distribution/eks/ekscluster/eks.go
pregnor: The AWS API expects it to be `[]*eks.EncryptionConfig` (slice of pointers) and I wanted to avoid any unnecessary data transformation logic, so this seemed more straightforward, I don't really see any benefits of changing it to non-pointer and doing the additional data transformation aside from aesthetic gain.

internal/cluster/distribution/eks/eksprovider/driver/cluster_creator.go
pregnor: Well those are different cases.

Here we decide whether we want to validate the encryption configuration at all.
Specifying nil encryption configuration is a valid use case when the user opts out of envelope encryption, in such case s we run no validation, this conditional branching checks that case.

The check for the list not being empty is implemented in the validation function's logic as in case of any encryption config being specified we do not want to accept empty configurations.

What do you think, should we approach this `non-existent OK, otherwise validate` situation differently?

pregnor:
Sure I will assemble an ADR proposal.

I agree formal validation should be enough here, so I'm inclined to remove the last check validating the symmetric key, because it complicates the validation and as I stated in the PR description, I don't see much of a gain from it. It's very easy to get rid of.

What made me do deeper validation was the way we validated the VPC in the `pkg/providers/amazon/ec2` package, where in the `network.go`'s `VpcAvailable()` function we do similar deep validation in the same `EksClusterCreator.validate()` function when a not empty VPC ID is specified.

pregnor:
I made the validation formal-only in https://github.com/banzaicloud/pipeline/compare/dc41f07e02f491c45ccb33b88ba754068165842c..13fcb1cea8fb9893dd98beb41f26c115c9467f94, also moved the remaining validation logic to `pkg/providers/amazon/eks.ValidateEncryptionConfiguration()`, also moved the "nil encryption config is OK to create unencrypted cluster" logic into the validation function, added an `empty encryption config is OK to create unencrypted cluster` case and changed the configuration item count checking to `> 1` from `!= 1`.

internal/cluster/distribution/eks/eksprovider/driver/cluster_creator.go
pregnor: See my [update](https://github.com/banzaicloud/pipeline/pull/3117#issuecomment-674568253), I convinced myself it is better as I described in the last part of my last comment.

database/migrations/mysql/1597442135_add_eks_encryption_config_column.up.sql
pregnor: I moved the encryption config out of the cluster model [here](https://github.com/banzaicloud/pipeline/compare/d9ba632ed04d449bcfc4601cee8c5caeeea20c72..0df34bec89a6668d01dc76aac60a168d02b4ce89).

And removed the database migration files [here](https://github.com/banzaicloud/pipeline/compare/0df34bec89a6668d01dc76aac60a168d02b4ce89..194ac0de0ef9f8145ee2995cb56c716007c389c2).

pkg/providers/amazon/eks/encryption_configuration.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"emperror.dev/errors"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+)
+
+// ValidateEncryptionConfiguration returns an error in case the specified
+// encryption configuration is invalid.
+func ValidateEncryptionConfiguration(encryptionConfig []*eks.EncryptionConfig, clusterLocation string) (err error) {
+	if encryptionConfig == nil ||
```
sagikazarmark, line 28: What I was trying to say: `nil != []*eks.EncryptionConfig{}`, so the proper way of checking a slice for empty value is len.

You don't need the extra nil check here though: len takes care of it and will return 0 (if it is a typed nil that has a length): https://play.golang.org/p/8GqjE204U0P

pregnor:
We discussed IRL we would like to release this feature this week, preferably on Tuesday, so I'm planning to merge this EOB (16:00) today CEST unless changes are requested.

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -31,10 +32,11 @@ import (
 
 // CreateClusterEKS describes Pipeline's Amazon EKS fields of a CreateCluster request
 type CreateClusterEKS struct {
-	Version      string               `json:"version,omitempty" yaml:"version,omitempty"`
-	NodePools    map[string]*NodePool `json:"nodePools,omitempty" yaml:"nodePools,omitempty"`
-	Vpc          *ClusterVPC          `json:"vpc,omitempty" yaml:"vpc,omitempty"`
-	RouteTableId string               `json:"routeTableId,omitempty" yaml:"routeTableId,omitempty"`
+	Version          string                  `json:"version,omitempty" yaml:"version,omitempty"`
+	EncryptionConfig []*eks.EncryptionConfig `json:"encryptionConfig,omitempty" yaml:"encryptionConfig,omitempty"`
```
sagikazarmark, line 36: I'm not sure how I feel about using an EKS structure directly. Providing the same API is one thing, but this is a direct leak of the abstraction. I know transformation can be tedious...but it's kind of a coincidence that we expose the same API here. We have generated types, I'd rather use those and do a one time conversion somewhere. Then we can clearly separate our API from the underlying EKS API.

But as long as they are the same, I guess this is fine for the short-term.

pregnor: Then I'm going to typedef the AWS structure locally.

pregnor: (Or generate something somewhere if possible.)

pkg/providers/amazon/eks/encryption_configuration.go
pregnor: Yeah, you are absolutely right, I do know the nil-handling of the `len` function, it just bugs me to hell that a forced-to-have-one-item slice has a singular name and thus I always treated it as an object in my mind and thus missed the simplification chance.

I'm going to fix this to using len.

internal/cluster/distribution/eks/ekscluster/eks.go
sagikazarmark: There is already a generated version in the API package. But I would only use it on the HTTP API surface.

pregnor:
Unsetting the merge deadline, won't release this week.

pkg/providers/amazon/eks/encryption_configuration.go
pregnor: [Fixed](https://github.com/banzaicloud/pipeline/compare/963282696f03dbad5b8db17ab08e5f8e7901a596..5e416a8ca1511e125a88e52e1add1016d5f646b2).

internal/cluster/distribution/eks/ekscluster/eks.go
pregnor: Should I duplicate the redefine in `internal/cluster/distribution/eks/ekscluster` and import that package into the `.../eksprovider/workflow` sources or should I duplicate the redefine in those 2 packages?
Not sure if architecturally we allow such relation between the two, but if a duplication can be avoided I would do so, because I'm having a hard time accepting conceptually there is no single place to define this.
(For example `Subnet` is identically duplicated in the two.)

pregnor: I implemented the duplicate redefinition approach [here](https://github.com/banzaicloud/pipeline/compare/6472650b045852246c96e9524c145a422a7049b6..f5bb49ddce4628811c995a6da8cf2d09409a1d0b).
It's easy to move to any other possible implementation choice from here.

internal/cluster/distribution/eks/ekscluster/eks.go
pregnor: [Fixed](https://github.com/banzaicloud/pipeline/compare/6472650b045852246c96e9524c145a422a7049b6..f5bb49ddce4628811c995a6da8cf2d09409a1d0b).

pregnor:
FYI I have changed the API field's casing `keyArn`->`keyARN` [here](https://github.com/banzaicloud/pipeline/compare/f5bb49ddce4628811c995a6da8cf2d09409a1d0b..15ac782ceb94ccc0f7ba7a75826c526bebdaa31c).

Technically speaking both are accepted by the Golang JSON parsers we use when an API request is sent, so the user can specify it with arbitrary casing, the parser is case insensitive.
Practically speaking we have all kinds of initialism casings in the currently available API specification:
1. Upper case (Golang style): bootID , loadBalancerIPRange , machineID , podCIDR (2x), privateIP , securityGroupID , serviceCIDR , systemUUID , vnetSubnetID (2x), vpcID .
2. Mixed case (AWS/C#/Java style): `accessId` , `clientId` , `clusterId` , `clusterRoleId` , `deploymentId` , `desiredCpu` (and co.), `lbSubnetId1` , `lbSubnetId2` , `nodeInstanceRoleId` , `odPcts` , `onDemandPct` , `organizationId` , `orgId` , `parentId` , `processId` , `projectId` , `regionId` , `resourceId` , `routeTableId` , `secretId` , `subnetId` , `subnetIds` , `sumCpu` (and co.), `vcnId` , `vpcId` , `vswitchId` , `zoneId`.
3. Lower case: `rolearn` , `userarn`. 

I'm going to bring this topic up independently from the PR for discussion to find out whether we have a convention or whether we want to converge towards a convention in the near future or we deliberately allow al kinds of casinngs.
(Similarly `lowercased`, `camelCased` and `snake_cased` fields can also be found in the API specification, but that's a different topic.)

In the meantime I'm going to use the upper case Golang style variant for opinionated reasons, if anyone feels like something should be cased otherwise, change requests are welcome.

sagikazarmark:
> I'm going to bring this topic up independently from the PR for discussion to find out whether we have a convention or whether we want to converge towards a convention in the near future or we deliberately allow all kinds of casings.

Not sure if this is ADR worthy, but it could be. We don't have a better place for it at the moment anyway, so let's create an ADR for this.

pregnor:
Waiting for review.

orymate (approved):

bonifaido (approved):
👍

pregnor:
Alpha test:
```shell
regnor_27419:Pregnor-BanzaiCloud-MBP152019@194.39.44.200/Users/pregnor/development/src/github.com/banzaicloud/pipeline-enterprise (master u=)
> aws --profile banzaicloud-trash eks describe-cluster --name pregnor-alpha-200824-envelope-happy-2 --region us-east-1

# Begin: 2020-08-24T13:13:01.879+02:00

{
    "cluster": {
        "name": "pregnor-alpha-200824-envelope-happy-2",
        "arn": "arn:aws:eks:us-east-1:999653275263:cluster/pregnor-alpha-200824-envelope-happy-2",
        "createdAt": "2020-08-24T12:51:09.318000+02:00",
        "version": "1.17",
        "endpoint": "https://18C658BAE2778C69F7B60DC697B128BB.yl4.us-east-1.eks.amazonaws.com",
        "roleArn": "arn:aws:iam::999653275263:role/pipeline-eks-iam-pregnor-alpha-200824-ClusterRole-59PJS55DI70Q",
        "resourcesVpcConfig": {
            "subnetIds": [
                "subnet-06cfc97c7ee8a5b68",
                "subnet-00a9cb8b26fd48739"
            ],
            "securityGroupIds": [
                "sg-087bc0bd045645c92"
            ],
            "clusterSecurityGroupId": "sg-0acf5004ad2c811e6",
            "vpcId": "vpc-034e81c04ed48c69a",
            "endpointPublicAccess": true,
            "endpointPrivateAccess": true,
            "publicAccessCidrs": [
                "0.0.0.0/0"
            ]
        },
        "logging": {
            "clusterLogging": [
                {
                    "types": [
                        "api",
                        "audit",
                        "authenticator",
                        "controllerManager",
                        "scheduler"
                    ],
                    "enabled": false
                }
            ]
        },
        "identity": {
            "oidc": {
                "issuer": "https://oidc.eks.us-east-1.amazonaws.com/id/18C658BAE2778C69F7B60DC697B128BB"
            }
        },
        "status": "ACTIVE",
        "certificateAuthority": {
            "data": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUN5RENDQWJDZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwKY201bGRHVnpNQjRYRFRJd01EZ3lOREV4TURBeU5sb1hEVE13TURneU1qRXhNREF5Tmxvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBT1hCCjBTd2NmMnV3cXdUc1BrL2hIVFVhalRJY25sTmZuY1ZoeTFoblo1ZFphOEI5L0V0aTFZTUhBdzluTmVXMnBpQWcKVTNSTUFvL0w2MTc1UE9LQWxYK2hSVVZpcG0yeWpKNjdVYVJxb3lqYWptRHZrcTlVQnV1cXhIUXdvdlVIOUxWNgpTOGpXekdsRmpkSTdlcmt5NEhNTS9KTElodEVDTVR6YkdVdmpnSVFEVmZYYkhUSlVzTTVNYVdaM21LUklvYmRSClUxNFIwdVltVUtCQ0pTc3pDcjlVZ3NucGF3eGlnb0t5UUowQ3hiM3htaVk1WVZqQXdiY1RVVmpVT3JsRWNJbmQKMnppdUV0enpUNHVkcnJuaWNjVFpHWEZtRFo2UEZaMGtjbHYzY0ZwVUJlWWRuazZRYWI2VWVGZ2ZNcExvZXFCbQo5ajU5a0gvWEV5dndIMGNnY0tzQ0F3RUFBYU1qTUNFd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dFQkFHMjNOdEJzT1paU1RkT1hUVUtUelRKZVFpWWwKOHdJZjE5M3FvOTNBSmQ0a2FpKzBwMWpnUEU0WGFvM2xHYndHSXJtZURNU2wyYVZvWkQ1c2FkQW4ySVV2NEFUUApSUHg3M0dxTEgyZVIyZW9mWm1jZ2sxTmxnNS83MFdJUmdBNDNKN2tqM3lJMFFwREpnak1MWnc5bVJNT1VGNkQvCnhGSmhZMU5tWE1Ib05iNWJLcEdqM0JqeFlZbG9oS05IWExHY010S2NzUEpvYk5ZbW1JaFAvVVlCd2NSbTJMT1QKdzBEVys0YndCYmZpOWp6OURpTzEzcVROTVZ1T0gvT0JLWVlMcDlJK2tORFR6YWRGMXdRZFoyNHRKYTBBVEtoUgpuN0dIcWp6Ylc0QzBPT2ttaERDVDBMVzZETTE1V1djM3lYajNWazEveEV5M3N0eW03YXIrdkxEdG5lTT0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo="
        },
        "platformVersion": "eks.2",
        "tags": {
            "banzaicloud-pipeline-managed": "true"
        },
        "encryptionConfig": [
            {
                "resources": [
                    "secrets"
                ],
                "provider": {
                    "keyArn": "arn:aws:kms:us-east-1:999653275263:key/a60e83a6-7e06-4aa5-9a89-153b8f64ccbb"
                }
            }
        ]
    }
}

# End: 2020-08-24T13:13:09.244+02:00
# Duration: 7s365ms
# Exit code: 0
```

pregnor:
Alpha test:
```shell
pregnor_27419:Pregnor-BanzaiCloud-MBP152019@194.39.44.200/Users/pregnor/development/src/github.com/banzaicloud/pipeline-enterprise (master u=)
> aws --profile banzaicloud-trash eks describe-cluster --name pregnor-alpha-200824-envelope-happy-2 --region us-east-1

# Begin: 2020-08-24T13:13:01.879+02:00

{
    "cluster": {
        "name": "pregnor-alpha-200824-envelope-happy-2",
        "arn": "arn:aws:eks:us-east-1:-:cluster/pregnor-alpha-200824-envelope-happy-2",
        "createdAt": "2020-08-24T12:51:09.318000+02:00",
        "version": "1.17",
        "endpoint": "https://18C658BAE2778C69F7B60DC697B128BB.yl4.us-east-1.eks.amazonaws.com",
        "roleArn": "arn:aws:iam::-:role/pipeline-eks-iam-pregnor-alpha-200824-ClusterRole-59PJS55DI70Q",
        "resourcesVpcConfig": {
            "subnetIds": [
                "subnet-06cfc97c7ee8a5b68",
                "subnet-00a9cb8b26fd48739"
            ],
            "securityGroupIds": [
                "sg-087bc0bd045645c92"
            ],
            "clusterSecurityGroupId": "sg-0acf5004ad2c811e6",
            "vpcId": "vpc-034e81c04ed48c69a",
            "endpointPublicAccess": true,
            "endpointPrivateAccess": true,
            "publicAccessCidrs": [
                "0.0.0.0/0"
            ]
        },
        "logging": {
            "clusterLogging": [
                {
                    "types": [
                        "api",
                        "audit",
                        "authenticator",
                        "controllerManager",
                        "scheduler"
                    ],
                    "enabled": false
                }
            ]
        },
        "identity": {
            "oidc": {
                "issuer": "https://oidc.eks.us-east-1.amazonaws.com/id/18C658BAE2778C69F7B60DC697B128BB"
            }
        },
        "status": "ACTIVE",
        "certificateAuthority": {
            "data": "-"
        },
        "platformVersion": "eks.2",
        "tags": {
            "banzaicloud-pipeline-managed": "true"
        },
        "encryptionConfig": [
            {
                "resources": [
                    "secrets"
                ],
                "provider": {
                    "keyArn": "arn:aws:kms:us-east-1:-:key/-"
                }
            }
        ]
    }
}

# End: 2020-08-24T13:13:09.244+02:00
# Duration: 7s365ms
# Exit code: 0
```

PR #3119, opened 2020-08-17
chore(generation): keep BUILD file

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

This workaround saves the plz BUILD configuration file. 

### Why?

The OpenAPI code regeneration deletes all generated files. 

### Additional context

Tried moving it to the .gen/pipeline/go directory in advance or moving away and back before the generation completes, but those approaches did not work.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ OPENAPI_DESCRIPTOR = apis/pipeline/pipeline.yaml
 # Build variables
 BUILD_DIR ?= build
 BUILD_PACKAGE = ${PACKAGE}/cmd/pipeline
+TEMPORARY_DIRECTORY = tmp
 VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || git symbolic-ref -q --short HEAD)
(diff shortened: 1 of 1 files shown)


orymate:
shouldn't we add `/tmp` to gitignore?

pregnor:
> shouldn't we add /tmp to gitignore?
`tmp` is left empty after the process so no difference happens as git doesn't care about empty directories, only files, but I can add it to the ignored files list, give me a sec.

pregnor:
The process could even be improved if I were capable of specifying cascadic delete from `tmp/.gen/pipeline/pipeline/BUILD` upwards through empty directories, going to check that idea out.

pregnor:
Significant changes were pushed so I re-requested the reviews as a dismissal of previous approves in order to ensure integrity.

sagikazarmark:
Doesn't matter that much, it's temporary anyway

Issue #3094, opened 2020-08-05
EKS cluster version upgrade


balazsbc:
For supporting cluster version upgrades we start with EKS.

Pipeline currently supports Kubernetes versions on EKS from 1.14 to 1.17

According to the [official EKS guide](https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html) the following steps are needed for upgrading the Kubernetes master node:

1. Update the cluster `version` with the [update API call](https://docs.aws.amazon.com/sdk-for-go/api/service/eks/#EKS.UpdateClusterVersion)
2. Wait of the update to be finished with [describe API call](https://docs.aws.amazon.com/sdk-for-go/api/service/eks/#EKS.DescribeUpdate)
3. Save the new version information to the DB.
**This is where the open-source version ends.**
4. Update image in the `kube-proxy` DaemonSet on the cluster according to the `version`:
    ```bash
    kubectl set image daemonset.apps/kube-proxy \
        -n kube-system \
        kube-proxy=602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/kube-proxy:v1.17.7
    ```
5. Patch the AWS VPC CNI image version in the `aws-node` DaemonSet, according to the above-linked docs if it is earlier than `1.6.3`:
    ```bash
    kubectl describe daemonset aws-node --namespace kube-system | grep Image | cut -d "/" -f 2
    ```

**NOTE:** GPU nodes must update the NVIDIA device plugin DaemonSet, but that is not in-scope for this task.

Implementation details:
- Implement this in an `eks-update-cluster` workflow - specifically for each provider behind the same API call.
- We can reuse the already existing Pipeline API [endpoint](https://github.com/banzaicloud/pipeline/blob/master/apis/pipeline/pipeline.yaml#L176) as a frontend for this when the future `version` attribute changes we start the upgrade.

bonifaido:
Open-source version done, enterprise version tracked in that repository.

PR #3123, opened 2020-08-19
eks: cluster version update

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | partially #3094 

### What's in this PR?

Basic implementation of EKS cluster version update.

### Additional context

Testable until the CLI implementation becomes available with:
```bash
curl -k -X 'PUT' -H 'Accept: application/json' \
-H "Authorization: Bearer ${PIPELINE_API_TOKEN}" \
-H 'Content-Type: application/json' \
https://127.0.0.1:9090/pipeline/api/v1/orgs/1/clusters/1/update \
-d '{"version":"1.17"}'
```

### Checklist


internal/cluster/distribution/eks/eksprovider/workflow/update_eks_version_activity.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func (a *UpdateClusterVersionActivity) Execute(ctx context.Context, input UpdateClusterVersionActivityInput) (*UpdateClusterVersionActivityOutput, error) {
+	session, err := a.awsSessionFactory.New(input.OrganizationID, input.SecretID, input.Region)
+	if err = errors.WrapIf(err, "failed to create AWS session"); err != nil {
+		return nil, err
+	}
+
+	eksClient := eks.New(session)
```
pregnor, line 58: **Optional**: we **MAY** consider using an EKS API interface and a corresponding factory here for better mocking capability and thus testing capability.
The `eksinterface.EKSAPI` can be mocked locally similarly to how the `cloudformation.go.cloudFormationAPI` is done and a factory analogous to the `cloudformation.go.CloudFormationAPIFactory` can be added to the activity.

Same at:
https://github.com/banzaicloud/pipeline/pull/3123/files#diff-ad1a4a93014f9f6fe3506f142f5a11ffR53

internal/cluster/distribution/eks/eksprovider/workflow/wait_update_eks_version_activity.go
```diff
@@ -0,0 +1,85 @@
...
+func (a *WaitUpdateClusterVersionActivity) Execute(ctx context.Context, input WaitUpdateClusterVersionActivityInput) error {
+	session, err := a.awsSessionFactory.New(input.OrganizationID, input.SecretID, input.Region)
+	if err = errors.WrapIf(err, "failed to create AWS session"); err != nil {
+		return err
+	}
+
+	eksClient := eks.New(session)
+
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
```
pregnor, line 56: **Note**: I'm not sure if this ticker based waiting is the more conventional choice here, or the AWS status based request waiter solution applied [here](https://github.com/banzaicloud/pipeline/blob/master/internal/cluster/distribution/eks/eksprovider/workflow/create_eks_control_plane_activity.go#L207). Whichever is the current conventional approach (preferably the one containing less custom logic) is fine by me. Also apologies if I missed some condition which invalidates the other approach.

(The other approach bypasses the separate waiter activity entirely if I'm reading it correctly which seems to spare a lot of lines of code, but I really don't want to force it if the ticker is the better choice for our current conventions.)

internal/cluster/distribution/eks/eksprovider/workflow/save_cluster_version_activity.go
```diff
@@ -0,0 +1,54 @@
...
+
+package workflow
+
+import (
+	"context"
+
+	"github.com/banzaicloud/pipeline/internal/cluster/distribution/eks/eksmodel"
+)
+
+const SaveClusterVersionActivityName = "eks-save-cluster-version"
```
pregnor, line 23: **Question**: is `save` the correct terminology here?

On opinionated ground I **MAY** propose `persist` or `record` here, because saving intuitively felt like something used for backup/restore or other failsafe mechanisms, but in reality it seems like only an administrational step for an information source.

Feel free to correct me if I'm misreading the nomenclature here.

src/cluster/eks_update_cluster_version.go
```diff
@@ -0,0 +1,117 @@
...
+	var updateOutput eksWorkflow.UpdateClusterVersionActivityOutput
+	{
+		activityInput := eksWorkflow.UpdateClusterVersionActivityInput{
+			EKSActivityInput: commonActivityInput,
+			Version:          input.Version,
+		}
+		err := workflow.ExecuteActivity(ctx, eksWorkflow.UpdateClusterVersionActivityName, activityInput).Get(ctx, &updateOutput)
+		if err != nil {
+			err = errors.WrapIff(pkgCadence.UnwrapError(err), "%q activity failed", eksWorkflow.UpdateClusterVersionActivityName)
+			eksWorkflow.SetClusterStatus(ctx, input.ClusterID, pkgCluster.Warning, err.Error()) // nolint: errcheck
```
pregnor, line 79:
```suggestion
			_ = eksWorkflow.SetClusterStatus(ctx, input.ClusterID, pkgCluster.Warning, err)
```

**Required**: we **MUST** unwrap the wrapped error here otherwise the cluster status message might end up being short without details such as `CLOUDFORMATION_STACK_FAILED` only, due to our Cadence error wrapping mechanism.
The `eksWorkflow.SetClusterStatus` function does the unwrapping, but it requires the original error, not just the error string for the details to be transferred.

Same at
https://github.com/banzaicloud/pipeline/pull/3123/files#diff-170c8394b68e0b56c2acd006e859f9c9R97
https://github.com/banzaicloud/pipeline/pull/3123/files#diff-170c8394b68e0b56c2acd006e859f9c9R110

internal/cluster/distribution/eks/eksprovider/workflow/save_cluster_version_activity.go
bonifaido: I used `save` because that was used across the activities here, like `eks-save-network-details` and `eks-save-node-pools`, I think we need to change them as well in this case.

src/cluster/eks_update_cluster_version.go
bonifaido: Thanks, fixed!

internal/cluster/distribution/eks/eksprovider/workflow/wait_update_eks_version_activity.go
bonifaido: TBH I started to use the waiter to resolve this (I have never used it before), but doesn't gave me better looking or even shorter code inside `WaitUpdateClusterVersionActivity.Execute`.

internal/cluster/distribution/eks/eksprovider/workflow/save_cluster_version_activity.go
pregnor: You are right, those examples indicate we should conventionally use save here.

It is a discussion separate from this PR to decide whether we want to change the corresponding terminology.

internal/cluster/distribution/eks/eksprovider/workflow/wait_update_eks_version_activity.go
pregnor: All right.
I'm leaving the discussion unresolved so others can read it, but otherwise I consider this thread closed.

internal/cluster/distribution/eks/eksprovider/workflow/update_eks_version_activity.go
bonifaido: Good point, I will add this.

pregnor: Thanks for the fix.

pregnor (approved):
My requested changes were implemented.

The topic of the proper separation of the cluster version and node pool upgrade flows are still up for discussion, but even if we decide to change the location of separation/branching we can implement it in another PR if we wish.
(It can still be discussed and implemented here, I'm just not sure this PR is the right way to do it, maybe it calls for a separate discussion or an ADR proposal.)

.gen/pipeline/api/openapi.yaml
```diff
@@ -11813,6 +11813,139 @@ paths:
...
+                OrganizationNotFound:
+                  value:
+                    code: 404
+                    message: Organization not found
+                    error: Organization not found
+                SecretAlreadyExists:
+                  value:
+                    code: 409
+                    message: Secret already exists
+                    error: Secret already exists
```
pregnor, line 11869: **Note**: is this the correct message for a 409 Conflict here?
I know it's auto-generated and not manually specified, but it might be worth an investigation when we have the time, although it is out of the scope of this PR I guess.

Same at
https://github.com/banzaicloud/pipeline/pull/3123/files#diff-2b6bfc5b5d040867ca5e5f7b1df191f1R11921

.gen/pipeline/api/openapi.yaml
```diff
@@ -11813,6 +11813,139 @@ paths:
...
+                  value:
+                    type: about:blank
+                    title: Access denied
+                    status: 403
+                    detail: You are unauthorized to perform this action access this
+                      resource.
+                    instance: ""
+                    code: 403
+                    message: You are unauthorized to perform this action access this
+                      resource.
```
pregnor, line 11909: **Note**: I think this message could be improved by either using `You are unauthorized to perform this action` or `You are unauthorized to perform this action on the corresponding resource`.
I know this is auto-generated and not manually specified, but it might be worth an investigation when we have the time, although it is out of the scope of this PR I guess.

apis/pipeline/pipeline.yaml
```diff
@@ -2662,6 +2662,31 @@ paths:
                 default:
                     $ref: '#/components/responses/Error'
 
+    /api/v1/orgs/{orgId}/clusters/{id}/update:
```
pregnor, line 2665: **Note**: I haven't made up my mind yet whether having a `/api/v1/orgs/{orgId}/clusters/{id}` endpoint for node pool update (and potentially other cluster level, under-cluster resource updates) and  `/api/v1/orgs/{orgId}/clusters/{id}/update` for cluster version update is clear/intuitive enough from user perspective, going to think on it more, I don't have any alternative suggestions for now.

(I think the distinction between the cluster master node changes and everything else "under" the cluster such as other nodes *can* be communicated and understood well enough, it's only the endpoint paths I'm still chewing on, but for now we are kind of locked in with the latter due to backward compatibility reasons anyway and I have nothing against the new endpoint path.)

pregnor (approved):

.gen/pipeline/api/openapi.yaml
sagikazarmark: This is the generated descriptor. Please comment on the original one.

.gen/pipeline/api/openapi.yaml
sagikazarmark: This is the generated descriptor. Please comment on the original one.

sagikazarmark (changes requested):

apis/pipeline/pipeline.yaml
```diff
@@ -2662,6 +2662,31 @@ paths:
...
+                required: true
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/ClusterUpdateRequest'
+            responses:
+                202:
+                    description: Accepted
+                default:
+                    $ref: '#/components/responses/Error'
```
pregnor, line 2688: **Recommended**: we **SHOULD** update the the `409 Conflict` `error` and `message` here, because it refers to the secret by default.
https://github.com/banzaicloud/pipeline/pull/3123/files#diff-2b6bfc5b5d040867ca5e5f7b1df191f1R11868-R11869

Also we **SHOULD** update the `403 Access Denied` `message` at the default values, because it seems to be misphrased `You are unauthorized to perform this action access resource.`.

.gen/pipeline/api/openapi.yaml
pregnor: Follow up at: https://github.com/banzaicloud/pipeline/pull/3123/files#r478930068

.gen/pipeline/api/openapi.yaml
pregnor: Follow up at https://github.com/banzaicloud/pipeline/pull/3123/files#r478930068

apis/pipeline/pipeline.yaml
```diff
@@ -2662,6 +2662,31 @@ paths:
...
+            - $ref: '#/components/parameters/clusterId'
+
+        put:
+            security:
+                - bearerAuth: []
+            tags:
+                - cluster-update
+            summary: Update a cluster
+            description: Update a cluster
+            operationId: ClusterUpdate
```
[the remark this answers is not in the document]
bonifaido: The issue is that UpdateCluster is taken by the old update.

apis/pipeline/pipeline.yaml
```diff
@@ -7503,6 +7527,13 @@ components:
                     type: string
                     example: 10.20.30.40
 
+        ClusterUpdateRequest:
```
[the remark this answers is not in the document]
bonifaido: The nodepool update request looks a bit bad to me, it has cloud or distribution for example. Or do you mean the EKS part?

apis/pipeline/pipeline.yaml
```diff
@@ -2662,6 +2662,31 @@ paths:
...
+    /api/v1/orgs/{orgId}/clusters/{id}/update:
+        parameters:
+            - $ref: '#/components/parameters/orgId'
+            - $ref: '#/components/parameters/clusterId'
+
+        put:
+            security:
+                - bearerAuth: []
+            tags:
+                - cluster-update
```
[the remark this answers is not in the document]
bonifaido: fixed, thanks.

internal/cluster/distribution/eks/eksprovider/workflow/eks.go
```diff
@@ -0,0 +1,61 @@
...
+// is referenced.
+type eksAPI interface {
+	eksi.EKSAPI
+}
+
+// +testify:mock
+
+// EKSAPIFactory provides an interface for instantiating AWS
+// EKS API objects.
+type EKSAPIFactory interface {
```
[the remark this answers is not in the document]
bonifaido: This is just a wrapper to not to rely directly on the AWS lib, for mocking. @pregnor am I right?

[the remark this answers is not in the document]
pregnor: That's correct, we use it for test mocking purposes to be able to provide a mocked API in place of an `eksiface.EKSAPI`, we did it similarly with the `CloudFormation` factories and client/service interfaces.

Context: It helps the Cadence activities to have a constructor argument/field for factories returning client/service interfaces to the corresponding AWS APIs so we don't use `eks.New(awsSession)` because it makes test mocking much more difficult as it creates a real API object, this way we can use a `EKSFactoryMock{}` object in place of the real factory wrapping the `New()` call which allows us to write extensive mock tests.

apis/pipeline/pipeline.yaml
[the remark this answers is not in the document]
sagikazarmark: I don't mind changing that. We don't have a BC promise for the generated code.

apis/pipeline/pipeline.yaml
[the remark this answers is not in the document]
sagikazarmark: I mean this:

```
        UpdateNodePoolRequest:
            oneOf:
                - $ref: '#/components/schemas/EksUpdateNodePoolRequest'
```

We should have a similar structure for clusters IMHO.

sagikazarmark (approved):
LGTM

pregnor (approved):

PR #3138, opened 2020-08-28
Add snapshot script

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
A new snapshot script for creating snapshots instead of the old git tag workflow.

### Why?
We don't want to spam the repo with tags created for internal testing.

### DIFF
--- a/BUILD.plz
+++ b/BUILD.plz
@@ -84,6 +84,7 @@ Arguments:
 
 Options:
   -r, --ref              The ref to create the tag for.
+  --major                Bump a major version on master instead of a minor version.
   -n, --dry-run          Print the commands to be executed.
   -h, --help             Print this help message.
 
@@ -112,6 +113,7 @@ function error {
 
 # Default values of arguments
 REF=""
+MAJOR=false
 DRY_RUN=false
 HELP=false
 ARGS=()
@@ -123,6 +125,10 @@ do
         -r|--ref)
         REF="\\\$2"
         shift 2 # Remove argument name and value from processing
+        ;;
+        --major)
+        MAJOR=true
+        shift # Remove --major from processing
         ;;
         -n|--dry-run)
         DRY_RUN=true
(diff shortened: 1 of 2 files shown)


BUILD.plz
```diff
@@ -61,6 +61,171 @@ sh_cmd(
...
+        ;;
+        *)
+        ARGS+=("\\\$1")
+        shift # Remove generic argument from processing
+        ;;
+    esac
+done
+
+# Setting here so the last shift won't fail the script
+set -o errexit
```
pregnor, line 143: **Optional**: we **MAY** consider moving this closer to the `# Auto-detect tag (if possible)` conditional statement to make modifications to the `plz` command less error-prone thus avoiding the introduction of non-failing pipe commands between the setting and the mentioned conditional statement when those new commands are supposed to fail on pipe failure, but the author and reviewer miss the setting.

In addition I would consider turning pipefail off for the mentioned conditional block only and reenable pipefail right after that.

BUILD.plz
```diff
@@ -61,6 +61,171 @@ sh_cmd(
...
+        fi
+
+        TAG="\\\$($(out_location ///pleasings2//tools/misc:semver) bump patch \\\$LATEST_VERSION)-snapshot.\\\\$(date +%s)"
+    elif [[ "\\\$CURRENT_BRANCH" == "master" ]]; then
+        LATEST_VERSION="\\\$(git tag | grep -E "\\\$VERSION_REGEX" | sort -r --version-sort | head -1)"
+        if [[ "\\\$LATEST_VERSION" == "" ]]; then
+            error "failed to determine latest version"
+        fi
+
+        TAG="\\\$($(out_location ///pleasings2//tools/misc:semver) bump minor \\\$LATEST_VERSION)-snapshot.\\\\$(date +%s)"
```
pregnor, line 188: **Note**: post 1.0.0 we could introduce a flag in this command for major version bumped snapshots and conditionally branch on it here.

pregnor (approved):

BUILD.plz
orymate, line 143: The `shift`s should not fail in the current code (and setting `-e` earlier would detect errors like when you run the script with `--ref` but no argument to it). So I'd set it at the beginning.

BUILD.plz
sagikazarmark: Done

BUILD.plz
sagikazarmark: Done

pregnor (approved):

orymate (approved):

asdwsda (approved):

PR #3139, opened 2020-08-28
ci(GHA): added generated file integrity check

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Added generated file integrity check to the CIs.

Example:
```shell
generated files are not up to date with the configured dependency versions or current source code files, differences:

diff --git a/internal/cluster/distribution/eks/eksprovider/workflow/zz_generated.mock.go b/internal/cluster/distribution/eks/eksprovider/workflow/zz_generated.mock.go
index ed8dedf..519e9f8 100644
--- a/internal/cluster/distribution/eks/eksprovider/workflow/zz_generated.mock.go
+++ b/internal/cluster/distribution/eks/eksprovider/workflow/zz_generated.mock.go
@@ -2298,6 +2298,20 @@ func (_m *MockcloudFormationAPI) DetectStackSetDriftWithContext(_parameter_0 con
 	return r0, r1
 }
 
+// Error provides a mock function.
+func (_m *MockcloudFormationAPI) Error() (_result_0 string) {
+	ret := _m.Called()
+
+	var r0 string
+	if rf, ok := ret.Get(0).(func() string); ok {
+		r0 = rf()
+	} else {
+		r0 = ret.Get(0).(string)
+	}
+
+	return r0
+}
+
 // EstimateTemplateCost provides a mock function.
 func (_m *MockcloudFormationAPI) EstimateTemplateCost(_parameter_0 *cloudformation.EstimateTemplateCostInput) (_result_0 *cloudformation.EstimateTemplateCostOutput, _result_1 error) {
 	ret := _m.Called(_parameter_0)
```

### Why?

To avoid desyncing client version and generated files, OpenAPI specification and generated files, source code files and generated files.

### Checklist

### To Do

- ~Add to CircleCI.~ CircleCI has some issues such as deleting generated folders with sude and stuff which would require substantially more effort to resolve.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,12 @@ jobs:
         if: steps.go-module-cache.outputs.cache-hit != 'true'
         run: go mod download
 
+      - name: Check generated file integrity
+        run: |
+          make generate-all \
+          && test "$(git diff --name-only)" == "" \
+          || ( printf >&2 "\ngenerated files are not up to date with the configured dependency versions or current source code files, differences:\n\n%s\n\n" "$(git diff)" ; exit 1 ; )
+
       - name: Build
         run: ./pleasew -p --profile ci build //...
 
@@ -184,7 +190,6 @@ jobs:
         with:
           name: lint-log
           path: plz-out/log
-
 
   docker:
     name: Docker
--- a/Makefile
+++ b/Makefile
@@ -233,6 +233,9 @@ bin/mga-${MGA_VERSION}:
 	curl -sfL https://git.io/mgatool | bash -s v${MGA_VERSION}
 	@mv bin/mga $@
 
+.PHONY: generate-all
+generate-all: generate-anchore-client generate-cloudinfo-client generate generate-openapi
+
 .PHONY: generate
 generate: bin/mga ## Generate code
 	go generate -x ./...
--- a/Makefile
+++ b/Makefile
@@ -292,8 +292,8 @@ apis/anchore/swagger.yaml:
 generate-anchore-client: apis/anchore/swagger.yaml ## Generate client from Anchore OpenAPI spec
 	$(call back_up_file,.gen/anchore/BUILD)
 	$(call generate_openapi_client,apis/anchore/swagger.yaml,anchore,.gen/anchore)
-	sed -i '' 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go
-	sed -i '' 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go
+	@ sed -i~ 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go && rm .gen/anchore/model_mapping_rule.go~
+	@ sed -i~ 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go && rm .gen/anchore/model_policy_rule.go~
 	$(call restore_backup_file,.gen/anchore/BUILD)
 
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -185,7 +191,6 @@ jobs:
           name: lint-log
           path: plz-out/log
 
-
```
pregnor, line 188: **Note**: this change came from the autoformatter, but unless we have a serious reason to keep 2 empty lines here, I would let it be merged.

Makefile
```diff
@@ -289,8 +292,7 @@ apis/anchore/swagger.yaml:
 generate-anchore-client: apis/anchore/swagger.yaml ## Generate client from Anchore OpenAPI spec
 	$(call back_up_file,.gen/anchore/BUILD)
 	$(call generate_openapi_client,apis/anchore/swagger.yaml,anchore,.gen/anchore)
-	sed -i '' 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go
-	sed -i '' 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go
+	@ if [[ "$$OSTYPE" == "linux-gnu" ]] ; then sed -i 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go ; sed -i 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go ; else sed -i '' 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go ; sed -i '' 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go ; fi
```
orymate, line 295:
```suggestion
	@ if sed --version &>/dev/null ; then sed -i 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go ; sed -i 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go ; else sed -i '' 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go ; sed -i '' 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go ; fi
```

pregnor: If you insist, I'm willing to use your version, but it kind of bugs that it doesn't convey clear information on why the conditional branching is required, it conceals the connection between the macOS/Linux sed flag differences.

If I'm seeing this code for the first time, I'm going to have a hard time understanding how the --version flag correlates to using the in-place flag two different way.

orymate, line 295:
```suggestion
	@ sed -i~ 's/whitelist_ids,omitempty/whitelist_ids/' .gen/anchore/model_mapping_rule.go && rm .gen/anchore/model_mapping_rule.go\~
	@ sed -i~ 's/params,omitempty/params/' .gen/anchore/model_policy_rule.go && rm .gen/anchore/model_policy_rule.go~
```

pregnor: See the other suggestion https://github.com/banzaicloud/pipeline/pull/3139#discussion_r479152479

pregnor: [Fixed](https://github.com/banzaicloud/pipeline/compare/18895619c902b797e4a1d881f2728d04194fa616..f87f6c595f10a862dff24060b4476da9147a3179).

orymate: the intent behind this approach was that `--version` is a gnu addition just like the optional argument of `-i`

pregnor: Yeah, I understood it, but it is more of an implicit GNU-specific thing than the OS type which means it's more difficult to understand for a newcomer who meets the issue here the first time.

sagikazarmark (approved):
LGTM

PR #3144, opened 2020-09-01
eks: enterprise cluster update workflow

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/pipeline-enterprise/pull/101

### What's in this PR?

Starting the enterprise EKS cluster update workflow in case of an enterprise deployment.
Removing the autoscaler update skip in case of EKS.

### Additional context

Has to be merged after: https://github.com/banzaicloud/pipeline-enterprise/pull/101

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/eksadapter/cluster_manager.go
+++ b/internal/cluster/distribution/eks/eksadapter/cluster_manager.go
@@ -28,23 +28,33 @@ import (
 
 type clusterManager struct {
 	workflowClient client.Client
+	enterprise     bool
 }
 
 // NewClusterManager returns a new eks.ClusterManager
 // that manages clusters asynchronously via Cadence workflows.
-func NewClusterManager(workflowClient client.Client) eks.ClusterManager {
+func NewClusterManager(workflowClient client.Client, enterprise bool) eks.ClusterManager {
 	return clusterManager{
 		workflowClient: workflowClient,
+		enterprise:     enterprise,
 	}
 }
 
-func (n clusterManager) UpdateCluster(
+func (m clusterManager) UpdateCluster(
 	ctx context.Context,
 	c cluster.Cluster,
 	clusterUpdate eks.ClusterUpdate,
 ) error {
+	taskList := "pipeline"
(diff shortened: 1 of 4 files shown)


pregnor (approved):
LGTM, but I would wait for Mark's opinion to be sure.

bonifaido:
Thanks for the check @pregnor

sagikazarmark (approved):
Looks ok for now, but I think it would be better to duplicate the workflow and replace the tasklist instead, just like in case of the node pool upgrade.

bonifaido:
Just duplicated the workflow on the enterprise side.

internal/cluster/distribution/eks/eksadapter/cluster_manager.go
```diff
@@ -28,23 +28,33 @@ import (
...
 	ctx context.Context,
 	c cluster.Cluster,
 	clusterUpdate eks.ClusterUpdate,
 ) error {
+	taskList := "pipeline"
+	workflowName := eksworkflow.UpdateClusterWorkflowName
+
+	if m.enterprise {
+		taskList = "pipeline-enterprise"
+		workflowName = "eks-update-cluster"
```
sagikazarmark, line 53: Why is this different?

sagikazarmark (approved):

internal/cluster/distribution/eks/eksadapter/cluster_manager.go
bonifaido: We have to use "-v2" in pipeline, since the previous is already taken, but we can change to "-v2" in enterprise as well.

sagikazarmark: How about renaming the old one to something else?

bonifaido: If that doesn't break backward compatibility I'm happy to rename it (since probably this will be the official update workflow and that one will be deprecated).

PR #3147, opened 2020-09-02
Validate cluster role ID

### LINKED ISSUE #3034: EKS cluster role id is not validated properly
**Describe the bug**
If I create an EKS cluster with a role ARN instead of its name, the cluster is created, but at the end it will go to error because the EKS resource created has an incorrect cluster role. Fix the heuristics or validate the role id earlier.

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3034

### What's in this PR?
Validate cluster role ID in case of EKS

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/eksprovider/workflow/create_infra.go
+++ b/internal/cluster/distribution/eks/eksprovider/workflow/create_infra.go
@@ -110,6 +110,25 @@ func CreateInfrastructureWorkflow(ctx workflow.Context, input CreateInfrastructu
 
 	ctx = workflow.WithActivityOptions(ctx, ao)
 
+	// create IAM role validate activity
+	var validateIAMRoleActivityFuture workflow.Future
+	if input.ClusterRoleID != "" {
+		{
+			activityInput := &ValidateIAMRoleActivityInput{
+				EKSActivityInput: commonActivityInput,
+				ClusterRoleID:    input.ClusterRoleID,
+			}
+			validateIAMRoleActivityFuture = workflow.ExecuteActivity(ctx, ValidateIAMRoleActivityName, activityInput)
+		}
+	}
+
+	if validateIAMRoleActivityFuture != nil {
+		validateIAMRoleActivityOutput := ValidateIAMRoleActivityOutput{}
+		if err := validateIAMRoleActivityFuture.Get(ctx, &validateIAMRoleActivityOutput); err != nil {
+			return nil, err
+		}
+	}
+
 	// create IAM roles activity
 	var iamRolesCreateActivityFuture workflow.Future
 	{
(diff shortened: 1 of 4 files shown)


internal/cluster/distribution/eks/eksprovider/driver/cluster_creator.go
```diff
@@ -333,6 +334,15 @@ func (c *EksClusterCreator) validate(r *pkgCluster.CreateClusterRequest, logger
 		return errors.WrapIf(err, "failed to create AWS session")
 	}
 
+	if r.Properties.CreateClusterEKS.IAM.ClusterRoleID != "" {
+		iamSession := iam.New(awsSession)
```
sagikazarmark, line 338: As per our recent validation efforts, in API requests we only do formal validation.

pregnor: Márk is referring to [this](https://github.com/banzaicloud/adr/blob/master/adrs/pipeline/0006-API-input-validation-user-feedback.md) ADR.

I think we should execute this validation during the cluster creation workflow asynchronously due to such a validation requiring additional AWS calls - in according to what we discussed about the validation kinds and user feedback relations and subsequently the validation location.

I would propose implementing a `GetIAMRole` activity in the `distribution/eks/provider/workflow` package to retrieve the role corresponding to the cluster role ID, execute it in the workflows where such a check is relevant - such as `CreateInfraWorkflow` - and fail the enclosing workflow in case no such role is found.

pregnor (commented):
See my [comment](https://github.com/banzaicloud/pipeline/pull/3147#discussion_r481934382).

internal/cluster/distribution/eks/eksprovider/driver/cluster_creator.go
colin014: thx 👍  I updated the PR

internal/cluster/distribution/eks/eksprovider/workflow/create_infra.go
```diff
@@ -110,6 +110,25 @@ func CreateInfrastructureWorkflow(ctx workflow.Context, input CreateInfrastructu
...
+			validateIAMRoleActivityFuture = workflow.ExecuteActivity(ctx, ValidateIAMRoleActivityName, activityInput)
+		}
+	}
+
+	if validateIAMRoleActivityFuture != nil {
+		validateIAMRoleActivityOutput := ValidateIAMRoleActivityOutput{}
+		if err := validateIAMRoleActivityFuture.Get(ctx, &validateIAMRoleActivityOutput); err != nil {
+			return nil, err
+		}
+	}
```
pregnor, line 130: **Question**: do we benefit from using the future here instead of calling get on the execute call directly?
It's fine by me either way, just don't see much of a use for the async option.

pregnor (approved):

internal/cluster/distribution/eks/eksprovider/workflow/create_infra.go
colin014: good point, I changed the code to calling it directly

pregnor (approved):

orymate:
Tested this locally, it works as expected.

orymate (approved):

PR #3168, opened 2020-09-15
vsphere: pass vcpu and ram to cloneSpec

orymate:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
VCPU and RAM inputs were missing from the clone specification when creating nodes.

### Checklist

### DIFF
--- a/internal/providers/vsphere/pke/workflow/create_node_activity.go
+++ b/internal/providers/vsphere/pke/workflow/create_node_activity.go
@@ -73,7 +73,7 @@ type CreateNodeActivityInput struct {
 type Node struct {
 	AdminUsername          string
 	VCPU                   int
-	RAM                    int
+	RAM                    int // MiB
 	Name                   string
 	SSHPublicKey           string
 	UserDataScriptParams   map[string]string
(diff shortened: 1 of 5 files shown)


internal/providers/vsphere/pke/workflow/create_node_activity.go
```diff
@@ -122,7 +122,11 @@ func generateVMConfigs(input CreateNodeActivityInput) (*types.VirtualMachineConf
 		return nil, err
 	}
 
-	vmConfig := types.VirtualMachineConfigSpec{}
+	vmConfig := types.VirtualMachineConfigSpec{
+		NumCPUs:           int32(input.VCPU),
+		NumCoresPerSocket: int32(input.VCPU),
```
pregnor, line 127: **Note**: I'm kinda out of the loop, but do we have 1 socket only (whatever it refers to)?

Before starting to look into what these config values mean exactly my intuition said `NumCPUs` is probably the total value and `socket * NumCoresPerSocket` should not be over that, but I could be completely wrong, I'm going to search for these, just wanted to start the discussion in the meantime.

pregnor (approved):
As far as I could tell with my limited contextual knowledge it looked good.

internal/providers/vsphere/pke/workflow/create_node_activity.go
orymate: I think there's no major drawback of putting all the virtual cores into one virtual socket (and I think the above spec means exactly that), but some strange commercial license checks tend to prefer this layout, and that's why I normally opt for it.

orymate: Additionally, if you want to have a custom layout, you can always set the vcpu value to zero, and that will fall back to the template's settings.

pregnor: Thanks.

sancyx (approved):

PR #3179, opened 2020-09-21
Node pool update API

### LINKED ISSUE #3178: PKE/AWS node pool update API
Implement node pool update and related APIs for PKE/AWS (including CLI support and process API). Use workflow from banzaicloud/pipeline-enterprise#110

bonifaido:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/pipeline-enterprise/issues/110, fixes #3178

### What's in this PR?

The pipeline part of https://github.com/banzaicloud/pipeline-enterprise/issues/110 (PKE nodepool update).

### Checklist


internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
```diff
@@ -0,0 +1,40 @@
...
+	NodePoolName    string
+
+	NodeImage string
+
+	Version string
+
+	Options pke.NodePoolUpdateOptions
+
+	ClusterTags map[string]string
+}
```
pregnor, line 40: **Question**: even with only supporting it in enterprise, shouldn't we rather have a regular registered activity here returning an error from its exec function?

I feel it would be more consistent with the current EKS node pool update workflow's "duplicate" approach.

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
```diff
@@ -0,0 +1,40 @@
...
+
+	StackName string
+
+	OrganizationID  uint
+	ClusterID       uint
+	ClusterSecretID string
+	ClusterName     string
+	NodePoolName    string
+
+	NodeImage string
```
pregnor, line 33: **Note**: I believe we are going to need a node pool version here as well.

Also it would be great to implement the update with volume sizes as well.

(As long as we want to use the same input in enterprise.)

internal/cluster/distribution/pke/service.go
```diff
@@ -0,0 +1,215 @@
...
+// NodePool encapsulates information about a cluster node pool.
+type NodePool struct {
+	Name         string            `mapstructure:"name"`
+	Labels       map[string]string `mapstructure:"labels"`
+	Size         int               `mapstructure:"size"`
+	Autoscaling  Autoscaling       `mapstructure:"autoscaling"`
+	VolumeSize   int               `mapstructure:"volumeSize"`
+	InstanceType string            `mapstructure:"instanceType"`
+	Image        string            `mapstructure:"image"`
+	SpotPrice    string            `mapstructure:"spotPrice"`
```
pregnor, line 107: **Note**: I have just merged an EKS side change adding subnet ID to this structure which was accidentally missing from here and also I'm currently working on a change probably adding a status and a status message strings to the EKS equivalent of this struct. Not sure whether you want to update this one accordingly, just wanted to let you know in case you would.

bonifaido: I just added the SubnetID as well, probably in the future it will be added (but not in this stage).

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
bonifaido: I added the version parameter in a subsequent commit.

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
bonifaido: Yeah, let's do it like that.

bonifaido: Done in the latest commit.

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"fmt"
+
+	"go.uber.org/cadence/workflow"
+
+	"github.com/banzaicloud/pipeline/internal/cluster"
+	"github.com/banzaicloud/pipeline/internal/cluster/distribution/pke"
+)
+
+const UpdateNodePoolWorkflowName = "pkeaws-update-node-pool"
```
pregnor, line 26: **Note**: FYI at the PKE Azure and vSphere implementation we used `pke-azure-<activity>` and `pke-vsphere-<activity>` activity names. Not sure if we want to do it that way here as well or use the `pkeaws-`, `pke-aws-`, `pkeamazon` or `pke-amazon`, just wanted to point it out and let you think about it, I'm unsure about what would be best.

internal/providers/pke/pkeworkflow/calculate_node_pool_version_activity.go
```diff
@@ -0,0 +1,63 @@
...
+func (a CalculateNodePoolVersionActivity) Execute(
+	_ context.Context,
+	input CalculateNodePoolVersionActivityInput,
+) (CalculateNodePoolVersionActivityOutput, error) {
+	return CalculateNodePoolVersionActivityOutput{
+		Version: eks.CalculateNodePoolVersion(
+			input.Image,
+			fmt.Sprint(input.VolumeSize),
+			input.Version,
+		),
```
pregnor, line 61: **Question**: shouldn't we use a pke variant of the calculation function instead of using the eks package's implementation?

It feels unusual to couple the `internal/providers/pke/pkeworkflow` to `internal/cluster/distribution/eks`, even though I understand the actual implementation is always going to be very similar and simple.

internal/providers/pke/pkeworkflow/ec2.go
```diff
@@ -17,10 +17,16 @@ package pkeworkflow
...
 	"github.com/aws/aws-sdk-go/aws/client"
+	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
+
+	internalAmazon "github.com/banzaicloud/pipeline/internal/providers/amazon"
 )
 
+// ErrReasonStackFailed cadence custom error reason that denotes a stack operation that resulted a stack failure
+const ErrReasonStackFailed = "CLOUDFORMATION_STACK_FAILED"
```
pregnor, line 28: **Recommended**: I think this **SHOULD** be in the `cloudformation.go` file instead of the `ec2.go`, because this is not part of EC2, but it is part of CloudFormation.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
```diff
@@ -0,0 +1,277 @@
...
+	pkgCluster "github.com/banzaicloud/pipeline/internal/cluster"
+	"github.com/banzaicloud/pipeline/pkg/brn"
+	pkgCloudFormation "github.com/banzaicloud/pipeline/pkg/providers/amazon/cloudformation"
+	sdkAmazon "github.com/banzaicloud/pipeline/pkg/sdk/providers/amazon"
+	sdkCloudFormation "github.com/banzaicloud/pipeline/pkg/sdk/providers/amazon/cloudformation"
+)
+
+const awsNoUpdatesError = "No updates are to be performed."
+
+const UpdateNodeGroupActivityName = "pkeaws-update-node-group"
```
pregnor, line 40: **Note**: in this package the consistent activity naming is `pke-<activity>`. Not sure what would be best here, I can accept being more strict in this one.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
```diff
@@ -0,0 +1,277 @@
...
+				err = errors.WrapIff(err, "waiting for %q CF stack create operation to complete failed", input.StackName)
+				if pkgCloudFormation.IsErrorFinal(err) {
+					return UpdateNodeGroupActivityOutput{}, cadence.NewCustomError(ErrReasonStackFailed, err.Error())
+				}
+				return UpdateNodeGroupActivityOutput{}, errors.WrapIff(err, "waiting for %q CF stack create operation to complete failed", input.StackName)
+			}
+		}
+	}
+
+	return UpdateNodeGroupActivityOutput{NodePoolChanged: true}, nil
```
pregnor, line 276: **Note**: not sure if it is intended, but currently in case no error occurs we always state the node pool has been changed which kind of defeats the purpose of including it in the output in my opinion unless we do it for interfacing purposes.

cmd/worker/pke.go
```diff
@@ -17,12 +17,17 @@ package main
...
-func registerPKEWorkflows(passwordSecrets pkeworkflow.PasswordSecretStore) {
+func registerPKEWorkflows(passwordSecrets pkeworkflow.PasswordSecretStore) error {
 	{
 		a := pkeworkflow.NewAssembleHTTPProxySettingsActivity(passwordSecrets)
 		activity.RegisterWithOptions(a.Execute, activity.RegisterOptions{Name: pkeworkflow.AssembleHTTPProxySettingsActivityName})
 	}
+
+	pkeawsworkflow.NewUpdateNodePoolWorkflow().Register()
+
+	return nil
```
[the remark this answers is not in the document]
bonifaido: Ah yes, this is a leftover! Thanks!

internal/providers/pke/pkeworkflow/calculate_node_pool_version_activity.go
bonifaido: Thanks, again, a leftover, I will fix this.

internal/providers/pke/pkeworkflow/ec2.go
bonifaido: Good point. Fixed.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
bonifaido: Yeah, this was basically copied from the EKS UpdateNodeGroupActivity, which returns `NodePoolChanged{}, nil` in an if case a few lines above, so I would leave it as here as it is.

pregnor: Yeah, but those cases also return an error in case the output should be discarded as might not contain valid values (as a general Golang idiom).

Anyway, I see your point and also because of staying consistent with EKS is a value I would keep this too, this is kinda the interfacing purpose I meant originally. 👍🏻

apis/pipeline/pipeline.yaml
```diff
@@ -4672,6 +4672,7 @@ components:
                 # https://github.com/OpenAPITools/openapi-generator/pull/2897 
                 - $ref: '#/components/schemas/EksUpdateNodePoolRequest'
                 - $ref: '#/components/schemas/EksUpdateNodePoolRequest'
+                - $ref: '#/components/schemas/PkeAwsUpdateNodePoolRequest'
```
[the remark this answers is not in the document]
bonifaido: I don't really get, so should be PkeAwsUpdateNodePoolRequest listed two times as well?

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
bonifaido: I followed mostly the EKS pattern here, so PKE AWS namings are now a mix of EKS and general PKE and some PKE AWS specific namings. Let me check this! :)

apis/pipeline/pipeline.yaml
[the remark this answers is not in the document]
pregnor: AFAIK the note and the second `EksUpdateNodePoolRequest` can be removed once we add this second reference, but make sure and check it with OpenAPI generation.

The issue only affects single reference composition types.

(Sorry I missed it on review.)

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
bonifaido: TBH `pke-aws-` looks much better to me (and we have `pke-azure-`, and `pke-vsphere-` already), so I will rename it like that if you are okay with ith.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
bonifaido: This is an interesting case, I would use the newly established `pke-aws-` naming here because that should have been used here as well in the first case. The general rule of thumb was that everything which is simple "pke" is the AWS implementation, but that was a bad idea in the original implementation.

I think we should use the pke-aws- naming and reregister the activities also with that name later on in this package (but keeping the old name as well).

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/workflow_update_node_pool.go
pregnor: That sounds fine.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
pregnor: I agree with your reasoning.

internal/providers/pke/pkeworkflow/update_node_group_activity.go
```diff
@@ -0,0 +1,277 @@
...
+		},
+		{
+			ParameterKey:     aws.String("MaxSize"),
+			UsePreviousValue: aws.Bool(true),
+		},
+		// sdkCloudFormation.NewOptionalStackParameter(
+		// 	"NodeAutoScalingGroupMaxBatchSize",
+		// 	input.MaxBatchSize > 0,
+		// 	fmt.Sprintf("%d", input.MaxBatchSize),
+		// ),
```
pregnor, line 190: **Question**: why is this commented out? (Do we need it? Does it have any value as a comment?)

pregnor (approved):

internal/providers/pke/pkeworkflow/update_node_group_activity.go
bonifaido: Yes, just removed it, once (if ever) we will have an open-source version of the upgrade flow we will probably used it (leftover from EKS).

pregnor (approved):

go.sum
```diff
@@ -330,6 +330,8 @@ github.com/denisenkom/go-mssqldb v0.0.0-20191001013358-cfbb681360f0/go.mod h1:xb
 github.com/denisenkom/go-mssqldb v0.0.0-20200206145737-bbfc9a55622e h1:LzwWXEScfcTu7vUZNlDDWDARoSGEtvlDKK2BYHowNeE=
 github.com/denisenkom/go-mssqldb v0.0.0-20200206145737-bbfc9a55622e/go.mod h1:xbL0rPBG9cCiLr28tMa8zpbdarY27NDyej4t/EjAShU=
 github.com/denverdino/aliyungo v0.0.0-20190125010748-a747050bb1ba/go.mod h1:dV8lFg6daOBZbT6/BDGIz6Y3WFGn8juu6G+CQ6LHtl0=
+github.com/dexidp/dex v0.6.1 h1:8cMvsPai2Eey6Ly77AMv/yO0Mlb8P4wVJHQW6rb9y68=
+github.com/dexidp/dex v2.13.0+incompatible h1:EQPpzCi51omkwBe0KYpRGaV3rk6CVvjcqeMGCe3Q00w=
```
sagikazarmark, line 334: This doesn't look good. Why do we need this?

internal/cluster/distribution/pke/service.go
```diff
@@ -0,0 +1,200 @@
...
+type NodePoolUpdate struct {
+	Image string `mapstructure:"image"`
+
+	Version string `mapstructure:"version"`
+
+	Options NodePoolUpdateOptions `mapstructure:"options"`
+}
+
+type NodePoolUpdateOptions struct {
+
```
sagikazarmark, line 64:
```suggestion
```

internal/cluster/service.go
```diff
@@ -355,7 +355,11 @@ func (NotSupportedDistributionError) ServiceError() bool {
 }
 
 func (s service) getDistributionService(cluster Cluster) (Service, error) {
-	service, ok := s.distributions[cluster.Distribution]
+	key := cluster.Distribution
+	if key == "pke" {
+		key += cluster.Cloud
+	}
+	service, ok := s.distributions[key]
```
sagikazarmark, line 362: This could go away if the pke cloud distribution was hidden inside the pke service. See my comment above.

go.sum
bonifaido: This is some kind of VScode/language server magic probably, it wasn't intentional.

apis/pipeline/pipeline.yaml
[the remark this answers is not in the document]
bonifaido: just removed the note, it works fine now

PR #3186, opened 2020-09-24
Please finally compiles everything

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
This PR adds proper Please support utilizing Go modules and generated library and test rules.

### Why?
To reach Please's full potential.

### Additional context
Initial build time might be worse, needs some measurements.

### To Do
- [ ] Separate integration tests (if necessary?)
- [x] Wollemi is still executed from a custom patched version (Latest release supports `import_path`)
- [ ] Need a way to detect diffs (from go modules and from package content) and notify the user
- [ ] We need a way that compares plain old `go build` and `plz build` (Circle vs GHA?)


sagikazarmark:
It doesn't compile on linux because of sagikazarmark/please-go-modules#14

PR #3187, opened 2020-09-25
feat(ListNP): fixed edge-case and error handling

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | resolves #3173 
| License         | Apache 2.0

### What's in this PR?

Added status to `ListNodePools` output and fixed node pool creation edge cases.

### Why?

Because there were issues with node pool listing when the CF stack creation failed, was not started yet or the stack was already deleted.

### Additional context

CLI [PR](https://github.com/banzaicloud/banzai-cli/pull/312).
Documentation [PR](https://github.com/banzaicloud/banzaicloud.github.io/pull/1854).

One issue was partially due to using stack name for description which fails in case of deleted stacks.

Another issue was that if the node pool creation validation failed, no stacks exists at that point, but the database has the node pool already.

Another implicit issue was that when the node pool create validation succeeds, but the CF stack create hasn't started yet, the node pool listing could still fail due to not existing stack.

4 approaches were considered for DB stack ID update:
1. add DB update activity around `CreateASGActivity` and return the stack ID from `CreateASGActivity`.
2. split `CreateASGActivity`
3. change `CreateASGActivity` to workflow with multiple subactivities.
4. add DB changing to CreateASGActivity (chosen)

The `CreateASGActivity` executes further actions after retrieving the stack ID from `CreateStack` and can fail later and also the Cadence activity output synchronization according to its documentation either returns an output value or an error and not both which means I could not return the stack ID and potential further errors in those cases.
I think the activity executes a logical unit and should not be split because of that.
The third approach would only make sense if we were restricting actions such as database modification to workflows and disallow those in activities, haven't experienced such distinction yet, but correct me if I'm wrong.

#### Test matrix (tested with PR CLI variant for simplicity)

```yaml
creationStates:
    - operations:
        - cluster create
        - cluster update (with new node pools)
        - node pool create
    - stateChangesToCheck:
        - Not existing -> Creating -> Error (pre-stack, validation failed)
        - Not existing -> Creating -> Ready
deletionStates:
    - operations:
        - node pool delete
        - cluster delete
    - stateChangesToCheck:
        - Error -> Deleting -> Deleted
        - Ready -> Deleting -> Deleted
```

<details>

  <summary>Cluster create, Not existing -> Creating -> Error</summary>

```yaml
# ~/banzai_cluster_create.yaml
name: pregnor-3173-error
location: eu-central-1
[...]

PR #3191, opened 2020-09-29
fix(API): error logging and error response handling

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Fixed 3 issues around API error response and its logging.

1. Implemented the `NotFound` error interface on the `SecretNotFoundError` service error type so the corresponding status problem matcher would catch it and turn it into a special (404) status error.
2. Implemented the `BadRequest` error interface on the `MismatchError` service error type so the corresponding status problem matcher would catch it and turn it into a special (400) status error.
3. Errors not having API level problem matchers (e.g. errors without dedicated response encoder) returned `500 Internal server error, something went wrong` responses, changed this to `500 Internal server error, <error.Error()>` in case the error implements the service error interface.

### Why?

During the review of #3187 a discussion emerged regarding errors during cluster creation while listing node pools which wasn't part of that PR's scope, but contained valid user expectations. This is the fix for the issue of proper API error response.

### Checklist

### To Do

- [x] Implement status problem matchers for known service errors along with the new matcher to handle service errors properly instead of 500 something went wrong. (See the first two commits, existing error matchers were used with interface implementations.)
- [x] Discuss how to proceed with the response encoder, error handler, error encoder behaviors. (We want to catch service errors in the response encoder's error encoding branch.)
- [x] Discuss whether we want to log 500 service errors or not-500 errors. (We are already logging them, but we explicitly avoid logging service errors.)

### DIFF
--- a/internal/common/secret.go
+++ b/internal/common/secret.go
@@ -51,6 +51,12 @@ func (e SecretNotFoundError) Details() []interface{} {
 	return []interface{}{"secretId", e.SecretID}
 }
 
+// NotFound tells a client that this error is related to a resource being not
+// found. Can be used to translate the error to eg. status code.
+func (SecretNotFoundError) NotFound() bool {
+	return true
+}
+
 // ServiceError tells the consumer whether this error is caused by invalid input supplied by the client.
 // Client errors are usually returned to the consumer without retrying the operation.
 func (SecretNotFoundError) ServiceError() bool {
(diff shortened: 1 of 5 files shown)


sagikazarmark (approved):
LGTM, thanks!

bonifaido (approved):
👍

PR #3201, opened 2020-10-06
Fix failing integration test

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     |yes
| Deprecations?   | yes

### What's in this PR?

The releaser integration test used invalid test data to manage releases. (app version instead of version)

### Why?
The releaser integration test failed


sagikazarmark (approved):
Thanks @lpuskas

PR #3202, opened 2020-10-06
Add isPassword field to secret-type response

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | partially https://github.com/banzaicloud/banzai-cli/issues/315

### What's in this PR?
Add isPassword field to secret-type response

### Checklist

### DIFF
--- a/.gen/pipeline/api/openapi.yaml
+++ b/.gen/pipeline/api/openapi.yaml
@@ -20883,9 +20883,11 @@ components:
       example:
         fields:
         - name: name
+          isSafeToDisplay: true
           description: description
           required: true
         - name: name
+          isSafeToDisplay: true
           description: description
           required: true
       properties:
@@ -24117,12 +24119,15 @@ components:
     SecretTypeResponse_fields:
       example:
         name: name
+        isSafeToDisplay: true
         description: description
         required: true
       properties:
         name:
           type: string
         required:
+          type: boolean
+        isSafeToDisplay:
           type: boolean
         description:
           type: string
--- a/.gen/pipeline/pipeline/model_secret_type_response_fields.go
+++ b/.gen/pipeline/pipeline/model_secret_type_response_fields.go
@@ -16,5 +16,7 @@ type SecretTypeResponseFields struct {
 
 	Required bool `json:"required,omitempty"`
 
+	IsSafeToDisplay bool `json:"isSafeToDisplay,omitempty"`
+
 	Description string `json:"description,omitempty"`
 }
--- a/internal/secret/secrettype/secret.go
+++ b/internal/secret/secrettype/secret.go
@@ -16,10 +16,11 @@ package secrettype
 
 // FieldMeta describes how a secret field should be validated
 type FieldMeta struct {
-	Name        string `json:"name"`
-	Required    bool   `json:"required"`
-	Opaque      bool   `json:"opaque,omitempty"`
-	Description string `json:"description,omitempty"`
+	Name            string `json:"name"`
+	Required        bool   `json:"required"`
+	Opaque          bool   `json:"opaque,omitempty"`
+	Description     string `json:"description,omitempty"`
+	IsSafeToDisplay bool   `json:"isSafeToDisplay,omitempty"`
 }
 
 // Meta describes how a secret is built up and how it should be sourced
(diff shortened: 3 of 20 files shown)


.gen/pipeline/pipeline/model_secret_type_response_fields.go
```diff
@@ -16,5 +16,7 @@ type SecretTypeResponseFields struct {
 
 	Required bool `json:"required,omitempty"`
 
+	IsPassword *bool `json:"isPassword,omitempty"`
```
pregnor, line 19: **Question**: What are we using the nullability for?
If it's unset then it's true, if its set to false then false, if it's set to true then true?

In this case I would propose trying to reverse the field and using it without nullability.
I'm trying to find a non-negated name for the reverse field, but failed so far.
If we can't I'm fine with this, just an idea.

In general I support sensitive default for false value and explicit turn-off.

(If I assumed it wrongly then please clarify it for me and let's talk about it.)

(If we change it, the references need to be updated as well.)

.gen/pipeline/api/openapi.yaml
```diff
@@ -24124,6 +24127,9 @@ components:
           type: string
         required:
           type: boolean
+        isPassword:
+          nullable: true
+          type: boolean
```
pregnor, line 24132: **Question**: what triggered these changes?
Have you edited the generated file instead of the manual one?

colin014: No, `generate-openapi` task (from Makefile) updated these files in `.gen` directory

.gen/pipeline/pipeline/model_secret_type_response_fields.go
pregnor: Checked the CLI change and if I understood correctly we are using nil value equals false in behavior so I'm confused why is it nullable.

.gen/pipeline/api/openapi.yaml
pregnor: I understood that, but what was the source.

Isn't this change supposed to be coming from apis/pipeline/pipeline.yaml only?

pregnor: I could find it in the `apis/pipeline/pipeline.yaml` now, so it's fixed, I must have missed it on first look I guess.

.gen/pipeline/pipeline/model_secret_type_response_fields.go
pregnor: We discussed in private my original assumption was right, the CLI code was mistaken.

internal/secret/secrettype/secret.go
```diff
@@ -266,40 +267,40 @@ var DefaultRules = map[string]Meta{
...
 		Fields: []FieldMeta{
-			{Name: OracleUserOCID, Required: true, Description: "Your Oracle user OCID. Find more about, generating public key and fingerprint here: https://banzaicloud.com/docs/pipeline/secrets/providers/oci_auth_credentials/"},
-			{Name: OracleTenancyOCID, Required: true, Description: "Your tenancy OCID"},
-			{Name: OracleAPIKey, Required: true, Description: "Your public key"},
-			{Name: OracleAPIKeyFingerprint, Required: true, Description: "Fingerprint of you public key"},
-			{Name: OracleRegion, Required: true, Description: "Oracle region"},
-			{Name: OracleCompartmentOCID, Required: true, Description: "Your compartment OCID"},
+			{Name: OracleUserOCID, Required: true, IsSafeToDisplay: true, Description: "Your Oracle user OCID. Find more about, generating public key and fingerprint here: https://banzaicloud.com/docs/pipeline/secrets/providers/oci_auth_credentials/"},
+			{Name: OracleTenancyOCID, Required: true, IsSafeToDisplay: true, Description: "Your tenancy OCID"},
+			{Name: OracleAPIKey, Required: true, IsSafeToDisplay: true, Description: "Your public key"},
```
pregnor, line 272: **Question**: are we sure this is a public key?
API keys are usually private that's why I'm asking, the name and the description are somewhat contradicting each other according to my understanding, but that's just my intuition.

Same [at](https://github.com/banzaicloud/pipeline/pull/3202/files#diff-3aae6f1133ed7ec044650bb3eb2efec4R44) `internal/secret/types/type_oracle.go:44`.

pregnor (approved):

.gen/pipeline/pipeline/model_secret_type_response_fields.go
pregnor: [Fixed](https://github.com/banzaicloud/pipeline/pull/3202/commits/fb1bf55e3d9609c04f7fca8eccb6cd573cbbe7bd).

internal/secret/secrettype/secret.go
colin014: Yes, this is a public key (https://banzaicloud.com/docs/pipeline/secrets/providers/oci_auth_credentials/#generate-api-token). In Oracle side we need to upload a public key to use this secret

pregnor: Cool, it's all good then.

PR #3207, opened 2020-10-08
Improve docker build

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets |

### What's in this PR?
Improve docker build
Push images to ghcr.io

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,13 @@
 name: CI
 
 on:
+  schedule:
+    - cron: '0 5 * * *' # everyday at 5 am UTC
   push:
     branches:
       - master
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
   pull_request:
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,24 +200,70 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout code
+      - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Determine tag
-        uses: haya14busa/action-cond@v1
-        id: imagetag
+      - name: Repo metadata
+        id: repo
+        uses: actions/github-script@v3
         with:
-          cond: ${{ github.event_name == 'pull_request' }}
-          if_true: ${{ github.sha }}
-          if_false: "master"
+          script: |
+            const repo = await github.repos.get(context.repo)
+            return repo.data
 
-      - name: Build
-        uses: docker/build-push-action@v1
+      - name: Docker build info
+        id: info
+        run: |
+          DOCKER_IMAGE=ghcr.io/banzaicloud/pipeline
+          VERSION=noop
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,9 +1,13 @@
 name: CI
 
 on:
+  schedule:
```
bonifaido, line 4: Why do we need this schedule?

.github/workflows/ci.yml
```diff
@@ -196,24 +200,70 @@ jobs:
...
+          elif [[ $GITHUB_REF == refs/heads/* ]]; then
+            VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
+            if [ "${{ github.event.repository.default_branch }}" = "$VERSION" ]; then
+              VERSION=edge
+            fi
+          elif [[ $GITHUB_REF == refs/pull/* ]]; then
+            VERSION=pr-${{ github.event.number }}
+          fi
+          TAGS="${DOCKER_IMAGE}:${VERSION}"
+          if [[ $VERSION =~ ^v?[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$ ]]; then
```
pregnor, line 232: **Question**: shouldn't we drop the `v` prefix as we are not using `v1.2.3` tags in this repository, but `1.2.3` instead?
Or are we planning to change the tag naming strategy in accordance to this change?

According to the tag matching expression [earlier](https://github.com/banzaicloud/pipeline/pull/3207/files#diff-e9f950f17198d3d5e3122a44230a09b9R10) I'm supposing this is just a copy-paste mistake.

pregnor, line 232: **Question**: why are we limiting the number of version digits to 3?
Realistically it should be enough (so this is not a hard change request), but there is always a possiblity to use dates as valid semver version which is not matched by this expression.

Also at the GHA [event trigger](https://github.com/banzaicloud/pipeline/pull/3207/files#diff-e9f950f17198d3d5e3122a44230a09b9R10) we defined it unbound to the number of digits.

.github/workflows/ci.yml
sagikazarmark: For nightly builds to detect flaky tests or issues caused by automaticly updated components.

.github/workflows/ci.yml
sagikazarmark: Note the questionmark after 'v'.  It makes the step more portable.

sagikazarmark: That's a copy-paste issue

pregnor: Yeah, I totally missed it when read, apologies.

sagikazarmark: No worries 🙂

sagikazarmark: Fixed

pregnor (approved):

bonifaido (approved):

PR #3213, opened 2020-10-12
Delete CRDs when deactivating the monitoring integrated service

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | yes

### What's in this PR?
* helm dependency changed to the "official" v3.3.4
* cleanupCRDs has been switched off for the monitoring IS
* CRDs are deleted from the cluster on deactivation 

### Why?
Monitoring Integrated service deactivation failed due to the related CRDs being deleted before CRs. To handle the problem a patched helm version was used.

### Checklist


pepov (approved):
LGTM although haven't tested yet. We should have tests around integrated services and some tests around the deployment API in general.

PR #3219, opened 2020-10-13
give better version number for builds from detached heads

orymate:
The version was empty if `make build-release` was run on an untagged detached head.

Examples:

```
0.60.0-21-gfac732148-fix-version-guessing
       # 21 commits after 0.60.0, 
           # commit fac732148,
                     # the current top of the fix-version-guessing branch
0.60.0-21-gfac732148-master
0.60.0-22-g8245381df-   # if not a tag and not the top of a branch (do we want to handle this?)
0.58.0-4-gbd36844e8-pke-log
0.60.999
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ OPENAPI_DESCRIPTOR = apis/pipeline/pipeline.yaml
 BUILD_DIR ?= build
 BUILD_PACKAGE = ${PACKAGE}/cmd/pipeline
 TEMPORARY_DIRECTORY = tmp
-VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || git symbolic-ref -q --short HEAD)
+VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || echo "`git describe --tags`-`git symbolic-ref -q --short HEAD 2>/dev/null || git describe --all --exact-match | sed 's!heads/!!'`")
 COMMIT_HASH ?= $(shell git rev-parse --short HEAD 2>/dev/null)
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Version probably doesn't make sense on a detached head. It should either be a tag or a branch name. I'm guessing the GitHub checkout does something which ends the source repo up in a detached state.

I'd expose a version variable instead and override it with the GITHUB_REF env var.

orymate:
> Images built from the master don't have an exact version. This is the direct consequence of not using tagged versions to trigger dev builds anymore.

That's why I committed this PR...

Makefile
```diff
@@ -13,7 +13,7 @@ OPENAPI_DESCRIPTOR = apis/pipeline/pipeline.yaml
 BUILD_DIR ?= build
 BUILD_PACKAGE = ${PACKAGE}/cmd/pipeline
 TEMPORARY_DIRECTORY = tmp
-VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || git symbolic-ref -q --short HEAD)
+VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || echo "`git describe --tags`-`git symbolic-ref -q --short HEAD 2>/dev/null || git describe --all --exact-match | sed 's!heads/!!'`")
```
sagikazarmark, line 16: Adding `${GITHUB_REF#refs/heads/}` temporarily should be enough for now. (Probably needs a build arg in the dockerfile too)

sagikazarmark (changes requested):

PR #3229, opened 2020-10-14
feat(PKE/AWS): NLB health check

### LINKED ISSUE #3211: Investigate and implement AWS specific health checks:
* all targets behind the nlb healthy?

pbalogh-sa:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3211 

### Why?
Implementation of AWS NLB targetgroup health check

### Checklist

### Example of activity usage
```
	if multiMaster {
		var healthCheclActivityOutput pkeawsworkflow.HealthCheckActivityOutput

		activityInput := pkeawsworkflow.HealthCheckActivityInput{
			OrganizationID: input.OrganizationID,
			SecretID:       input.SecretID,
			Region:         input.Region,
			ClusterID:      input.ClusterID,
			ClusterName:    input.ClusterName,
		}
		err = workflow.ExecuteActivity(ctx, pkeawsworkflow.HealthCheckActivityName, activityInput).Get(ctx, &healthCheclActivityOutput)
		if err != nil {
			return err
		}
	}
```


internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,143 @@
...
+		loadBalancerOutput, errInt = elbv2Svc.DescribeLoadBalancers(
+			&elbv2.DescribeLoadBalancersInput{
+				Names: []*string{&nlbName},
+			})
+		if errInt != nil {
+			return errors.WrapIf(errInt, "failed to describe load balancers")
+		}
+
+		return nil
+	}, backoffPolicy); err != nil {
```
pregnor, line 98: **Optional**: this call with the anonymous function definition would **MAY** be more readable preceding the if function instead of being inlined, because it does 3 things at the moment.

1. executes the backoff retry call.
2. defines the inlined anonymous function to describe the load balancers,
3. checks the error condition.

I would even consider extracting the anonymous function to a variable for readability purposes, but it is a secondary opinion.

Same at
https://github.com/banzaicloud/pipeline/pull/3229/files#diff-07dfcca59ca7ddb4d7bf740160fa2bdc2bc1668420e7549b77e0106c15a728e1R103
https://github.com/banzaicloud/pipeline/pull/3229/files#diff-07dfcca59ca7ddb4d7bf740160fa2bdc2bc1668420e7549b77e0106c15a728e1R119

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, err
+	}
+
+	var targetGroupOutput *elbv2.DescribeTargetGroupsOutput
+	if err := backoff.Retry(func() error {
+		var errInt error
+		targetGroupOutput, errInt = elbv2Svc.DescribeTargetGroups(
+			&elbv2.DescribeTargetGroupsInput{
+				LoadBalancerArn: loadBalancerOutput.LoadBalancers[0].LoadBalancerArn,
```
pregnor, line 107: **Required**: we **MUST** check the existence of the 0th value before trying to dereference it to avoid panic.

Same at
https://github.com/banzaicloud/pipeline/pull/3229/files#diff-07dfcca59ca7ddb4d7bf740160fa2bdc2bc1668420e7549b77e0106c15a728e1R123

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,143 @@
...
+		if errInt != nil {
+			return errors.WrapIf(errInt, "failed to describe target health")
+		}
+
+		return nil
+	}, backoffPolicy); err != nil {
+		return nil, err
+	}
+
+	for _, tgHealhtDesc := range targetHealthOutput.TargetHealthDescriptions {
```
pregnor, line 134: **Required**: typo `tgHealhtDesc` -> `tgHealthDesc` (check all references)

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,143 @@
...
+		if *tgHealhtDesc.TargetHealth.State != elbv2.TargetHealthStateEnumHealthy {
```
pregnor, line 135: **Recommended**: we **SHOULD** add a preceding `lgHealthDesc.TargetHealth.State != nil` condition to make sure we are not trying to dereference a nil pointer.

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,143 @@
...
+			return &HealthCheckActivityOutput{
+				nlbStatus: targetHealthOutput.TargetHealthDescriptions,
+			}, errors.New("not all targets of the network load balancer are healthy")
+		}
+	}
+
+	return nil, nil
```
pregnor, line 142: **Optional**: we **MAY** consider the more idiomatic approach of either returning a successful value with nil error or a nil value with an error, but never both.

Could we incorporate the returned value into the error content as details or would we loose any information then?
As I can see in success case we return <nil, nil>.

pregnor (changes requested):

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil
+	}, backoffPolicy); err != nil {
+		return nil, err
+	}
+
+	for _, tgHealhtDesc := range targetHealthOutput.TargetHealthDescriptions {
+		if *tgHealhtDesc.TargetHealth.State != elbv2.TargetHealthStateEnumHealthy {
+			return &HealthCheckActivityOutput{
+				nlbStatus: targetHealthOutput.TargetHealthDescriptions,
+			}, errors.New("not all targets of the network load balancer are healthy")
```
orymate, line 138: we may want to give some details (or at least the name) of the unhealthy target in the error

orymate (approved):

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed n https://github.com/banzaicloud/pipeline/pull/3229/commits/6146e54ad33a4aa0d22fad813f9c67841de394ea

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/6146e54ad33a4aa0d22fad813f9c67841de394ea

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/6146e54ad33a4aa0d22fad813f9c67841de394ea

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/6146e54ad33a4aa0d22fad813f9c67841de394ea

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/93d075890463a1f969c8f4469a4a31a713cebbd3

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/db2f970c859fd030d62d292d475e7b86164f0478

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/activity_nlb_health_check.go
pregnor: You only partially addressed what I commented.

In the error case we are still returning <non-nil-value, error message> which is not really idiomatic in Golang.
Is this intentional and necessary?

pregnor: This is also important, because I read it somewhere in the Cadence documentation/code that they propagate nil value in case of error so I believe the non-nil return value is thrown away anyway, trying to find the source of that information again.

Meaning these

```golang
			return &HealthCheckActivityOutput{
				nlbStatus: targetHealthOutput.TargetHealthDescriptions,
			}, errors.NewWithDetails("unable to getting state of the target health description", "targetHealthDesc", tgHealthDesc)
...
			return &HealthCheckActivityOutput{
				nlbStatus: targetHealthOutput.TargetHealthDescriptions,
			}, errors.NewWithDetails("not all targets of the network load balancer are healthy", "targetHealthDesc", tgHealthDesc)
```

Can't we do something like this:

```golang
return nil, errors.NewWithDetails("<message>", "targetHealthDescriptions", targetHealthOutput.TargetHealthDescriptions, "targetHealthDesc", tgHealthDesc)
```

pregnor: Nevermind, based on the [0.13.4 implementation](https://github.com/uber-go/cadence-client/blob/cadef60d6bd144a8b36c1c555bbf559273e9b494/internal/workflow.go#L462) it seems like the return values get propagated in case of error as well.

pbalogh-sa: fixed: https://github.com/banzaicloud/pipeline/pull/3229/commits/12c03d74786f9c865b56632b3481ce203a09b584

pregnor (approved):

bonifaido (approved):
👍

Issue #3197, opened 2020-10-05
split PKE distribution cloud flavours from clusteradapter.NewPKEService


bonifaido:
Originally from comment https://github.com/banzaicloud/pipeline/pull/3179#r498875727

> In terms of code structuring: will clusteradapter.NewPKEService be reusable for other cloud distributions?
> I think it would be better though (if possible at all) if we registered pke as a distribution, and inside the PKE service it can forward the request to the right cloud instance.
> Just to give you an idea, how this would look like:
```go
		"pke": clusteradapter.NewPKEService(
			map[string]pke.CloudDistribution{
				"aws": pkeaws.NewService(...),
			},
		),
```

PR #3233, opened 2020-10-16
Split PKE distribution cloud flavours from clusteradapter.NewPKEService

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | https://github.com/banzaicloud/pipeline/issues/3197

### What's in this PR:
A simplified solution for the given ticket. 

### Checklist

### TODO: 
- [ ] Is this simple code reorganisation is enough for now, or should we implement further changes like update code/logic to decide distribution in deeper parts and update all services according to this.

### DIFF
--- a/internal/cluster/service.go
+++ b/internal/cluster/service.go
@@ -355,8 +355,7 @@ func (NotSupportedDistributionError) ServiceError() bool {
 }
 
 func (s service) getDistributionService(cluster Cluster) (Service, error) {
-	key := cluster.Distribution
-	service, ok := s.distributions[key]
+	service, ok := s.distributions[cluster.Distribution]
 	if !ok {
 		return nil, errors.WithStack(NotSupportedDistributionError{
 			ID:           cluster.ID,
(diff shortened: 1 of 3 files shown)


internal/cluster/service.go
```diff
@@ -356,9 +356,6 @@ func (NotSupportedDistributionError) ServiceError() bool {
 
 func (s service) getDistributionService(cluster Cluster) (Service, error) {
 	key := cluster.Distribution
```
bonifaido, line 358: we can remove event this variable here (`key`).

sagikazarmark:
This is not really what I had in mind. This PR basically relocates some code, but relies on the same solution.

The issue describes the solution I imagined (with an example). It'll probably require a bit more refactoring.

sagikazarmark:
This isn't really what we wanted, so closing for now.

PR #3236, opened 2020-10-16
use latest velero chart & API and prefix to buckets for cluster backup

### LINKED ISSUE #3215: Support using the same backup bucket for multiple clusters
Support using the same backup bucket for multiple clusters using the clusterName (plus eventually cluster UUID) as a prefix to separate different clusters in bucket.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3215 

### What's in this PR?
- use latest Velero 1.5.1 API, library & helm chart modified slightly to support different cloud provider secrets at a time
- add prefix field to ark_backup_buckets table
- set cluster name as a prefix for buckets when creating backups

### Why?
To be able to support using same object store bucket for multiple clusters

### Checklist

- [X] User guide and development docs updated (if needed)
- [X] Related Helm chart(s) updated (if needed)

### Test Plan

This PR has been tested following these steps:

- create EKS/PKE AWS cluster
- `helm install wordpress  bitnami/wordpress`
- create a new blog record
- create backup to s3 new bucket
- check backup is created in bucket
- create new EKS cluster from the previously created backup
- check the previously create blog record is there, while Velero resources were deleted
- create backup to GS bucket, check backup dir is created on Google Console 
- delete backup, create backup to Azure Blob Storage, check backup is created on Azure Console

- create GKE cluster, create backup to GS and S3 then restore new cluster from backup
- create AKS/PKE Azure cluster, create backup to Azure Blob Storage and S3 then restore new cluster from backup

### Upgrade plan

Following the upgrade, Pipeline will attempt to sync the followings:

- buckets from objectstore: since Velero backup format is different then earlier Ark format, no backups will be found in storage buckets, all existing backups going to be deleted from Pipeline database.
- backup and restore resources from cluster: as the new custom resources group (velero.io) is different then the CRD group installed on clusters (ark.io) there will be no backup/restore found on cluster

In short: earlier backups will be deleted from Pipeline database (backup resources on cluster, backup files on object storage buckets stay untouched) scheduled backups are performed by Ark operator running on clusters, however users are not able to do backups/restores with Pipeline UI/CLI until backup is disabled and enabled again, installing the latest Velero operator.

Users can disable backup using banzai cli:

```
banzai cluster --cluster-name=CLUSTER_NAME service backup disable
```


sagikazarmark:
> Implementation tested (with at least one cloud provider)

I think we should extensively test the behavior of Pipeline and its managed clusters before and after applying this change. For example: enable backups (with schedule) using ark, checkout this change and see how Pipeline/the cluster behaves.

Obviously, they will be incompatible to some degree, but we need to make sure that this change doesn't cause any crashes. I'd even consider running it for a few days with a daily schedule (or hourly if available) to make sure upgrading Pipeline, but not upgrading ark doesn't cause any fatal errors.

sancyx:
> > Implementation tested (with at least one cloud provider)
> 
> I think we should extensively test the behavior of Pipeline and its managed clusters before and after applying this change. For example: enable backups (with schedule) using ark, checkout this change and see how Pipeline/the cluster behaves.
> 
> Obviously, they will be incompatible to some degree, but we need to make sure that this change doesn't cause any crashes. I'd even consider running it for a few days with a daily schedule (or hourly if available) to make sure upgrading Pipeline, but not upgrading ark doesn't cause any fatal errors.

Added the `Upgrade plan` above

internal/ark/api/backups.go
```diff
@@ -19,7 +19,7 @@ import (
 	"strconv"
```
pregnor, line 19: Is there any plan, to update the package reference from ark to velero in a future PR?

pregnor (commented):
As I lack a lot of contextual information I could not review the semantics properly, so just went through it as a glorified code analyzer, commented what I had found.

internal/ark/config.go
```diff
@@ -272,9 +363,20 @@ func (req ConfigRequest) getCredentials() (credentials, error) {
...
+	switch pullPolicy {
+	case string(v1.PullAlways):
+		return v1.PullAlways
+
+	case string(v1.PullNever):
+		return v1.PullNever
+	default:
+		return v1.PullIfNotPresent
+	}
+}
```
pregnor, line 382: **Optional**: opinionated, but this **MAY** be easier to read/maintain the following way:

```suggestion
func getPullPolicy(pullPolicy string) v1.PullPolicy {
	switch pullPolicy {
	case string(v1.PullAlways), string(v1.PullIfNotPresent), string(v1.PullNever): // Note: known values.
		return v1.PullPolicy(pullPolicy)
	default:
		return v1.PullIfNotPresent
	}
}
```

internal/ark/api/backups.go
sancyx: Not sure, may be we can address that when we create Velero integrated service.

internal/cmd/config.go
```diff
@@ -747,18 +762,33 @@ traefik:
...
-	v.SetDefault("cluster::disasterRecovery::namespace", "")
+	v.SetDefault("cluster::disasterRecovery::namespace", "pipeline-system")
 	v.SetDefault("cluster::disasterRecovery::ark::syncEnabled", true)
 	v.SetDefault("cluster::disasterRecovery::ark::bucketSyncInterval", "10m")
 	v.SetDefault("cluster::disasterRecovery::ark::restoreSyncInterval", "20s")
 	v.SetDefault("cluster::disasterRecovery::ark::backupSyncInterval", "20s")
 	v.SetDefault("cluster::disasterRecovery::ark::restoreWaitTimeout", "5m")
-	v.SetDefault("cluster::disasterRecovery::charts::ark::chart", "banzaicloud-stable/ark")
-	v.SetDefault("cluster::disasterRecovery::charts::ark::version", "1.2.3")
+	v.SetDefault("cluster::disasterRecovery::charts::ark::chart", "banzaicloud-stable/velero")
```
sagikazarmark, line 771: We should probably rename at least the chart to velero....in a subsequent PR. Please open an issue to track it.

internal/cmd/config.go
```diff
@@ -747,18 +762,33 @@ traefik:
...
+	v.SetDefault("cluster::disasterRecovery::charts::ark::version", "2.13.3-bnz1")
```
sagikazarmark, line 772: These version numbers kill me. We have `bnz`, `bzc` and god knows what else. It should either be something like `2.13.3-bc.1` or `2.13.3-banzaicloud.1` or just `2.13.3.1`.

Please come up with a better versioning strategy.

sagikazarmark (approved):

internal/cmd/config.go
sancyx: https://github.com/banzaicloud/pipeline/issues/3270

PR #3240, opened 2020-10-17
Improve docker build

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
This PR adds a number of improvements to the Docker image build job:

- image layer caching
- skip logging into registries when not necessary
- remove unused qemu step (copy&paste error)

### Why?
Caching might speed up the Docker image build (time will tell).

Removing unnecessary steps from the workflow might have the same effect.

### Additional context
The cache size might not fit into the [limits](https://github.com/actions/cache#cache-limits)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,9 +235,6 @@ jobs:
           echo ::set-output name=commit_hash::${GITHUB_SHA::8}
           echo ::set-output name=build_date::$(git show -s --format=%cI)
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -266,8 +266,8 @@ jobs:
         with:
           push: ${{ github.event_name == 'push' }}
           tags: ${{ steps.tags.outputs.tags }}
-          cache-from: type=local,src=/tmp/.buildx-cache
-          cache-to: type=local,dest=/tmp/.buildx-cache
+          cache-from: type=local,mode=max,src=/tmp/.buildx-cache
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -247,22 +252,24 @@ jobs:
...
           username: ${{ github.repository_owner }}
           password: ${{ secrets.CR_PAT }}
+        if: ${{ github.event_name == 'push' }}
 
       - name: Login to Docker Hub
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
+        if: ${{ github.event_name == 'push' }}
```
pregnor, line 262: **Question**: why don't we need to log in when we are pushing a tag image for example?

Dockerfile
```diff
@@ -20,8 +20,8 @@ COPY pkg/sdk/go.* /build/pkg/sdk/
 RUN go mod download
 
 ARG VERSION
-ARG COMMIT_HASH
-ARG BUILD_DATE
+ARG COMMIT_HASHH
+ARG BUILD_DATEE
```
pregnor, line 24: **Question**: what's the reasoning behind the deliberate misspelling? I guess it clashes with some other names, but I might propsose adding a `DOCKER_ARG_` prefix or something similar, or maybe updating the other reference.

.github/workflows/ci.yml
sagikazarmark: We only push images on a push event.

Dockerfile
sagikazarmark: I was testing something, but I reverted the change since then.

bonifaido (approved):

PR #3242, opened 2020-10-20
Pke nodepool delete

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #3251 

### What's in this PR?
Node pool delete logic implementation for pke

### Checklist


cmd/worker/pke.go
```diff
@@ -17,15 +17,44 @@ package main
...
+	// node pool delete helper activities
+	deleteStackActivity := awsworkflow.NewDeleteStackActivity(awsSessionFactory)
+	activity.RegisterWithOptions(deleteStackActivity.Execute, activity.RegisterOptions{Name: awsworkflow.DeleteStackActivityName})
+
+	deleteStoredNodePoolActivity := workflow.NewDeleteStoredNodePoolActivity(nodePoolStore)
+	activity.RegisterWithOptions(deleteStoredNodePoolActivity.Execute, activity.RegisterOptions{
+		Name: workflow.DeleteStoredNodePoolActivityName,
+	})
+
+	listStoredNodePoolsActivity := workflow.NewListStoredNodePoolsActivity(nodePoolStore)
```
sagikazarmark, line 49: Is this activity called from anywhere? If not, drop it.

internal/cluster/distribution/pke/node_pool.go
```diff
@@ -0,0 +1,40 @@
...
+// +testify:mock
+
+// NodePoolStore provides an interface for PKE node pool persistence.
+type NodePoolStore interface {
+	// DeleteNodePool deletes an existing node pool from the storage.
+	DeleteNodePool(ctx context.Context, organizationID, clusterID uint, clusterName string, nodePoolName string) error
+
+	// ListNodePools retrieves the node pools for the cluster specified by its
+	// cluster ID.
+	ListNodePools(
```
sagikazarmark, line 34: As far as I can tell, this only exists (together with `ExistingNodePool`), because the list activity calls it. But I don't think the list activity is used anywhere, so this should be dropped with it.

internal/cluster/distribution/pke/pkeaws/pkeawsworkflow/common.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // TODO: this is temporary
-func setClusterStatus(ctx workflow.Context, clusterID uint, status, statusMessage string) error {
+func SetClusterStatus(ctx workflow.Context, clusterID uint, status, statusMessage string) error {
```
sagikazarmark, line 26: I believe we discussed this should probably not be exported.

internal/cluster/distribution/pke/service.go
```diff
@@ -120,12 +124,18 @@ type service struct {
...
+	nodePoolStore   NodePoolStore
 }
 
 // +testify:mock:testOnly=true
 
 // NodePoolManager is responsible for managing node pools.
 type NodePoolManager interface {
+	// DeleteNodePool deletes an existing node pool in a cluster.
+	DeleteNodePool(
+		ctx context.Context, c cluster.Cluster, existingNodePool ExistingNodePool, shouldUpdateClusterStatus bool,
```
sagikazarmark, line 136: What kind of formatting is this? If the line is too long, break every parameter into a new line.

sagikazarmark (changes requested):
I can't see any tests covering this change. 😢

internal/cluster/distribution/pke/service.go
pregnor: I believe this is on me.

I have been experimenting with a 3 step function signature formatting approach in some of my previous PRs where it was approved so I highlighted this possibility of formatting for Jani.

(In order of precedence.)

1. **Single line formatting**: If the function signature fits into 120 characters, then it should be formatted to a single line.
```golang
func (receiver receiverType) functionName(parameter1 parameterType1, parameter2 parameterType2) (returnValue1 returnType1, returnValue2 returnType2)
```

2. **Single line parameter list**: If the function signature doesn't fit into 120 characters, but the function parameter list fits into 120 characters, then the function name, the parameter list and the return values should be formatted into their own dedicated lines.
```golang
func (receiver receiverType) functionName(
    parameter1 parameterType1, parameter2 parameterType2, parameter3 parameterType3, parameter4 parameterType4, parameter5 parameterType5,
) (returnValue1 returnType1, returnValue2 returnType2)
```

3. **One parameter per line**: If the function parameter list doesn't fit into 120 characters, then the function name, each individual parameter and the return values should be formatted into their own dedicated lines.
```golang
func (receiver receiverType) functionName(
    parameter1 parameterType1,
    parameter2 parameterType2, 
    parameter3 parameterType3, 
    parameter4 parameterType4, 
    parameter5 parameterType5,
) (returnValue1 returnType1, returnValue2 returnType2)
```

The reason why I introduced the single-line-parameter-list formatting is because the one-parameter-per-line formatting occasionally in minimally arrow-head like situations with long function names takes up so much vertical space with very short arguments, it makes the code more difficult to read IMO, although this is debatable.

Examples:
```golang
// Original.
func Name(c cluster.Cluster) {
    // ...
    if condition {
        // ...
        err := operation()
        if err != nil {
            err = methodWithAVeryLongQualifiedName(many, arguments, with, short, names, barely, over, the, line, length, limits)

            return err
        }
        // ...
    }
    // ....
}
// single-line-parameter-list.
func Name(c cluster.Cluster) {
    // ...
    if condition {
        // ...
        err := operation()
        if err != nil {
            err = methodWithAVeryLongQualifiedName(
                many, arguments, with, short, names, barely, over, the, line, length, limits
            )

            return err
        }
        // ...
    }
    // ...
}
// one-parameter-per-line.
func Name(c cluster.Cluster) {
    // ...
    if condition {
        // ...
        err := operation()
        if err != nil {
            err = methodWithAVeryLongQualifiedName(
                many, 
                arguments, 
                with, 
                short, 
                names, 
                barely, 
                over, 
                the, 
                line, 
                length, 
                limits,
            )

            return err
        }
        // ...
    }
    // ...
}
```

I apologize for the confusion, I should have highlighted the single-line-parameter-list formatting in the PR's I experimented with to make sure it is not overlooked.

I'm fine with dropping the single-line-parameter-list formatting and reverting back to the single-line or one-parameter-per-line 2 step function signature formatting approach.
Although I still feel like in some edge cases the single-line-parameter-list formatting approach is preferable, I can accept it being opinionated and thus not favorable.

internal/cluster/distribution/pke/node_pool.go
janosSarusiKis: It's called from the service as well.In the delete [function](https://github.com/banzaicloud/pipeline/pull/3242/files#diff-d746cfc80d0e748007eb394ac902c14cedde894d4b3e652c2e7dcc10d787290aR163)

internal/cluster/distribution/pke/service.go
pregnor: I can correct these in a future PR, it's not difficult to find them by regular expression.

internal/cluster/distribution/pke/service.go
```diff
@@ -120,17 +124,28 @@ type service struct {
...
 
-	// ListNodePools lists node pools from a cluster.
-	ListNodePools(ctx context.Context, c cluster.Cluster, nodePoolNames []string) ([]NodePool, error)
+	// UpdateNodePool updates an existing node pool in a cluster.
+	UpdateNodePool(
+		ctx context.Context,
+		c cluster.Cluster,
+		nodePoolName string,
+		nodePoolUpdate NodePoolUpdate,
+	) (string, error)
```
pregnor, line 148: **Question**: why is the `NodePoolManager.UpdateNodePool` addition part of this PR?
Does `DeleteNodePool` use it anywhere (it should not IMO)?

pregnor (commented):
There are no glaring issues to fix, so it is approve-ready once I get answers for my questions.

internal/cluster/distribution/pke/node_pool.go
```diff
@@ -0,0 +1,40 @@
...
+
+package pke
+
+import (
+	"context"
+)
+
+type ExistingNodePool struct {
+	Name string
+}
```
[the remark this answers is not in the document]
pregnor: Also the stored footprint of a node pool is possibly going to be different for different providers under PKE providers so it might be a good idea to define this in the common PKE part as an interface type so each provider can supply its own stored node pool implementation.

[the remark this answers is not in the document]
pregnor: This might be true for other types which are aimed to be used as generic PKE types for various PKE providers.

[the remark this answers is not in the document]
pregnor: Or alternatively define this storage specific struct in the provider specific PKE package.

internal/cluster/distribution/pke/node_pool.go
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pke
+
+import (
+	"context"
+)
+
+type ExistingNodePool struct {
```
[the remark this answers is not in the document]
janosSarusiKis: Ok

internal/cluster/distribution/pke/service.go
janosSarusiKis: It's not an addition, I change the order and formatted the code.

internal/cluster/distribution/pke/node_pool.go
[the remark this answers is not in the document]
janosSarusiKis: Making an interface and provider specific structs will be created as part of future implementations.

pregnor (approved):

PR #3278, opened 2020-11-02
Cluster backup: add useClusterSecret option to enable backup request

### LINKED ISSUE #3260: Cluster backup: add cluster access option to cluster backup
Add a new option to Cluster Backup enable where user can choose between granting access to volume snapshots by adding required permissions to NodeInstanceRole or deploying Pipeline cluster secret to cluster.

https://github.com/orgs/banzaicloud/projects/47#card-48160745

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3260 

### What's in this PR?
Add a new property useClusterSecret to Cluster Backup enable request which enables user  to choose weather he wants  to deploy Pipeline cluster secret to cluster or not. In this latter case Velero operator will fall back to use the NodeInstanceProfile so the user has to add the below permissions to NodeInstanceRole:

```
ec2:DescribeVolumes
ec2:DescribeSnapshots
ec2:CreateTags
ec2:CreateVolume
ec2:CreateSnapshot
ec2:DeleteSnapshot
```

EKS: permissions has to be appended to the NodeInstanceRole referred by nodeInstanceRoleId specified in EKS cluster create request. Note that these permissions are not appended automatically to the default NodeInstanceRole created by Pipeline.

PKE on AWS: permissions has to be appended to pke-global-worker NodeInstanceRole as Pipeline is not appending these permissions by default.

### Checklist

### DIFF
--- a/.gen/pipeline/api/openapi.yaml
+++ b/.gen/pipeline/api/openapi.yaml
@@ -22558,7 +22558,11 @@ components:
       example:
         cloud: google
         bucketName: ark-backup-bucket-123
+        resourceGroup: my-resource-group
+        prefix: my-backup-prefix
         secretId: 61b9707ca463cad79188bb0ddfa1345e41fc01b651feddf06eea58a25c56d717
+        location: us-east
+        storageAccount: my-storage-account
       properties:
         cloud:
           example: google
@@ -22568,6 +22572,21 @@ components:
           type: string
         secretId:
           example: 61b9707ca463cad79188bb0ddfa1345e41fc01b651feddf06eea58a25c56d717
+          type: string
+        prefix:
+          description: Optional, if not provided will be set to the name of the cluster
+          example: my-backup-prefix
+          type: string
+        location:
+          example: us-east
+          type: string
+        storageAccount:
+          description: required only case of Azure
+          example: my-storage-account
+          type: string
+        resourceGroup:
+          description: required only case of Azure
+          example: my-resource-group
           type: string
       required:
       - bucketName
@@ -22690,6 +22709,24 @@ components:
           type: string
         secretId:
           example: 62bc3c75-91fb-4670-bad4-24b401a9deac
+          type: string
+        location:
+          example: us-east
+          type: string
+        useClusterSecret:
+          description: relevant only in csse of Amazon clusters. By default set to
+            false in which case you must add snapshot permissions to your instance
+            profile. Should you set to true Pipeline will deploy your cluster secret
+            to the cluster.
+          example: false
+          type: boolean
+        storageAccount:
+          description: required only case of Azure
+          example: my-storage-account
+          type: string
+        resourceGroup:
+          description: required only case of Azure
+          example: my-resource-group
           type: string
         labels:
           $ref: '#/components/schemas/Labels'
--- a/.gen/pipeline/api/openapi.yaml
+++ b/.gen/pipeline/api/openapi.yaml
@@ -22581,7 +22581,7 @@ components:
           example: us-east
           type: string
         storageAccount:
-          description: required only case of Azure
+          description: required only in case of Azure
           example: my-storage-account
           type: string
         resourceGroup:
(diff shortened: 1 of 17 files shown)


pepov:
If I understand right right now it's not possible to work without a cluster secret, but I don't see `UseClusterSecret` set to true by default anywhere which means that this PR aims the change the default behaviour?

sancyx:
> If I understand correctly right now it's not possible to work without a cluster secret, but I don't see `UseClusterSecret` set to true by default anywhere which means that this PR aims the change the default behaviour?

The actual behaviour is that in case you have different cluster and bucket secrets we only set the bucket secret so in case this doesn't have volume snapshot permissions the backup will partially fail. With this PR you can choose between deploying your cluster secret or add volume snapshot permissions to NodeInstanceProfile yourself. By default we don't deploy the cluster secret so we presume user has granted perm. to NodeInstanceProfile.

pregnor (commented):
LGTM, still not confident in my contextual knowledge, don't forget to update the permission and usage documentation accordingly.

apis/pipeline/pipeline.yaml
```diff
@@ -8057,6 +8072,21 @@ components:
...
+                    example: false
+                    description: "relevant only in case of Amazon clusters. By default set to false in which case you must add snapshot permissions to your instance profile. Should you set to true Pipeline will deploy your cluster secret to the cluster."
+                storageAccount:
+                    type: string
+                    example: "my-storage-account"
+                    description: "required only case of Azure"
+                resourceGroup:
+                    type: string
+                    example: "my-resource-group"
+                    description: "required only case of Azure"
```
pregnor, line 8089: **Note**: these may be a little cleaner to be defined in a oneof structure where the Amazon and EKS only values are defined in their respective schemas.

This approach may be worth considering for `CreateBackupBucketRequest` as well.

apis/pipeline/pipeline.yaml
```diff
@@ -7954,6 +7954,21 @@ components:
...
+                    type: string
+                    example: "my-backup-prefix"
+                    description: "Optional, if not provided will be set to the name of the cluster"
+                location:
+                    type: string
+                    example: "us-east"
+                storageAccount:
+                    type: string
+                    example: "my-storage-account"
+                    description: "required only case of Azure"
```
pregnor, line 7967: **Note**: typo.
```suggestion
                    description: "required only in case of Azure"
```

Similarly at:
https://github.com/banzaicloud/pipeline/pull/3278/files#diff-d41e335500d6bf91169555b68a970b732b21f70dbecef99585ec49d11f7758b7R7971
https://github.com/banzaicloud/pipeline/pull/3278/files#diff-d41e335500d6bf91169555b68a970b732b21f70dbecef99585ec49d11f7758b7R8085
https://github.com/banzaicloud/pipeline/pull/3278/files#diff-d41e335500d6bf91169555b68a970b732b21f70dbecef99585ec49d11f7758b7R8089

.gen/pipeline/api/openapi.yaml
```diff
@@ -22691,6 +22710,24 @@ components:
         secretId:
           example: 62bc3c75-91fb-4670-bad4-24b401a9deac
           type: string
+        location:
+          example: us-east
+          type: string
+        useClusterSecret:
+          description: relevant only in case of Amazon clusters. By default set to
```
sagikazarmark, line 22717: s/instance profile/node instance role/

internal/ark/config.go
```diff
@@ -321,9 +324,14 @@ func (req ConfigRequest) getCredentials() (credentials, error) {
 
 	switch req.Cluster.Provider {
 	case providers.Amazon:
-		ClusterSecretContents, err = amazon.GetSecret(req.ClusterSecret)
+		// In case of Amazon we set up one credential file with different profiles for cluster & bucket secret.
+		// If UseClusterSecret is false there's no need for cluster secret user will make sure NodeInstanceProfile has the right permissions
```
sagikazarmark, line 328: node instance role

sagikazarmark (changes requested):

sagikazarmark (approved):

PR #3281, opened 2020-11-04
fix(#3280/AWS/DeleteStack): fixed existence check

### LINKED ISSUE #3280: AWS failed to create node pool CloudFormation stacks cannot be deleted.
**Describe the bug**

When an AWS node pool creation fails after the corresponding entry is stored, but before the node pool stack is created, node pool deletion fails with non-existing stacks, because the DescribeStack call which would check whether deletion is required fails.

1. Create an EKS or PKE/AWS cluster.
2. Try creating a node pool that is going to fail before the stack is created (I tested it with a transient development state so it cannot be reproduced that way, maybe try using subnets which aren't part of the cluster subnet list, that should fail the creation before stack creation is initiated).
3. Try deleting the failed to create node pool.
4. Observe the cluster run into warning state because the node pool cannot be deleted, because the stack cannot be found (actually at the deletion required check).

**Expected behavior**

The node pool should successfully delete itself from the storage.

**Screenshots**

**Additional context**

Catch the non-existent stack error during the delete stack activity's delete required check and return gracefully if no corresponding stack is found.

pregnor:
Handled pre-creation failure non-existent CloudFormation stack properly.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | resolves #3280 

### What's in this PR?

Fixed the handling of non-existent CloudFormation stack handling in `AWSCommon.DeleteStackActivity` by adding the corresponding validation error to the non-existent stack error handling branch on the existence check.

### Why?

Because there are different errors for already deleted and never created stacks, only the former was handled before.

### Additional context

Unfortunately I couldn't find any predefined error code/message for this case in https://github.com/aws/aws-sdk-go.

### Checklist

### DIFF
--- a/internal/cluster/infrastructure/aws/awsworkflow/delete_stack_activity.go
+++ b/internal/cluster/infrastructure/aws/awsworkflow/delete_stack_activity.go
@@ -16,6 +16,7 @@ package awsworkflow
 
 import (
 	"context"
+	"fmt"
 
 	"emperror.dev/errors"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


bonifaido (approved):

sagikazarmark (approved):
It's probably fine for now, but I guess in the future the describe stack step should be a separate activity from the Cadence SDK call, right?

pregnor:
@sagikazarmark I think so, yeah, other activities are using it inline as well, so it might be worth extracting it, but it would require a somewhat bigger refactor, because right now the activities are calling the describe and then it would have to be put up onto the workflow level and the activities should get the stacks through their input arguments.

PR #3286, opened 2020-11-05
objectstore_test: retry the most flaky part

orymate:
### What's in this PR?
My short research about failing integration test runs showed that this single API call fails most of the time. This may be caused by the kind of eventual consistency of bucket creation.

### Why?
No idea about other low effort solution.

### DIFF
--- a/pkg/providers/amazon/objectstore/objectstore_test.go
+++ b/pkg/providers/amazon/objectstore/objectstore_test.go
@@ -340,7 +340,12 @@ func testObjectStoreDelete(t *testing.T) {
 
 	err = s.DeleteBucket(bucketName)
 	if err != nil {
-		t.Fatal("could not test bucket deletion: ", err.Error())
+		// this test seems to be the most flaky one, give it another chance
+		time.Sleep(time.Second)
+		err = s.DeleteBucket(bucketName)
+		if err != nil {
+			t.Fatal("could not test bucket deletion: ", err.Error())
+		}
 	}
 
 	head := &s3.HeadBucketInput{


pkg/providers/amazon/objectstore/objectstore_test.go
```diff
@@ -340,7 +340,12 @@ func testObjectStoreDelete(t *testing.T) {
 
 	err = s.DeleteBucket(bucketName)
 	if err != nil {
-		t.Fatal("could not test bucket deletion: ", err.Error())
+		// this test seems to be the most flaky one, give it another chance
+		time.Sleep(time.Second)
+		err = s.DeleteBucket(bucketName)
+		if err != nil {
+			t.Fatal("could not test bucket deletion: ", err.Error())
+		}
```
pregnor, line 348: **Note**: as much as I love the effort put into this (it's one of the first things I put on a todo list of mine, because it bugs me to hell), I think it would be better if we found a more suitable solution to the test flakyness, because I really dislike introducing artificial delays into tests as a workaround for outside conditions, because A, 1s here and there can quickly add up to a significant amount of time and B, we are hiding behavior behind an opaque cloak of "time passing".

What my proposal would be here is that we should introduce a piece of logic inbetween the `CreateBucket` and `DeleteBucket` calls which waits for the bucket to be created, because IMO the issue comes from the creation not finishing and thus the deletion cannot start/finish.

My assumptions are based on the error message I saved months ago to debug this when I have the time: `TestIntegration/ObjectStore_Delete: objectstore_test.go:342: could not test bucket deletion:  failed to wait for bucket to be deleted: ResourceNotReady: exceeded wait attempts`
Correct me if I'm wrong and the resource creates successfully and only the deletion requires some delay.

So if there is a low-effort wait method available to check for the resource being ready, I would rather opt for that approach.

I'm not going to veto this short term solution, just wanted to express my thoughts.

orymate: I'd normally agree with this, but 1. this is a test that runs only if explicitly selected by the integration test CI run 2. I wouldn't invest more here, because this approach was what prevented us in the previous 2 years to do this quick workaround. We can still do this better, if this delay doesn't help in most cases, which I hope it does.

sagikazarmark (approved):
TBH, I don't think we should run these tests in an automated way. This code probably won't ever change again and at some point, we will get rid of it (hopefully). Therefore I don't care what happens to it.

pkg/providers/amazon/objectstore/objectstore_test.go
pregnor: I'm concerned only because I feel like we are introducing this delay but we cannot properly verify whether it works at all and thus we don't know how long it's going to be required, in case of a refactor this could easily be dropped without verifying whether it is still required.

pregnor: I mean `ResourceNotReady` seems to be a state we could easily check and add a workaround resource waiting for that.

Instead of saying stop the system for 1 second and hope it is resolved outside.

pregnor: So I would rather try the narrower workaround first and the broader workaround later, that's what I meant to say, but it's up to you.

pregnor: Can this test be run locally?

If yes, I can set up a scenario where I run this 1000s of times to see if it fails somewhat reliably locally and test my narrower workaround later.

pregnor: (And then I can gladly let this PR go in.)

orymate: I've seen two different error messages in the last ~10 failures. Anyway, this is one of the parallel tests in a CI step that typically lasts more than 10 minutes. In the rare (~10%) case when this step fails, we can add another second + the time api call takes, to prevent a rerun, which is more than 10 min. This seems to be a win in this situation. The test itself isn't really useful, and we did know this in the last 2 years too, but no change has been made. The only time this difference you suggest counts is when the api call fails with a different error, which we consider to be a real error, so we won't rerun the test. I don't think that this will ever happen.

pregnor: All right.

PR #3288, opened 2020-11-05
fix(Token): fixed cluster token validation

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Fixed the JWT token storing and thus validation with the new go-jose JWT implementation.

### Why?

The cluster tokens are non-expiring, but they were stored with an explicit expiration set to 0 explicit time (0001-01-01T00:00:00Z), because at one point where we converted between non-pointer to pointer expiration the address of the zero value time was taken instead of setting expiration to nil.

Example: (token saved with no expiration/zero time value)
```json
{
  "token": {
    "expiresAt": "0001-01-01T00:00:00Z",
    "id": "c81ee9e4-e925-40b6-8034-9e8b7d0b79b8",
    "name": "clusters/1/14",
    "value": "eyJhbGciOiJIUzI1NiJ9.eyJhdWQiOlsiaHR0cHM6Ly9waXBlbGluZS5iYW56YWljbG91ZC5jb20iXSwiaWF0IjoxNjA0NTk4NzcxLCJpc3MiOiJodHRwczovL2JhbnphaWNsb3VkLmNvbS8iLCJqdGkiOiJjODFlZTllNC1lOTI1LTQwYjYtODAzNC05ZThiN2QwYjc5YjgiLCJzY29wZSI6ImFwaTppbnZva2UiLCJzdWIiOiJjbHVzdGVycy8xLzE0IiwidGV4dCI6ImNsdXN0ZXJzLzEvMTQiLCJ0eXBlIjoiY2x1c3RlciJ9.you_thought_so"
  }
}
```

### Additional context

Among long term design ideas up for debate:

1. we should use time pointers everywhere or no time pointers at all, but converting between values and pointers is always erroneous, especially when the zero value has a special meaning and not equals the semantic of any other value.
2. we should consider updating the Vault auth lookup token parsing to handle zero time value as well as nil value (it's a pointer there), because I think this is a somewhat "common" serialization issue.

### Checklist

- ~Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- ~Related Helm chart(s) updated (if needed)

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -188,14 +188,16 @@ func (m TokenManager) GenerateToken(
 	tokenText string,
 	tokenName string,
 	storeSecret bool,
-) (string, string, error) {
-	tokenID, signedToken, err := m.generator.GenerateToken(sub, expiresAt, string(tokenType), tokenText)
+) (tokenID string, signedToken string, err error) {
+	tokenID, signedToken, err = m.generator.GenerateToken(sub, expiresAt, string(tokenType), tokenText)
 	if err != nil {
 		return "", "", err
 	}
 
 	token := auth.NewToken(tokenID, tokenName)
-	token.ExpiresAt = &expiresAt
+	if expiresAt != NoExpiration {
+		token.ExpiresAt = &expiresAt
+	}
 
 	if storeSecret {
 		token.Value = signedToken
--- a/pkg/auth/token_test.go
+++ b/pkg/auth/token_test.go
@@ -18,6 +18,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


orymate (approved):

bonifaido (approved):
Nice catch!

pregnor:
I'm going to add some unit tests while I'm waiting for the CI, because the current ones didn't catch this issue, even though they easily should have.

sagikazarmark (approved):
Thanks!

PR #3289, opened 2020-11-05
Refactor Cadence worker registration

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?
Use the worker specific registration API instead of the global registry.

### Why?
Makes testing and workflow/activity registration easier.


pkg/cadence/worker/registry.go
```diff
@@ -0,0 +1,105 @@
...
+	// To override each name of activities defined through a structure register the methods one by one:
+	// activities := &Activities{ ... }
+	// worker.RegisterActivityWithOptions(activities.SampleActivity1, RegisterActivityOptions{Name: "Sample1"})
+	// worker.RegisterActivityWithOptions(activities.SampleActivity2, RegisterActivityOptions{Name: "Sample2"})
+	// See RegisterActivity function for more info.
+	// The other use of options is to disable duplicated activity registration check
+	// which might be useful for integration tests.
+	// worker.RegisterActivityWithOptions(barActivity, RegisterActivityOptions{DisableAlreadyRegisteredCheck: true})
+	RegisterActivityWithOptions(a interface{}, options activity.RegisterOptions)
+}
```
pregnor, line 105: **Question**: shouldn't we put this in `pkg/sdk/cadence/worker/registry.go` so we could depend on it from the enterprise repository?

pregnor (approved):

pkg/cadence/worker/registry.go
sagikazarmark: Could be, but it's not part of the SDK per se, so I'm fine with copying it until it's fixed upstream.

PR #3290, opened 2020-11-06
refactor(EKS/LegacyClusterAPI.CreateCluster): used CreateNodePoolWorkflow for node pool creation

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Refactored the legacy EKS `ClusterAPI.CreateCluster` (and subsequently EKS `CreateInfrastrcutureWorkflow)` implementations to use the recently introduced `CreateNodePoolWorkflow` for node pool creation. 

### Why?

in order to move towards a unified node pool creation solution, deduplicate business logic code and simplify expectations.

### Additional context

Advised to review by commit.

Another PR is coming where the legacy EKS `ClusterAPI.UpdateCluster` implementation is refactored accordingly, but it requires a bit more refactoring due to me missing a chance of refactor when the `DeleteNodePoolWorkflow` was introduced, so I decided to split the two apart into their own PRs for agility reasons.

### Checklist

### To Do

- [x] Manual test.


bonifaido (approved):

orymate (approved):

sancyx (approved):
lgtm

pregnor:
I have added a [commit](https://github.com/banzaicloud/pipeline/pull/3290/commits/c16014d89ba63cd7a0c907b1d25b99d0ef04befc) making label set creation optional in the `CreateNodePoolWorkflow`, because the order of execution in a `ClusterAPI.CreateCluster` call understandably causes the node pool creation to run before the node pool label set operator is installed.
So label set creation should be turned off for cluster creation cases and on for other cases including NodePoolAPI.CreateNodePool.

In case of cluster creation, the cluster creation workflow makes sure the node pool labels are synced to the existing node pools once the operator is installed, so no additional operations are required then.

pregnor:
Had to fix the node instance role handling making sure to always use the resource ID only, stripping the paths specified for explicit node instance role ID cluster creation.

PR #3293, opened 2020-11-06
Introduce integrated-service-sdk module

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Adds a separate modul that will be used to interact with the Integrated Service Operator.


lpuskas (approved):
lgtm

lib/integrated-service-sdk/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
pregnor, line 1: Just curious: what are the pros (and cons) of defining this as an inlined Golang module of the Pipeline repository (with its own build infrastructure, license, CI) as opposed to putting it into its own repository?

I believe I can see the possible pros of the separate repository approach, but I'm failing to grasp the cons.
It could also be git-submoduled under Pipeline if that is preferred over go modding it.
(This might be simple difference in perspectives - for example I'm preferring modular approach over monolithic structures - I'm just trying to understand the reasoning behind the current solution for now.)

Are there going to be any possible consumers other than the Pipeline?

lib/integrated-service-sdk/.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
+name: Check
```
pregnor, line 1: Is there any reason behind using `lib/integrated-service-sdk` location for this Go module as opposed to putting it into `pkg/sdk/integrated-service` similarly to other SDKs?

pregnor (commented):
LGTM.

PR #3308, opened 2020-11-12
Fix platform helm repo addition logic

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Fixed (platform) helm repo addition logic in case of cluster federation

### Why?
The helm repo addition logic was wrong (only in case of cluster federation) as it 
only checked the repository URL

### DIFF
--- a/internal/helm/helmadapter/helm_service.go
+++ b/internal/helm/helmadapter/helm_service.go
@@ -256,12 +256,25 @@ func (h *helm3UnifiedReleaser) AddRepositoryIfNotExists(repository helm.Reposito
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


sagikazarmark (changes requested):
Please submit the org ID refactor and the actually fix in separate commits to make the review easier.

orymate (approved):

pepov (approved):

sagikazarmark (approved):

lpuskas:
> Please submit the org ID refactor and the actually fix in separate commits to make the review easier.

fixed, thx

PR #3310, opened 2020-11-13
Integrated Service SDK: Change availableVersions to contain upgrade path

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | yes
| Deprecations?   | no

### What's in this PR?
Changes availableVersions structure to contain the shortest upgrade path to specific target versions.
Get rid of availableTransitions.


lpuskas (approved):
lgtm

PR #3319, opened 2020-11-17
Integrated service testing

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

These tests aim to verify that Integrated Service API specifications are met.

Considering efficiency as a key aspect, tests are executed against the highest logical, but still internal layer of the Pipeline Integrated Service API, which means that the http machinery is completely bypassed.

There is no user authentication, users and organizations are represented by fake entities.

Although the Pipeline Web - and any UI component - is not required for these tests to run certain dependencies are:
MySQL, Vault, Cadence as external dependencies (launched using docker-compose for example).

A running Pipeline Worker configured with the same external dependencies is also required, but to make debugging easier we don't make an assumption on how the worker is started. It is recommended to run the worker with the same codebase using the same config: testconfig/config.yaml

Example how to trigger this using make:

```
make test-integrated-service-up
make test-integrated-service-worker &
pid=$!
make test-integrated-service
kill $pid
make test-integrated-service-down
```

### Caveats

The test currently uses a docker-compose setup that is slightly different from the one used for development but conflicts with it at certain points:
- they use the same local bind mounts, thus they use the same database and vault secret keys
- they use the same ports for services

To avoid interfering with a running development environment `make test-integrated-service-up` first stops development docker-compose services, but we may eventually want a better/more isolated testing environment.

### DIFF
--- a/.github/workflows/is-test.yml
+++ b/.github/workflows/is-test.yml
@@ -23,6 +23,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Go build cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/go-build
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/*.go', '**/go.mod', '**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-build-
+
       - name: Create kind cluster and run tests
         run: |
           make test-integrated-service-up
--- a/.github/workflows/is-test.yml
+++ b/.github/workflows/is-test.yml
@@ -1,0 +1,33 @@
+name: Integrated Service test
+
+on:
+  push:
+    branches:
+      - master
+      - integrated-service-testing
+  pull_request:
+
+jobs:
+  is-test:
+    runs-on: ubuntu-latest
+    env:
+      VERBOSE: 1
+      GOFLAGS: -mod=readonly
+
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Create kind cluster and run tests
+        run: |
+          make test-integrated-service-up
+          make build-worker
+          make test-integrated-service-worker &
+          pid=$!
+          make test-integrated-service
+          kill $pid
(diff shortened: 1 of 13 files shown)


orymate (approved):

.github/workflows/is-test.yml
```diff
@@ -0,0 +1,41 @@
+name: Integrated Service test
+
+on:
+  push:
+    branches:
+      - master
+      - integrated-service-testing
```
sagikazarmark, line 7: This should be removed before merging.

.github/workflows/is-test.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Go build cache
```
sagikazarmark, line 26: Haven't you concluded that it's faster without cache?

internal/integratedservices/functional_helpers_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	err = cluster.CreateCluster()
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
+	}
```
sagikazarmark, line 94: Go error handling is not good, but it's not this terrible either. 😄

internal/integratedservices/testconfig/docker-compose.yml
```diff
@@ -0,0 +1,93 @@
...
+            - 127.0.0.1:7935:7935
+
+#    cadence-web:
+#        image: ubercadence/web:3.18.1
+#        environment:
+#            CADENCE_TCHANNEL_PEERS: cadence:7933
+#        depends_on:
+#            - cadence
+#        ports:
+#            - 127.0.0.1:8088:8088
```
sagikazarmark, line 93: This could be moved to an override dist file, just like in the repo root. The override file should be ignored.

sagikazarmark (approved):

.github/workflows/is-test.yml
pepov: faster without go mod cache, but not without build cache

internal/integratedservices/testconfig/docker-compose.yml
pepov: It won't work since this file is specified with `-f` (running from project root) thus override files are not processed automatically.

PR #3321, opened 2020-11-17
fix(template/EKS/NP): fixed parameter metadata

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Fixed the EKS node pool CloudFormation template's metadata parameter listing regarding the newly introduced security groups.

### Why?

Because the parameter was renamed, but the template was not consolidated.

### Additional context

This doesn't affect the behavior of the template, it only affects the listing of the resulting parameters on the stack.

#3300 depends on this

### Checklist

### DIFF
--- a/templates/eks/amazon-eks-nodepool-cf.yaml
+++ b/templates/eks/amazon-eks-nodepool-cf.yaml
@@ -12,7 +12,7 @@ Metadata:
           - ClusterName
           - ClusterControlPlaneSecurityGroup
           - NodeSecurityGroup
-          - CustomSecurityGroup
+          - SecurityGroups
       -
         Label:
           default: "Worker Node Configuration"


templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -12,7 +12,7 @@ Metadata:
           - ClusterName
           - ClusterControlPlaneSecurityGroup
           - NodeSecurityGroup
-          - CustomSecurityGroup
+          - SecurityGroups
```
sagikazarmark, line 15: Isn't it called CustomSecurityGroups?

pregnor: In the latest update (#3315) it was renamed to `SecurityGroups`, that's where the discrepancy was introduced which I'm trying to fix here.

sancyx (approved):

templates/eks/amazon-eks-nodepool-cf.yaml
pregnor: My2cents are to merge this and if we want to change the name again, let's do that in a different PR.

PR #3328, opened 2020-11-18
Add securityGroups input param to node pool update

### LINKED ISSUE #3324: Add securityGroups input param to node pool update
Add securityGroups input param to node pool update.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3324 

### What's in this PR?
Add securityGroups input param to node pool update.
Tested with suggested scenarios by @pregnor.

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -62,6 +62,7 @@ type UpdateNodeGroupActivityInput struct {
 	NodeVolumeSize  int
 	NodeImage       string
 	DesiredCapacity int64
+	SecurityGroups  []string
 
 	MaxBatchSize          int
 	MinInstancesInService int
@@ -167,6 +168,10 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 		{
 			ParameterKey:     aws.String("NodeSecurityGroup"),
 			UsePreviousValue: aws.Bool(true),
+		},
+		{
+			ParameterKey:   aws.String("CustomNodeSecurityGroups"),
+			ParameterValue: aws.String(strings.Join(input.SecurityGroups, ",")),
 		},
 		{
 			ParameterKey:     aws.String("VpcId"),
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -169,10 +169,11 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			ParameterKey:     aws.String("NodeSecurityGroup"),
 			UsePreviousValue: aws.Bool(true),
 		},
-		{
-			ParameterKey:   aws.String("CustomNodeSecurityGroups"),
-			ParameterValue: aws.String(strings.Join(input.SecurityGroups, ",")),
-		},
+		sdkCloudFormation.NewOptionalStackParameter(
+			"CustomNodeSecurityGroups",
+			input.SecurityGroups != nil,
+			strings.Join(input.SecurityGroups, ","),
+		),
 		{
 			ParameterKey:     aws.String("VpcId"),
 			UsePreviousValue: aws.Bool(true),
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -120,6 +120,11 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			input.NodeImage != "",
 			input.NodeImage,
 		),
+		sdkCloudFormation.NewOptionalStackParameter(
+			"CustomNodeSecurityGroups",
+			input.SecurityGroups != nil || input.TemplateVersion == "1.0.0",
+			strings.Join(input.SecurityGroups, ","),
+		),
 		{
 			ParameterKey:     aws.String("NodeInstanceType"),
 			UsePreviousValue: aws.Bool(true),
@@ -195,14 +200,6 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			ParameterKey:   aws.String("BootstrapArguments"),
 			ParameterValue: aws.String(fmt.Sprintf("--kubelet-extra-args '--node-labels %v'", strings.Join(nodeLabels, ","))),
 		},
-	}
-
-	if input.TemplateVersion != "1.0.0" {
-		stackParams = append(stackParams, sdkCloudFormation.NewOptionalStackParameter(
-			"CustomNodeSecurityGroups",
-			input.SecurityGroups != nil,
-			strings.Join(input.SecurityGroups, ","),
-		))
 	}
 
 	// we don't reuse the creation time template, since it may have changed
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -68,6 +68,8 @@ type UpdateNodeGroupActivityInput struct {
 	MinInstancesInService int
 
 	ClusterTags map[string]string
+
+	TemplateVersion string
 }
 
 type UpdateNodeGroupActivityOutput struct {
@@ -169,11 +171,6 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			ParameterKey:     aws.String("NodeSecurityGroup"),
 			UsePreviousValue: aws.Bool(true),
 		},
-		sdkCloudFormation.NewOptionalStackParameter(
-			"CustomNodeSecurityGroups",
-			input.SecurityGroups != nil,
-			strings.Join(input.SecurityGroups, ","),
-		),
 		{
 			ParameterKey:     aws.String("VpcId"),
 			UsePreviousValue: aws.Bool(true),
@@ -198,6 +195,14 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			ParameterKey:   aws.String("BootstrapArguments"),
 			ParameterValue: aws.String(fmt.Sprintf("--kubelet-extra-args '--node-labels %v'", strings.Join(nodeLabels, ","))),
 		},
+	}
+
+	if input.TemplateVersion != "1.0.0" {
+		stackParams = append(stackParams, sdkCloudFormation.NewOptionalStackParameter(
+			"CustomNodeSecurityGroups",
+			input.SecurityGroups != nil,
+			strings.Join(input.SecurityGroups, ","),
+		))
 	}
 
 	// we don't reuse the creation time template, since it may have changed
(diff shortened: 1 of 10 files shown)


internal/cluster/distribution/eks/service.go
```diff
@@ -65,7 +65,8 @@ type ClusterUpdate struct {
 type NodePoolUpdate struct {
 	VolumeSize int `mapstructure:"volumeSize"`
 
-	Image string `mapstructure:"image"`
+	Image          string   `mapstructure:"image"`
+	SecurityGroups []string `securityGroups:"image"`
```
pregnor, line 69: **Required**: this tag **MUST** be fixed.

```suggestion
	SecurityGroups []string `mapstructure:"securityGroups"`
```

internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
```diff
@@ -168,6 +169,10 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 			ParameterKey:     aws.String("NodeSecurityGroup"),
 			UsePreviousValue: aws.Bool(true),
 		},
+		{
+			ParameterKey:   aws.String("CustomNodeSecurityGroups"),
+			ParameterValue: aws.String(strings.Join(input.SecurityGroups, ",")),
+		},
```
pregnor, line 175: **Required**: we **MUST** be able to handle user requests where the custom node security groups are meant to keep their values and not be updated.

For example when the user only wants to modify the image ID or the volume size they don't have to specify all the fields of `eks.NodePoolUpdate` in the update request, they can use a minimal update descriptor YAML like the following:
```yaml
image: "image-id-value"
```
or
```yaml
volumeSize: 21
```

AFAIK we would like to support the same optional behavior with `securityGroups` thus the stack update parameters should behave accordingly.
So the examples above should not modify the custom node security groups, those values should only be modified by explicit update requests containing the `securityGroups` field. 
You can use the available optional stack parameter constructor to achieve this.

(For the record: I would check for nil value as condition for using the existing stack parameter value and let the empty value unset the security groups while non-empty value sets them.)

```suggestion
		sdkCloudFormation.NewOptionalStackParameter(
			"CustomNodeSecurityGroups",
			input.SecurityGroups != nil,
			strings.Join(input.SecurityGroups, ","),
		),
```

Please also manually test this case thoroughly, especially the behaviors on unspecified `securityGroups`, no-item specified `securityGroups` and multiple item specified `securityGroups`.

pregnor (changes requested):

pregnor (dismissed):

pregnor (changes requested):
If I'm not mistaken, the node pool version calculation **MUST** also be updated accordingly. It should incorporate all values which can be changed through `EKS.NodePoolAPI.UpdateNodePool` requests. 

Currently it contains image and volume ID. 

Also to calculate it properly in case of nil `securityGroups` specification in the update, the corresponding original/effective value must be [retrieved](https://github.com/banzaicloud/pipeline/blob/master/internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go#L130) and [passed](https://github.com/banzaicloud/pipeline/blob/master/internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go#L203) to the version calculation.

Please test this PR with the following node pool update scenarios:

I.
1. no previous custom security groups,
2. not specified any security groups,
3. no other changes,
=> no node pool change, no custom security groups present.

II.
1. no previous custom security groups, 
2. not specified any security groups,
3. specified new volume size or image ID,
=> node pool updated, no custom security groups present.

III.
1. no previous custom security groups, 
2. specified new security groups,
3. no other changes,
=> node pool updated, new security groups present.

IV.
1. existing previous custom security groups, 
2. not specified any security groups,
3. no other changes,
=> no node pool change, old security groups present.

IV.
1. existing previous custom security groups, 
2. specified the same (old) security groups,
3. no other changes,
=> no node pool change, no security groups present.

V.
1. existing previous custom security groups, 
2. specified new security groups,
3. no other changes,
=> node pool updated, new security groups present (old ones are not).

pregnor:
Oh and I forgot another thing: due to the `CustomNodeSecurityGroups` parameter is only available in newly created node pool stacks (template version >= 2.0.0) the parameter retrieval must check the template version before trying to retrieve the new parameter to make sure it works for stacks created from older templates.

This indicates more test cases where you create a cluster and node pool using a node pool CloudFormation template without this parameter and executing the following scenarios on an "old" node pool.

I.
1. stack created from old template, no previous custom security groups,
2. not specified any security groups,
3. no other changes,
    => no node pool change, no custom security groups present, no parameter present.

II.
1. stack created from old template, no previous custom security groups,
2. not specified any security groups,
3. specified new volume size or image ID,
    => node pool updated, no custom security groups present, parameter is present.

III.
1. stack created from old template, no previous custom security groups,
2. specified new security groups,
3. no other changes,
    => node pool updated, new security groups present.

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
```diff
@@ -130,6 +132,7 @@ func (w UpdateNodePoolWorkflow) Execute(ctx workflow.Context, input UpdateNodePo
 
 	effectiveImage := input.NodeImage
 	effectiveVolumeSize := input.NodeVolumeSize
+	var effectiveSecurityGroups []string
 	if effectiveImage == "" ||
 		effectiveVolumeSize == 0 { // Note: needing CF stack for original information for version.
```
pregnor, line 137: **Required**: the `effectiveSecurityGroups` **MUST** take the value of the passed in security group list by default and only retrieve the original value when no new value is specified.

```
effectiveSecurityGroups := input.SecurityGroups
if effectiveImage == "" ||
    effectiveVolumeSize == 0 ||
    effectiveSecurityGroups == nil { // Note: needing CF stack for original information for version.
```
(cannot put suggestion on this block properly)

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
```diff
@@ -147,6 +150,8 @@ func (w UpdateNodePoolWorkflow) Execute(ctx workflow.Context, input UpdateNodePo
 		var parameters struct {
 			NodeImageID    string `mapstructure:"NodeImageId"`
 			NodeVolumeSize int    `mapstructure:"NodeVolumeSize"`
+			// Note: CustomNodeSecurityGroups is from template version 2.0.0
+			CustomNodeSecurityGroups string `mapstructure:"CustomNodeSecurityGroups,omitempty"`
```
pregnor, line 154: **Note**: that's a clever workaround I missed. 👍🏻

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
```diff
@@ -160,6 +165,13 @@ func (w UpdateNodePoolWorkflow) Execute(ctx workflow.Context, input UpdateNodePo
 		if effectiveVolumeSize == 0 {
 			effectiveVolumeSize = parameters.NodeVolumeSize
 		}
+
+		securityGroups := parameters.CustomNodeSecurityGroups
```
pregnor, line 169: **Note**: I think this var could be avoided now that a strongly typed field is available.

internal/cluster/distribution/eks/service.go
```diff
@@ -144,6 +146,9 @@ func NewNodePoolFromCFStack(name string, labels map[string]string, stack *cloudf
 	nodePool.SubnetID = nodePoolParameters.Subnets // Note: currently we ensure a single value at creation.
 	nodePool.Status = NewNodePoolStatusFromCFStackStatus(aws.StringValue(stack.StackStatus))
 	nodePool.StatusMessage = aws.StringValue(stack.StackStatusReason)
+	if nodePoolParameters.CustomNodeSecurityGroups != "" {
+		nodePool.SecurityGroups = strings.Split(nodePoolParameters.CustomNodeSecurityGroups, ",")
+	}
```
pregnor, line 151: **Note**: we are going to need this change for #3334 , but I think it should be a separate PR, it's going to affect the API as well, regenerated code, etc.

pregnor (changes requested):

internal/cluster/distribution/eks/service.go
pregnor: Implemented it [here](https://github.com/banzaicloud/pipeline/pull/3337).

sancyx: fixed

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
sancyx: fixed

internal/cluster/distribution/eks/eksworkflow/workflow_update_node_pool.go
sancyx: fixed

internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
```diff
@@ -194,6 +197,14 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 		},
 	}
 
+	if input.TemplateVersion != "1.0.0" {
+		stackParams = append(stackParams, sdkCloudFormation.NewOptionalStackParameter(
+			"CustomNodeSecurityGroups",
+			input.SecurityGroups != nil,
+			strings.Join(input.SecurityGroups, ","),
+		))
+	}
```
pregnor, line 206: **Required**: we also **MUST** provide the value if explicitly set, even if the stack is older than the introduction, so there should be an else branch for specified argument with explicit stack parameter setting.

```golang
} else if input.SecurityGroups != nil {
    stackParams = append(stackParamse, &cloudformation.Parmeter{
        ParameterKey: "CustomNodeSecurityGroups",
        ParameterValue: strings.Join(input.SecurityGroups, ","),
    })
}
```

Also as an idea, I would think about whether `NewOptionalStackParameter` can be extended with a version restriction to only use `UsePreviousValue` if a version requirement is met, because this feels to be a common case moving forward with versioned templates.

internal/cluster/distribution/eks/service.go
```diff
@@ -121,6 +123,7 @@ func NewNodePoolFromCFStack(name string, labels map[string]string, stack *cloudf
 		NodeSpotPrice               string `mapstructure:"NodeSpotPrice"`
 		NodeVolumeSize              int    `mapstructure:"NodeVolumeSize"`
 		Subnets                     string `mapstructure:"Subnets"`
+		CustomNodeSecurityGroups    string `mapstructure:"CustomNodeSecurityGroups,omitempty"`
```
pregnor, line 126: **Note**: listing requires other changes as well (and should be in a separate PR), I have submitted the corresponding change set [here](https://github.com/banzaicloud/pipeline/pull/3337).

pregnor (changes requested):

internal/cluster/distribution/eks/service.go
pregnor: Same at https://github.com/banzaicloud/pipeline/pull/3328/files#diff-07fa2a5916b261c493d7d45fe3fc856a5203f30b8219443b62b5c08f41839382R99

pregnor: Fixed

internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
pregnor: Fixed

pregnor (approved):

PR #3339, opened 2020-11-20
Integrated Service v2 for External DNS

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The initial version 2 of the integrated service framework (w/ external dns in focus)

### Why?
To implement a different/better approach for handling integrated services

### Additional context
This is a dev version only

### Checklist

### To Do
- [x] cleanup / refactor organize new components (wired values, logging, etc ...)
- [x] more testing


cmd/worker/clusterfeature.go
```diff
@@ -23,8 +23,13 @@ import (
 	clusterfeatureworkflow "github.com/banzaicloud/pipeline/internal/integratedservices/integratedserviceadapter/workflow"
 )
 
-func registerClusterFeatureWorkflows(worker worker.Worker, featureOperatorRegistry integratedservices.IntegratedServiceOperatorRegistry, featureRepository integratedservices.IntegratedServiceRepository) {
-	worker.RegisterWorkflowWithOptions(clusterfeatureworkflow.IntegratedServiceJobWorkflow, workflow.RegisterOptions{Name: clusterfeatureworkflow.IntegratedServiceJobWorkflowName})
+func registerClusterFeatureWorkflows(worker worker.Worker, featureOperatorRegistry integratedservices.IntegratedServiceOperatorRegistry, featureRepository integratedservices.IntegratedServiceRepository, isV2 bool) {
+	workflowFunc := clusterfeatureworkflow.IntegratedServiceJobWorkflow
+	if isV2 {
```
sagikazarmark, line 28: Workflow functions could be registered with different names. It would only make sense if the workflows are started by different callers though. If not, I guess this is fine.

go.mod
```diff
@@ -159,3 +161,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.19.3
 	k8s.io/sample-controller => k8s.io/sample-controller v0.19.3
 )
+
+replace github.com/banzaicloud/integrated-service-sdk => ./lib/integrated-service-sdk
```
sagikazarmark, line 165: This should be moved to the replace block above.

sagikazarmark (approved):

cmd/worker/clusterfeature.go
pepov: Right now the caller is the same, but we will change this in a follow up PR because v1/v2 workflows should be available for the v1/v2 apis respectively when we add the router layer that can dynamically dispatch to both.

go.mod
lpuskas: thx, fixed

pepov (approved):

PR #3340, opened 2020-11-23
Cadence client error check and default config for the pipeline process

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Bail out in case the cadence client cannot be created, otherwise execution will fail with a nil pointer dereference.

Also add default cadence configuration for the pipeline process, otherwise only the worker will be able to create the `pipeline` domain if `createNonexistentDomain` is enabled. This behaviour, that the pipeline process can also create the domain if it doesn't exist was introduced in https://github.com/banzaicloud/pipeline/pull/3319

### DIFF
--- a/cmd/pipeline/config.go
+++ b/cmd/pipeline/config.go
@@ -181,4 +181,8 @@ func configure(v *viper.Viper, p *pflag.FlagSet) {
 
 	v.SetDefault("spotmetrics::enabled", false)
 	v.SetDefault("spotmetrics::collectionInterval", 30*time.Second)
+
+	// Cadence configuration
+	v.SetDefault("cadence::createNonexistentDomain", false)
+	v.SetDefault("cadence::workflowExecutionRetentionPeriodInDays", 3)
 }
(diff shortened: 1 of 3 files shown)


cmd/pipeline/config.go
```diff
@@ -181,4 +181,8 @@ func configure(v *viper.Viper, p *pflag.FlagSet) {
 
 	v.SetDefault("spotmetrics::enabled", false)
 	v.SetDefault("spotmetrics::collectionInterval", 30*time.Second)
+
+	// Cadence configuration
+	v.SetDefault("cadence::createNonexistentDomain", false)
+	v.SetDefault("cadence::workflowExecutionRetentionPeriodInDays", 3)
```
sagikazarmark, line 187: Why is this added here? AFAIK the worker is responsible for creating the domain.

pepov: It's in the PR description, but additional info is that it was introduced in #3319 because without this I was unable to bootstrap a test where the worker starts independently from the pipeline process, and the pipeline process always started sooner. If for some reason (concurrency?) we don't want this, then I think I can add this to the test setup only.

sagikazarmark: I don't really like the idea TBH. The fact that Pipeline (worker) creates the domain for itself is bad to begin with, but now the API will try that, even in production. But this is probably the easier solution for now. If you can avoid it in the future, great, but let's not block this one for now.

sagikazarmark (approved):

PR #3341, opened 2020-11-24
Add Integrated Service Operator update workflow

colin014:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Add new workflow to check integrated servive operator. For now we just check the CRD is installed or not. In the future we want to update this operator for the new version

### Checklist

### TODO
- [ ] Add new flag to turf off this workflow

### DIFF
--- a/src/cluster/operator_update_workflow.go
+++ b/src/cluster/operator_update_workflow.go
@@ -1,0 +1,128 @@
+// Copyright © 2019 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cluster
+
+import (
+	"context"
+	"time"
+
+	"emperror.dev/errors"
+	"go.uber.org/cadence/workflow"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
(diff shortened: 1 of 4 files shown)


lpuskas (approved):

src/cluster/operator_update_workflow.go
```diff
@@ -0,0 +1,128 @@
...
+func (w IntServiceOperatorUpdaterWorkflow) Execute(ctx workflow.Context, input IntServiceOperatorUpdaterActivityInput) error {
+	logger := workflow.GetLogger(ctx).Sugar().With(
+		"clusterID", input.ClusterID,
+	)
+
+	commonCluster, err := w.manager.GetClusterByIDOnly(context.Background(), input.ClusterID)
+	if err != nil {
+		logger.Errorf("failed to get cluster from database: %s", err.Error())
+		return err
+	}
```
pepov, line 59: We discussed this, so just a reminder that we need the check if the cluster exists and result cleanly in case it's gone.

In case there's an error we don't want to quit the loop.

src/cluster/operator_update_workflow.go
```diff
@@ -0,0 +1,128 @@
...
+			logger.Errorf("failed to check integrated service operator: %s", err.Error())
+		} else {
+			logger.Info("integrated service operator CRD installed")
+		}
+	}
+
+	err = workflow.Sleep(ctx, 1*time.Minute)
+	if err != nil {
+		return errors.WrapIf(err, "sleep cancelled")
+	}
```
pepov, line 81: don't return here either, continue the loop

pepov (changes requested):

src/cluster/operator_update_workflow.go
colin014: I fixed this

src/cluster/operator_update_workflow.go
colin014: I fixed this

sagikazarmark (changes requested):
WIP

colin014:
We discussed and we will solve this in another way. I will create a new PR soon

PR #3355, opened 2020-12-02
Decouple Integrated Service v2 deployment details from Pipeline, implemented for DNS

lpuskas:
Extend the ServiceInstance custom resource with the ServiceSpec field to decouple deployment details from Pipeline. 

DNS is the first Integrated Service that implements this and leverages the operator to assemble Helm chart values, which is now completely transparent to Pipeline.

The main challange that had to be solved was how secretID should be handled. The way it currently works is that clients select a secret and put the secretID into the spec which is known by Pipeline. In case of the Integrated Service operator however we need references to actual Kubernetes secrets. To solve this Pipeline transforms secret IDs to secret names and installs these secrets to the managed cluster. The spec is transformed back to replace secret names to their original secret ids before returned through the GetDetails endpoint.

Additionally this PR adds a mechanism to avoid overwriting `ServiceInstance` resources managed externally. `ServiceInstance` resources managed by Pipeline will have the annotation: `app.kubernetes.io/managed-by: banzaicloud.io/pipeline`


sagikazarmark (approved):
I haven't reviewed it in too much details. For example, I'm not 100% what the spec mapper abstraction does and whether it's necessary.

pepov:
The spec mapper is something that allows hooking in when any part of the spec needs transformation. In our case we need to transform secret names back to secret ids for the DNS service, because those are transformed when set in the spec for the operator:
https://github.com/banzaicloud/pipeline/pull/3355/files#diff-9a8e0aa36714ec61145da8b1c6ccb0192c1676f89cb04754fb88d0d4d84b0bc5R134

Issue #3347, opened 2020-11-30
Add support for EKS cluster node pool worker node EBS volume encryption state change


pregnor:
**Is your feature request related to a problem? Please describe.**

Add support of changing the encryption state of an EBS volume of a worker node in an EKS cluster node pool.

**Describe the solution you'd like to see**

The encryption state (encryption flag and possibly encryption key) should be added to the API's cluster and node pool representation and should be modifiable through node pool updates (and cluster updates).
This requires modification to the node pool CloudFormation template as well, the EBS volume description should be extended with the encryption details.
The encryption information should be propagated from the API level to the CloudFormation template level.

**Describe alternatives you've considered**

-

**Additional context**

The encryption key's special policy [requirements](https://docs.aws.amazon.com/autoscaling/ec2/userguide/key-policy-requirements-EBS-encryption.html) should be highlighted on release.

PR #3357, opened 2020-12-03
Fixed an edge case bug + Refactored the EKS node pool CloudFormation template version handling

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Fixed a bug which caused the current template version not being retrieved for the node pool stacks which are updated when all node pool version attributes were changed during the update due to a left in condition for old value retrieval.

Refactored the EKS node pool CloudFormation template version handling.

1. Simplified `TemplateVersion` handling and make it easier to use/update.
Single point to track the latest node pool version as the default value of the `TemplateVersion` parameter in the template itself, keeping the update requirement local when the template changes, so no manual setting/update is required from code to keep all new stacks up to date with the latest template version.
2. Changed the `TemplateVersion` handling in the Golang code to use a strongly typed semantic version representation with helper functions instead of a "weakly typed string".
3. Refactored the CloudFormation stack parameter value parsing to handle the strongly typed semantic version and also tuned it a little bit to make full use of `mapstructure` and avoid double direction decoding/parsing (this resulted in 1.5 function being dropped completely at the expense of implementing a proxy function interface).

### Why?

This is useful to ease the implementation of #3347, because it requires another template version update (new parameters being introduced) so I fine-tuned the template version handling for that purpose.

### Additional context

Tested using `NodePoolAPI.UpdateNodePool` with both enterprise and open-source Pipeline implementation due to the condition and template version retrieval change.
Double checked Template version, it updates as expected.

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -28,7 +28,6 @@ import (
 	"go.uber.org/cadence/activity"
 
 	"github.com/banzaicloud/pipeline/internal/cluster"
-	"github.com/banzaicloud/pipeline/internal/cluster/distribution/eks/eksprovider/workflow"
 	"github.com/banzaicloud/pipeline/pkg/cadence/worker"
 	pkgCloudFormation "github.com/banzaicloud/pipeline/pkg/providers/amazon/cloudformation"
 	sdkAmazon "github.com/banzaicloud/pipeline/pkg/sdk/providers/amazon"
@@ -107,10 +106,6 @@ func (a UpdateNodeGroupActivity) Execute(ctx context.Context, input UpdateNodeGr
 	}
 
 	stackParams := []*cloudformation.Parameter{
-		{
-			ParameterKey:   aws.String(workflow.GetStackTemplateVersionKey()),
-			ParameterValue: aws.String("2.0.0"),
-		},
 		{
 			ParameterKey:     aws.String("KeyName"),
 			UsePreviousValue: aws.Bool(true),
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
@@ -32,6 +32,7 @@ import (
 	pkgCloudFormation "github.com/banzaicloud/pipeline/pkg/providers/amazon/cloudformation"
 	sdkAmazon "github.com/banzaicloud/pipeline/pkg/sdk/providers/amazon"
 	sdkCloudFormation "github.com/banzaicloud/pipeline/pkg/sdk/providers/amazon/cloudformation"
+	"github.com/banzaicloud/pipeline/pkg/sdk/semver"
 )
 
 const awsNoUpdatesError = "No updates are to be performed."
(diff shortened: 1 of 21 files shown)


pregnor:
I'm going to hotfix 0.63.1 with this change if it is integrated.

internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
```diff
@@ -69,7 +69,7 @@ type UpdateNodeGroupActivityInput struct {
 
 	ClusterTags map[string]string
 
-	TemplateVersion string
+	CurrentTemplateVersion semver.Version
```
sagikazarmark, line 72: I can't recall what `semver.Version` is: is it a plain old Go struct with exported fields? Asking because serializability is usually an issue.

sagikazarmark (approved):

internal/cluster/distribution/eks/eksworkflow/activity_update_node_group.go
pregnor: No, [it](https://github.com/Masterminds/semver/blob/master/version.go#L47) is not.
According to my basic [test](https://play.golang.org/p/iDz0m5uoh1Q) string and JSON marshaling works fine, however YAML marshaling does not.
Also at the stack interface it is treated as basic string (which works fine).

I picked this library because this is the most common semantic version library we are [using](https://github.com/search?q=org%3Abanzaicloud+github.com%2FMasterminds%2Fsemver) through banzaicloud, so I thought it passes the necessary requirements.

If we wish, I can change this back to a simple string, but then we have to manually construct semver.Version from it for version checking at every use case.
Alternatively I could define an internal type for this which implements YAML marshaling as expected on top of the existing third party type.

What approach would we prefer here?

sagikazarmark: The reason it works with JSON is (obviously) because it implements the marshaler interfaces: https://github.com/Masterminds/semver/blob/7bb0c843b53d6ad21a3f619cb22c4b442bb3ef3e/version.go#L415-L437

I don't think we'll be changing the default serialization layer any time soon (although if/when we switch to temporal, this could be an issue if they use a different serialization mechanism), but the downside is this is a hidden source of potential bugs. If we stick to this solution, I'd probably mark it somehow (with a searchable comment?), so we have a better understanding of the situation if we ever get to a point where we need to make some changes.

The string solution doesn't have this kind of issue, but at the same time it allows passing invalid data to the activity, resulting in a potential failure in an async workflow.

Both solutions have their downsides, I leave the decision to you.

pregnor: Fixed in https://github.com/banzaicloud/pipeline/pull/3357/commits/e177bc75c5248d64bcd105857884a046db21e3cb, decided to roll string based internal implementaiton and updated the usage in https://github.com/banzaicloud/pipeline/pull/3357/commits/3d96e45861c8565f8d66c104213a665da8000301 and https://github.com/banzaicloud/pipeline/pull/3357/commits/2de1ced9a2933b4b60d1840b4d50b3f126be6ea0.

PR #3364, opened 2020-12-08
Integrated service router implementation

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The PR contains an implementation of the integrated service interface that routes incoming requests to the appropriate service implementation (legacy or v2)

### Why?
Preparation for switching to the new version of integrated services

### Additional context
The router service is not wired into the application so it doesn't affect the current version of pipeline

### Checklist

### DIFF
--- a/internal/integratedservices/servicerouter.go
+++ b/internal/integratedservices/servicerouter.go
@@ -20,9 +20,9 @@ import (
 	"emperror.dev/errors"
 )
 
-// serviceRouter component that routes api calls to the appropriate integrated service version. It's main role is
+// serviceRouter routes api calls to the appropriate integrated service version. Its main role is
 // to make integrated service versions transparent to clients
-// Generally service version 2 is preferred, if not applicable the router falls back to version v1
+// Generally service version 2 is preferred, but if an active service with v1 exists then it must be used until the service gets deactivated. After deactivation v2 is preferred for these service as well.
 type serviceRouter struct {
 	serviceV1 Service
 	serviceV2 Service
@@ -48,27 +48,21 @@ func (s serviceRouter) List(ctx context.Context, clusterID uint) ([]IntegratedSe
 		return nil, errors.WrapIf(err, "failed to retrieve integrated services - V2")
 	}
 
-	return append(issV1, issV2...), nil
+	return s.filterDuplicates(issV1, issV2)
 }
 
-// Details retrieves the service from the service v2, if not found retrieves it from v1
+// Details retrieves the service from the service v1 if not found retrieves it from v2
 // Note: an Integrated Service can only be managed by one of the service versions
 func (s serviceRouter) Details(ctx context.Context, clusterID uint, serviceName string) (IntegratedService, error) {
-	var combined error
-	if detailsV2, err := s.serviceV2.Details(ctx, clusterID, serviceName); err == nil {
-		return detailsV2, nil
-	} else {
-		combined = errors.Append(combined, err)
-	}
-
-	// fallback to v1
 	if detailsV1, err := s.serviceV1.Details(ctx, clusterID, serviceName); err == nil {
+		// return the legacy integrated service details
 		return detailsV1, nil
-	} else {
-		combined = errors.Append(combined, err)
-	}
+	} else if !IsIntegratedServiceNotFoundError(err) {
+		return IntegratedService{}, errors.Wrapf(err, "failed to retrieve legacy integrated service details")
+	} // ignore the not found error, proceed to the new implementation
 
-	return IntegratedService{}, combined
+	// delegate to the new version of the service
+	return s.serviceV2.Details(ctx, clusterID, serviceName)
 }
 
 // Activate delegates the activation request to the appropriate service version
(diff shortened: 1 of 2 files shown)


internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return nil, errors.WrapIf(err, "failed to retrieve integrated services - V1")
+	}
+
+	issV2, err := s.serviceV2.List(ctx, clusterID)
+	if err != nil {
+		return nil, errors.WrapIf(err, "failed to retrieve integrated services - V2")
+	}
+
+	return append(issV1, issV2...), nil
```
pepov, line 51: What if both list contains the same service? I'm not sure clients will be able to handle it properly. If we want v1 services to take precedence then we should make sure v1 takes priority.

internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Details retrieves the service from the service v2, if not found retrieves it from v1
+// Note: an Integrated Service can only be managed by one of the service versions
+func (s serviceRouter) Details(ctx context.Context, clusterID uint, serviceName string) (IntegratedService, error) {
```
pepov, line 56: I would say we need v1 to take priority here as well.

internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,104 @@
...
+		combined = errors.Append(combined, err)
+	}
+
+	return IntegratedService{}, combined
+}
+
+// Activate delegates the activation request to the appropriate service version
+// New services are always activated with the version 2 service
+func (s serviceRouter) Activate(ctx context.Context, clusterID uint, serviceName string, spec IntegratedServiceSpec) error {
+	if _, err := s.serviceV1.Details(ctx, clusterID, serviceName); err == nil {
```
pepov, line 77: Errors should be handled here with the exception that the notfound error should be handled differently.

Example: in case there is an intermittent issue with the database `servicev1.Details` will return an error. We move on and activate the service on v2, which is not affected by the database error and we end up having the service activated on both versions.

internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,104 @@
...
+		// if found on the legacy service, delegate to it
+		return s.serviceV1.Activate(ctx, clusterID, serviceName, spec)
+	}
+
+	// delegate to the new implementation
+	return s.serviceV2.Activate(ctx, clusterID, serviceName, spec)
+}
+
+func (s serviceRouter) Deactivate(ctx context.Context, clusterID uint, serviceName string) error {
+	if _, err := s.serviceV1.Details(ctx, clusterID, serviceName); err == nil {
```
pepov, line 87: Error handling like in the activate method.

internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,104 @@
...
+		// if found on the legacy service, delegate to it
+		return s.serviceV1.Deactivate(ctx, clusterID, serviceName)
+	}
+
+	// delegate to the new implementation
+	return s.serviceV2.Deactivate(ctx, clusterID, serviceName)
+}
+
+func (s serviceRouter) Update(ctx context.Context, clusterID uint, serviceName string, spec map[string]interface{}) error {
+	if _, err := s.serviceV1.Details(ctx, clusterID, serviceName); err == nil {
```
pepov, line 97: Error handling here as well.

internal/integratedservices/servicerouter.go
lpuskas: thx, fixed

internal/integratedservices/servicerouter.go
lpuskas: thx, fixed

internal/integratedservices/servicerouter.go
lpuskas: thx, fixed

internal/integratedservices/servicerouter.go
lpuskas: thx, fixed

internal/integratedservices/servicerouter.go
```diff
@@ -0,0 +1,132 @@
...
+	for _, isV1 := range v1Services {
+		isV1Map[isV1.Name] = isV1
+	}
+
+	deduped := v1Services
+	for _, s2 := range v2Services {
+		if _, ok := isV1Map[s2.Name]; !ok {
+			// the isv2 doesn't exist on v1
+			deduped = append(deduped, s2)
+		}
```
pepov, line 128: we should log if there is a collision

pepov (approved):
We should log if there is a collision when listing, otherwise looks good to me!

internal/integratedservices/servicerouter.go
lpuskas: 👍

PR #3373, opened 2020-12-14
Introduce DNS spec in Integrated Service SDK

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
The DNS service spec is introduced by making a copy of the Pipeline internal spec to prepare transitioning to it simultaneously from Pipeline and from the Integrated Service Operator. To avoid breaking existing code the SDK version is pinned to the latest stable version before the change has been introduced.

### Why?
This will make the integrated service deployment code decoupled from Pipeline and make the service spec configurable independently, which will be achieved through the CRD.

### Notes
A small change had to be introduced on top of the original spec which can be seen in a separate commit: "Flag domainFilters and sources as omitempty for the sake of the K8s API validation".

These changes have been tested against the current integrated service implementation and worked as expected through the UI.


lpuskas (approved):
lgtm

colin014 (approved):

PR #3375, opened 2020-12-14
Adapt to the public DNS service spec

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Use the DNS service spec from the Integrated Service SDK.

### Why?
We want the integrated service deployment code to be decoupled from Pipeline and make the service spec configurable independently.


lpuskas (approved):
lgtm

sagikazarmark (approved):
Thanks!

PR #3377, opened 2021-01-04
Add UseIntanceStorage param to EKS node pool CF template

### LINKED ISSUE #3352: Add UseIntanceStorage param to EKS node pool CF template
Investigate using instance store on AWS instances

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3352 

### What's in this PR?
Adds an additional parameter to EKS node pool CF template `UseIntanceStorage` which works as follows:

- UseIntanceStorage = true and number of available NVMe disks on instance type (for ex. i3.large) > 0, user data script will try to list available NVMe disks, format and mount on /media/local path. This path will be set as root dir for kubelet. In case of multiple disks they will be mounted as a RAID device.

- UseIntanceStorage = true and there are no available NVMe disks on instance type (for ex. c3.large) or UseIntanceStorage = false, no root dir is set for kubelet, bootstrap should work as before. 

### Why?
To allow customers using NVMe disks for EmptyDir volumes when is available depending on chosen instance type.
(https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes)

### Checklist

### DIFF
--- a/templates/eks/amazon-eks-nodepool-cf.yaml
+++ b/templates/eks/amazon-eks-nodepool-cf.yaml
@@ -30,6 +30,8 @@ Metadata:
           - NodeVolumeSize
           - KeyName
           - BootstrapArguments
+          - KubeletExtraArguments
+          - UseInstanceStore
           - DisableIMDSv1
           - TemplateVersion
       - Label:
@@ -42,7 +44,7 @@ Parameters:
   BootstrapArguments:
     Type: String
     Default: ""
-    Description: "Arguments to pass to the bootstrap script. See files/bootstrap.sh in https://github.com/awslabs/amazon-eks-ami"
+    Description: "Arguments to pass to the bootstrap script beside the kubelet parameters. See files/bootstrap.sh in https://github.com/awslabs/amazon-eks-ami"
 
   ClusterAutoscalerEnabled:
     Type: String
@@ -76,6 +78,11 @@ Parameters:
   #   Type: Number
   #   Default: 1
   #   Description: Desired capacity of Node Group ASG.
+
+  KubeletExtraArguments:
+      Type: String
+      Default: ""
+      Description: "Arguments to pass to the bootstrap script as kubelet parameters."
 
   NodeAutoScalingGroupMaxBatchSize:
     Type: Number
@@ -481,9 +488,18 @@ Parameters:
     Type: String
     Description: Enable detachment from ASG at instance termination (true/false)
 
+  UseInstanceStore:
+      Type: String
+      Default: "false"
+      "AllowedValues":
+        - "true"
+        - "false"
+      Description: "Mount and setup available instance stores for Kubelet if available."
+
   VpcId:
     Type: "AWS::EC2::VPC::Id"
     Description: The VPC of the worker instances
+
 
 Mappings:
   PartitionMap:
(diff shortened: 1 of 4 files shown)


pregnor (approved):

pregnor (approved):

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -654,14 +670,52 @@ Resources:
...
         UserData: !Base64
-          "Fn::Sub": |
-            #!/bin/bash
-            set -o xtrace
-            /etc/eks/bootstrap.sh ${ClusterName} ${BootstrapArguments}
-            /opt/aws/bin/cfn-signal --exit-code $? \
-                     --stack  ${AWS::StackName} \
-                     --resource NodeGroup  \
-                     --region ${AWS::Region}
+            "Fn::Join":
```
orymate, line 673: why do we use this format (individual !subs in a list of inline double quoted strings joined together) instead of the original multiline literal? https://yaml-multiline.info/

Or if this is a sample from aws (I couldn't find it in a similar state), please cite the original source.

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -654,14 +670,52 @@ Resources:
...
+                    - "  if [ $SSD_NVME_DEVICE_COUNT -gt 0 ]; then"
+                    - "    mkdir /media/local"
+                    - "    mount -o defaults,noatime,discard,nobarrier $DEVICE /media/local"
+                    - "    mkdir /media/local/kubelet"
+                    - !Sub "    /etc/eks/bootstrap.sh ${ClusterName} --kubelet-extra-args '--root-dir /media/local/kubelet ${KubeletExtraArguments}'"
+                    - "  else"
+                    - !Sub "  /etc/eks/bootstrap.sh ${ClusterName} --kubelet-extra-args '${KubeletExtraArguments}'"
+                    - "  fi"
+                    - "else"
+                    - !Sub "  /etc/eks/bootstrap.sh ${ClusterName} --kubelet-extra-args '${KubeletExtraArguments}'"
```
orymate, line 716: shouldn't we add BootstrapArguments here (and the other branches)? even if we don't use it, the parameter is kept in the template, and it's not even marked as deprecated or unused in the description

orymate (changes requested):

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: It's not a sample, I've put it in this form because Sub function appears to screw up things when there are bash env vars like here for ex. 'RAID_DEVICE=${RAID_DEVICE:-/dev/md0}' that's why I decided to use Sub fn only where it's necessary.

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: I didn't removed it for compatibility reasons and may be we will use it some time, I can append to the end of param list

templates/eks/amazon-eks-nodepool-cf.yaml
orymate: I see. It seems much easier to read/maintain with the block literal and proper escaping, but handle this as an advice only.

> To write a dollar sign and curly braces (`${}`) literally, add an exclamation point (!) after the open curly brace, such as `${!Literal}`. AWS CloudFormation resolves this text as `${Literal}`.

templates/eks/amazon-eks-nodepool-cf.yaml
orymate: I'd do that. Or replace the description of the parameter with something like "deprecated, not used"

orymate (approved):

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -28,6 +28,8 @@ Metadata:
           - NodeVolumeSize
           - KeyName
           - BootstrapArguments
+          - KubeletExtraArguments
```
sagikazarmark, line 31: This feels like a BC break to me. What happens in the following scenario?

- Create cluster with the previous version of the template
- Add a whole bunch of custom labels to the node pool
- Update node pool with this template

My assumption is the labels will get lost.

sagikazarmark (changes requested):

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: We only set the pool name and version labels via Kubelet the rest is handled by Node pool label operator. As we always send labels (pool name & version) in case of a stack update in case of updating with the new template we will set KubeletExtraArguments and doesn't matter the content of BootstrapArguments param, so I think it should be fine. Anyway will test out this scenario.

sagikazarmark: > We only set the pool name and version labels via Kubelet the rest is handled by Node pool label operator.

True. I forgot how that works. Since it's a calculated parameter, I guess it should be fine then.

sagikazarmark (approved):

templates/eks/amazon-eks-nodepool-cf.yaml
```diff
@@ -653,15 +669,51 @@ Resources:
...
-            /opt/aws/bin/cfn-signal --exit-code $? \
-                     --stack  ${AWS::StackName} \
-                     --resource NodeGroup  \
-                     --region ${AWS::Region}
+        UserData:
+            Fn::Base64:
+                Fn::Sub: |
+                      #!/usr/bin/env bash
+                      set -o xtrace
+                      if [ \"${UseInstanceStore}\" == \"true\" ]; then
```
orymate, line 677:
```suggestion
                      if [ "${UseInstanceStore}" == "true" ]; then
```

sancyx: fixed

templates/eks/amazon-eks-nodepool-cf.yaml
sancyx: converted to block literal

PR #3379, opened 2021-01-06
Separate integrated service workflows for the v1 (legacy) and v2 implementations

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | yes

### What's in this PR?
Separation of workflows for the new version of integrated services

### Why?
Legacy and new implementation should be available during the transition to the new version, this separation facilitates this approach.

### Checklist

### DIFF
--- a/internal/integratedservices/integratedserviceadapter/workflow/integrated_service_job_workflow.go
+++ b/internal/integratedservices/integratedserviceadapter/workflow/integrated_service_job_workflow.go
@@ -110,13 +110,6 @@ func IntegratedServiceJobWorkflowV2(ctx workflow.Context, input IntegratedServic
 		return err
 	}
 
-	switch op := signalInput.Operation; op {
-	case OperationApply:
-	case OperationDeactivate:
-	default:
-		workflow.GetLogger(ctx).Error("unsupported operation", zap.String("operation", op))
-	}
-
 	return nil
 }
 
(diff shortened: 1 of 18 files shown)


internal/integratedservices/integratedserviceadapter/workflow/integrated_service_job_workflow.go
```diff
@@ -67,7 +70,7 @@ func IntegratedServiceJobWorkflow(ctx workflow.Context, input IntegratedServiceJ
 		return err
 	}
 
-	if err := executeJobs(ctx, input, &signalInput, jobsChannel); err != nil {
+	if err := executeJobs(ctx, input, &signalInput, jobsChannel, false); err != nil {
```
pepov, line 73: it would be nice to use some constants (even if those are just defined locally)

internal/integratedservices/servicerouter.go
```diff
@@ -58,35 +58,64 @@ func (s serviceRouter) List(ctx context.Context, clusterID uint) ([]IntegratedSe
...
+		errV1, errV2         error
+	)
+
+	if detailsV1, errV1 = s.serviceV1.Details(ctx, clusterID, serviceName); errV1 == nil {
+		if detailsV1.Status != IntegratedServiceStatusInactive {
+			// return the legacy integrated service details
+			return detailsV1, nil
+		}
+	} else if !IsUnknownIntegratedServiceError(errV1) {
+		// ignore the unknown service error, proceed to the new implementation
```
pepov, line 73: this comment is confusing here since we are processing the errors which are not the Uknown ones instead of ignoring anything

pepov (approved):

PR #3382, opened 2021-01-08
fix: Handle missing integrated service CRD on GetDetails and List API calls

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Return explicit error message for GetDetails
Return empty result for List

### Why?
The list of integrated services will be merged based on the results of the v1 and v2 integrated services. If the CRD for the v2 integrated services is missing that simply means there are no v2 integrated services enabled, thus we want to avoid bailing out with an error.

However in case of the `getDetails` api call it's more useful for clients to return a more informal error instead of the ambiguous "something went wrong".


colin014 (approved):

lpuskas (approved):
lgtm

pregnor (approved):

PR #3384, opened 2021-01-11
Add output resolvers to handle v2 service output

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Adds a new abstraction `OutputResolver` to the `ServiceConversion` component to allow filling the service output field as appropriate with values based on the current `ServiceInstance` for v2 services.

### Why?
The current form of service output resolution is not suitable for v2 integrated services.


lpuskas (approved):
lgtm

PR #3387, opened 2021-01-12
Check for running workflows with v2 integrated services

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Check for running workflows for v2 integrated services directly in Cadence to be able to tell whether they are pending.

### Why?
Without this, the clients won't be able to see pending state between the service action and the actual workflow running, which is asynchronous.


lpuskas (approved):
lgtm

Issue #3378, opened 2021-01-05
Consolidate AWS EKS node pool version calculation to be calculated through activity and passed as input


pregnor:
**Is your feature request related to a problem? Please describe.**

Currently some activities such as CreateASG calculates the AWS EKS node pool version for themselves while others take it as an input which leaves room for diverging logics due to the function being generic/variadic.

**Describe the solution you'd like to see**

Consolidate all uses to precalculate the node pool version using the corresponding activity and pass it to other activities as input.

**Describe alternatives you've considered**

- Version could be calculated inside the activities using the calculation function, but then all the version components should be passed as input which are changing frequently. If we use a precalculated version input, we can avoid updating usages whenever a new component is used in the version calculation logic, only the activity needs to be updated.

pregnor:
Done, just not closed.

PR #3390, opened 2021-01-12
Nodepool version calculation consolidation

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| API breaks?     | no|
| Deprecations?   | no
| Related tickets | #3378

### What's in this PR?

- Update CalculateNodePoolVersionActivity with wrappers
- Using the CalculateNodePoolVersionActivity for node pool version calculation in CreateNodePoolWorkflow and passing to it to the CreateASG activity instead of calculating it inside.

### Why?
Currently some activities such as CreateASG calculates the AWS EKS node pool version for themselves while others take it as an input which leaves room for diverging logics due to the function being generic/variadic.

### Checklist


internal/cluster/infrastructure/aws/awsworkflow/activity_calculate_node_pool_version.go
```diff
@@ -12,14 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package eksworkflow
+package awsworkflow
```
pregnor, line 15: What was the reason behind unifying the EKS and PKE node pool version calculation activities?

As far as I understand those two distributions are being developed separately from each other feature-parity wise which means the node pool operations allow different attributes to be modified and thus the version calculation take different inputs for EKS and PKE.
This means for example EKS node pools support EBS encryption and thus it is a parameter of the version calculation, but PKE/AWS does not yet support this feature and we don't see whether it ever will, so it is always going to be a dummy argument in that context.

I would be fine with having a fully generic hash calculation activity with a string slice input and single string output being shared between distributions or even cloud providerss as it would be distribution- and provider-independent in its interface. (But then I would certainly add distribution-specific wrappers to lock the global input argument list for a single distribution.)
But I warn against coupling distribution specific interfaces together even under the same provider as those components historically tend to develop very much independently of each other and I think dummy values are always unnecessary added complexity and require extra out of band domain knowledge to be handled properly.

internal/cluster/distribution/eks/eksprovider/workflow/activity_calculate_node_pool_version.go
```diff
@@ -62,11 +64,59 @@ func (a CalculateNodePoolVersionActivity) Execute(
...
 			fmt.Sprintf("%d", input.VolumeSize),
 			strings.Join(input.CustomSecurityGroups, ","),
 		),
 	}, nil
 }
+
+// CalculateNodePoolVersion retrieves the calculated nodePoolVersion
+//
+// This is a convenience wrapper around the corresponding activity.
+func CalculateNodePoolVersion(
```
pregnor, line 79: I think wrappers should be unexported for now, that's how I did the rest - there are a few exceptions for backward compatibility reasons, but in general they should be unexported, the package's interface is much more self-explanatory that way IMO and also we only have a handful of activities/workflows outside this package which calls these activities and those are to be refactored anyway so we don't loose much convenience either.
(I'm still evaluating whether exporting is permissible/preferable, but didn't have much time lately to think about it.)

I know it makes it a bit more difficult to use in the current update workflow, but it's going to be refactored and moved here and then can be simplified using the unexported wrapper.

internal/cluster/distribution/eks/eksprovider/workflow/activity_calculate_node_pool_version.go
```diff
@@ -62,11 +64,59 @@ func (a CalculateNodePoolVersionActivity) Execute(
...
+) workflow.Future {
+	return workflow.ExecuteActivity(ctx, CalculateNodePoolVersionActivityName, CalculateNodePoolVersionActivityInput{
+		Image:                image,
+		VolumeEncryption:     volumeEncryption,
+		VolumeSize:           volumeSize,
+		CustomSecurityGroups: customSecurityGroups,
+	})
+}
+
+func calculateNodePoolVersion(input ...string) string {
```
pregnor, line 114: **Recommended**: as we want to avoid letting other activities even from this package use the function directly, I suggest to inline the functionality of this function into the activity execution, because otherwise it's going to name clash with the unexported wrappers

internal/cluster/distribution/eks/eksprovider/workflow/create_node_pool_workflow_test.go
```diff
@@ -747,6 +747,23 @@ func TestCreateNodePoolWorkflowExecute(t *testing.T) {
...
+				func(
+					ctx context.Context,
+					input CalculateNodePoolVersionActivityInput,
+				) (*CalculateNodePoolVersionActivityOutput, error) {
+					if testCase.expectedError != nil &&
+						strings.HasPrefix(testCase.expectedError.Error(), "node pool calculation error") {
+						return nil, testCase.expectedError
+					}
+
+					return &CalculateNodePoolVersionActivityOutput{}, nil
```
pregnor, line 760: **Required**: we should return a fake value such as `test-node-pool-version` here for easier test debugging.

pregnor (changes requested):
In general the solution looks good, there were a couple minor (arguably opinionated) suggestions I made.

pregnor (approved):
Make sure to test the last version, it looks good.

PR #3395, opened 2021-01-14
Added auto merge commit rejecter check to GHA workflows

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Added auto merge commit rejecter check to GHA workflows.

### Why?

Because it is difficult to track code/commit integration in a repository with rebase merge strategy - which we are currently using.
Thus we would like to change from rebase merge strategy to (optionally squash) merge commit strategy.
But we do not want to pollute the commit history with "Merge branch 'master' into my-little-feature-branch" and "Merge pull request #123456789 from organization/my-feature-branch" automatic commits from the "Update branch" button and subpar multiple feature branch handling.
In order to avoid the noise, the best defense is preventing such commits to be merged onto the master.
For this purpose I found it easiest to add a GHA check with a 3rd party action to check the commit message for the disallowed patterns.

### Additional context

The already integrated Mergeable action could do the same check with a similar configuration, but it's report is much harder to interpret for a human so I prefer the separate action for this purpose.
For example Mergeable reports all the commits in a comma separated list and then the regexp and the failed result while the commit-message-checker reports the commits individually line by line in the log with their results like a test case.

### Checklist

### To Do

- [x] Observe the PR to pass the check as is.
- [x] Click the update branch button to auto-merge master into this branch.
- [x] Observe the PR to fail the check with the automatic merge commit.
- [x] Amend the automatic merge commit into a fake PR automatic merge commit.
- [x] Observe the PR to fail the check with the automatic merge commit.
- [x] Rebase the automatic merge commit out.
- [x] Observe the PR to pass the check as is.
- [x] Team acceptance.

### DIFF
--- a/.github/workflows/git.yaml
+++ b/.github/workflows/git.yaml
@@ -1,0 +1,21 @@
+name: Git
+on:
+  pull_request:
+  push:
+    branches-ignore:
+      - master # Note: 1, we don't check older history, 2, if something gets to master, it is considered manually accepted.
+
+jobs:
+  check-commit-message:
+    name: Check Commit Message
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check unnecessary (rebase-removable) automatic merge commits
+        uses: gsactions/commit-message-checker@v1
+        with:
+          pattern: ^(?!Merge (branch '[^']+' into |pull request \#[1-9][0-9]* from ))(.*)$ # https://regex101.com/r/VMjzUN/5/
+          error: There is an automatic merge commit on the branch or pull request which is unnecessary noise and should be removed by rebasing the branch.
+          excludeDescription: "true" # Note: we don't care about the PR description, only care about the automatic short message.
+          excludeTitle: "true" # Note: we don't care about the PR title, only care about the automatic short message.
+          checkAllCommitMessages: "true" # Note: all commit messages should be checked for a PR to prevent merging.
+          accessToken: ${{ secrets.GITHUB_TOKEN }} # Note: required to check older commit messages.


pregnor:
Note: both the PR and push checks took 3s to complete.

pregnor:
Failed check example:

Screenshot 2021-01-14 at 01 16 51
Screenshot 2021-01-14 at 01 17 18

orymate (approved):
Would you merge this together with changing the merge options on this repo to merge-commit only? The current options (both rebase and squash) remove the merge commits.

.github/workflows/git.yaml
```diff
@@ -0,0 +1,21 @@
+name: Git
+on:
+  pull_request:
+  push:
+    branches-ignore:
+      - main # Note: 1, we don't check older history, 2, if something gets to master, it is considered manually accepted.
```
sagikazarmark, line 6: master?

sagikazarmark (approved):
💯

sagikazarmark:
> Would you merge this together with changing the merge options on this repo to merge-commit only?

Squashing is okay in my opinion, rebasing should be disabled IMHO.

orymate:
> > Would you merge this together with changing the merge options on this repo to merge-commit only?
> 
> Squashing is okay in my opinion, rebasing should be disabled IMHO.

Yeah, in some cases squashing would be ok, but it won't work if the branch contains a merge commit (which would disappear in the squashed commit).

.github/workflows/git.yaml
pregnor: [Fixed](https://github.com/banzaicloud/pipeline/compare/2fe86f7aeff8909ae4a06183756550e6255fe246..27abf612e8cc0b96082e9fe120991a531197a055).

pregnor:
Yeah, I poorly worded the description, what I meant is only disabling rebase merge strategy and allowing squash/merge commit strategy.

pregnor:
Just for the record, not in response, I have quickly checked how the automatic squash merge commit is constructed and it prepends an enumeration prefix ('* ' in my test) to the commit messages in the squashed commit message body, so the regex won't be triggered on a squash merge commit.
(This is not what Máté mentioned, this was a separate concern of mine in light of the emerging discussion.)

@orymate 

> Yeah, in some cases squashing would be ok, but it won't work if the branch contains a merge commit (which would disappear in the squashed commit).

I don't think there is an easy way to put checks on whether the merger would use squash or merge commit strategy.
Which means squash merging not allowed commits would still be disallowed, because checks are preceding the merging strategy selection.

~An alternative approach would be to fail the remote/server side master pre-push hook. We may discuss whether that approach is preferable over this.~ No, it seems (public) GitHub doesn't allow server side hooks, based on no resources found on the matter and an [SO answer](https://stackoverflow.com/questions/22100581/github-server-side-git-hooks-i-e-pre-commit-and-post-commit) stating so (not the strongest ground, but together with the lack of opposing results, it seems convincing to me).

For the current implementation in the PR the merger would have to squash the commits (including at least the auto merge commit) themselves or rebase-remove the disallowed commit manually before initiating a squash merge.

What is do you all think?
Should we consider this a major user experience change to not go ahead?

(As for myself I'm willing to go with this trade-off for having merge commit strategy, but I'm curious of others' opinions.)

orymate:
I'd even be happy with squash-merge disabled. You can do it yourself if really needed, then merge the single squashed commit with a merge commit.

PR #3406, opened 2021-01-23
Cadence based implementation of components for the integrated service operator installation

### LINKED ISSUE #3399: Integrated Service Operator installer in Pipeline
The integrated service operator should be an optional component that is installed on managed clusters if the integrated service v2 experiment is enabled.

We need the following things:
- For new clusters: An **activity** that updates the helm chart repo and installs/upgrades the operator to the appropriate version which should come from the configuration. We can add this activity to the clustersetup workflow that will handle new clusters.
- For existing clusters, or when we have to upgrade: A **workflow** that can iterate through all the existing clusters and handles large number of clusters gracefully (the history shouldn't grow indefinitely). Either trigger multiple workflows with explicit cluster list or make the workflow be able to do continueAsNew after reached a certain amount of history.
- Add the workflow trigger to a **hidden API endpoint** that can be called from localhost only (similar to drain).
- Add a **pipelinectl** command that calls this endpoint.

Update: for the workflow the decision is to make a recursive workflow using continueAsNew something like this:
```
func UpdateOperator(workflow.Context ctx, lastUpdatedId string) error {
    for i < 100 {
	nextID := getNextID(ctx, lastUpdatedID)
	updateCluster(ctx, nextID)
	lastUpdatedId = nextID
    }
    if i == 100 {
    	return workflow.NewContinueAsNewError(ctx, UpdateOperator, lastUpdatedId)
    }
    return nil
}
```

lpuskas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3399 

### What's in this PR?
Cadence based components to facilitate the installation of the integrated service operator

### Why?
The integrated service operator needs to be installed on clusters in order for the new version of integrated services to be supported

### Additional context
The final approach for triggering the installation process is under discussion

### Checklist

### To Do
- [x] test the activities and the workflow, finalize the configuration


internal/cmd/config.go
```diff
@@ -905,4 +909,13 @@ traefik:
 	v.SetDefault("telemetry::debug", true)
 
 	v.SetDefault("integratedservice::v2", false)
+
+	// Integrated Service Operator
+	v.SetDefault("isoperator::enabled", true)
+	v.SetDefault("isoperator::repourl", "")
+	v.SetDefault("isoperator::reponame", "")
```
pepov, line 916: I'm not sure this should be configurable either
```suggestion
	v.SetDefault("isoperator::reponame", "banzaicloud-stable")
```

internal/cmd/config.go
```diff
@@ -905,4 +909,13 @@ traefik:
 	v.SetDefault("telemetry::debug", true)
 
 	v.SetDefault("integratedservice::v2", false)
+
+	// Integrated Service Operator
+	v.SetDefault("isoperator::enabled", true)
+	v.SetDefault("isoperator::repourl", "")
```
pepov, line 915: Do we really need the url? we know we want to update the banzaicloud stable repo. If the chart would be in a different repo, we would have to add it first.

internal/cmd/config.go
```diff
@@ -905,4 +909,13 @@ traefik:
...
+	v.SetDefault("isoperator::reponame", "")
+	v.SetDefault("isoperator::chartversion", "latest")
```
pepov, line 917: do we automatically handle the "latest" version?

internal/cmd/config.go
```diff
@@ -905,4 +909,13 @@ traefik:
...
+	v.SetDefault("isoperator::chartname", "isoperator")
```
pepov, line 918:
```suggestion
	v.SetDefault("isoperator::chartname", "integrated-service-operator")
```

internal/cmd/config.go
```diff
@@ -905,4 +909,13 @@ traefik:
...
+	v.SetDefault("isoperator::namespace", "pipeline-system")
+	v.SetDefault("isoperator::batchsize", 1)
```
pepov, line 920:
```suggestion
	v.SetDefault("isoperator::batchsize", 100)
```

src/cluster/is_operator_workflow.go
```diff
@@ -0,0 +1,47 @@
...
+
+package cluster
+
+import (
+	"context"
+
+	"go.uber.org/cadence/workflow"
+)
+
+const InstallIntServiceOperatorWorkflowName = "install-int-service-operator"
```
pepov, line 23:
```suggestion
const InstallIntServiceOperatorWorkflowName = "install-integrated-service-operator"
```

src/cluster/is_operator_workflow.go
```diff
@@ -0,0 +1,47 @@
...
+	helmService HelmService
+}
+
+func NewInstallIntegratedServiceOperatorWorkflow(helmService HelmService) *InstallIntegratedServiceOperatorWorkflow {
+	return &InstallIntegratedServiceOperatorWorkflow{
+		helmService: helmService,
+	}
+}
+
+func (w *InstallIntegratedServiceOperatorWorkflow) Execute(ctx workflow.Context, input InstallIntServiceOperatorInput) error {
```
pepov, line 39: Is this workflow for testing only?

pepov (commented):
I would consider using consistent naming and avoid different abbreviations for integrated services.

E.g. ISOperator, IntServiceOperator, IntegratedServiceOperator

I would prefer writing out IntegratedServiceOperator in type and whatever you see fit in variable names.

internal/cluster/clusteradapter/gorm_clusters.go
```diff
@@ -174,3 +174,20 @@ func (c *Clusters) GetConfigSecretIDByClusterID(organizationID uint, clusterID u
...
+	cluster := model.ClusterModel{}
+
+	err := c.db.Where("id > ?", clusterID).First(&cluster).Error
+	if err != nil {
+		if gorm.IsRecordNotFoundError(err) {
+			return 0, 0, errors.WithStack(&clusterModelNotFoundError{
+				cluster: cluster,
+			})
+		}
+		return 0, 0, err
```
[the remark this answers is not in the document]
lpuskas: will commit w/ the other changes.

internal/cmd/config.go
lpuskas: no, this is a draft only...

src/cluster/is_operator_workflow.go
lpuskas: this is a leftover. from the original changeset

internal/cluster/clustersetup/workflow.go
```diff
@@ -131,5 +134,13 @@ func (w Workflow) Execute(ctx workflow.Context, input WorkflowInput) error {
 		}
 	}
 
+	{
```
pepov, line 137: Please add the integratedservice.v2 flag check here

internal/cluster/workflow/isoperator/is_operator_activity.go
```diff
@@ -0,0 +1,122 @@
...
+
+	return ClusterRef{ID: clusterID, OrgID: orgID}, nil
+}
+
+type clusterDataProvider struct {
+	clusterID uint
+}
+
+func (c clusterDataProvider) GetK8sConfig() ([]byte, error) {
+	// no op here!
```
pepov, line 116: How come that we don't need this?

src/cluster/is_operator_workflow.go
```diff
@@ -0,0 +1,47 @@
+// Copyright © 2021 Banzai Cloud
```
pepov, line 1: if I remember right we agreed we don't need this, so please remove it, or just update this to use the correct activity that we use in the other workflow

pepov (approved):

colin014 (approved):

Issue #3410, opened 2021-01-29
Add amd64 architecture node selector to cluster autoscaler based on k8s version


pregnor:
**Is your feature request related to a problem? Please describe.**

In order to be able to use mixed architecture clusters while using k8s version < 1.20, we want to add a node selector to cluster-autoscaler which forces the container onto amd64 nodes on these versions. (ARM container support is available from CA 1.20, supporting k8s 1.20).

**Describe the solution you'd like to see**

Add a [kubernetes.io/arch](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch) [node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) set to `amd64` value to the cluster-autoscaler component's chart values based on k8s version (< 1.20) to force the cluster-autoscaler to be scheduled on amd64 nodes until the ARM support is available (CA version 1.20, supporting k8s 1.20)

**Describe alternatives you've considered**

Architecture based tainting using an operator.

**Additional context**

One possible solution is to extend [src/cluster/autoscale.go.autoscalingInfo](https://github.com/banzaicloud/pipeline/blob/e682ae60ef5bd7ca5264b53b4f01f5bd5c8ad24c/src/cluster/autoscale.go#L78) with a `nodeSelector` map which should be populated accordingly in the corresponding `createAutoscalingFor...()` functions based on k8s version (package local `getK8sVersion()` can be used to obtain a structured k8s version).

PR #3414, opened 2021-02-01
Add amd64 architecture node selector to cluster autoscaler based on k8s version

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #3410 

### What's in this PR?
Extension of  src/cluster/autoscale.go.autoscalingInfo with a nodeSelector map which  populated accordingly in the corresponding createAutoscalingFor...() functions based on k8s version (package local getK8sVersion() is used to obtain a structured k8s version).

### Why?
From the ticket:
"In order to be able to use mixed architecture clusters while using k8s version < 1.20, we want to add a node selector to cluster-autoscaler which forces the container onto amd64 nodes on these versions. (ARM container support is available from CA 1.20, supporting k8s 1.20)."

### Checklist

### DIFF
--- a/src/cluster/autoscale.go
+++ b/src/cluster/autoscale.go
@@ -176,7 +176,6 @@ func getAzureNodeGroups(cluster CommonCluster) ([]nodeGroup, error) {
 
 func createAutoscalingForEks(cluster CommonCluster, groups []nodeGroup) *autoscalingInfo {
 	eksCertPath := "/etc/ssl/certs/ca-bundle.crt"
-	nodeSelector := setArchNodeSelector(cluster.GetID(), cluster)
 
 	return &autoscalingInfo{
 		CloudProvider: cloudProviderAws,
@@ -194,7 +193,7 @@ func createAutoscalingForEks(cluster CommonCluster, groups []nodeGroup) *autosca
 			},
 		},
 		SslCertPath:  &eksCertPath,
-		NodeSelector: nodeSelector,
+		NodeSelector: setArchNodeSelector(cluster),
 	}
 }
 
@@ -233,8 +232,6 @@ func createAutoscalingForAzure(cluster CommonCluster, groups []nodeGroup, vmType
 		return nil
 	}
 
-	nodeSelector := setArchNodeSelector(cluster.GetID(), cluster)
-
 	autoscalingInfo := &autoscalingInfo{
 		CloudProvider:     cloudProviderAzure,
 		AutoscalingGroups: groups,
@@ -250,7 +247,7 @@ func createAutoscalingForAzure(cluster CommonCluster, groups []nodeGroup, vmType
 			TenantID:       clusterSecret.Values[secrettype.AzureTenantID],
 			ClusterName:    cluster.GetName(),
 		},
-		NodeSelector: nodeSelector,
+		NodeSelector: setArchNodeSelector(cluster),
 	}
 
 	switch cluster.GetDistribution() {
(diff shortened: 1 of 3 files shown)


src/cluster/autoscale.go
```diff
@@ -445,3 +452,21 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
 
 	return selectedImageVersion
 }
+
+func setArchNodeSelector(clusterID uint, cluster interface{}) map[string]string {
```
pregnor, line 456: **Recommended**: IMO the function **SHOULD** take a strongly typed `cluster` and extract the cluster ID inside the body that way the input interface could be simplified to a cluster argument. Especially so because the cluster ID is merely a debug information in the function.

src/cluster/autoscale.go
```diff
@@ -445,3 +452,21 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
...
+			return nil
+		}
+
+		if k8sVersion.LessThan(compareVersion) {
+			return map[string]string{"kubernetes.io/arch": "amd64"}
+		}
+	}
+
+	return nil
+}
```
pregnor, line 472: Have you tested it with amd64 only, mixed and arm only cluster as well?

src/cluster/autoscale.go
```diff
@@ -445,3 +449,23 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
...
+
+func setArchNodeSelector(cluster CommonCluster) map[string]string {
+	k8sVersion, err := getK8sVersion(cluster)
+	if err != nil {
+		log.Error(errors.WrapIfWithDetails(err, "unable to retrieve K8s version of cluster", "clusterID", cluster.GetID()))
+
+		return nil
+	}
+
+	compareVersion, err := semver.NewVersion("1.20.0")
```
sagikazarmark, line 461: This could be stored in a global variable initialized once.

src/cluster/autoscale.go
```diff
@@ -445,3 +449,23 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
...
+	if err != nil {
+		return nil
+	}
+
+	if k8sVersion.LessThan(compareVersion) {
```
sagikazarmark, line 466: 1.20 should be added here for this to work: https://github.com/banzaicloud/pipeline/blob/master/internal/cmd/config.go#L716

src/cluster/autoscale.go
```diff
@@ -445,3 +449,23 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
 
 	return selectedImageVersion
 }
+
+func setArchNodeSelector(cluster CommonCluster) map[string]string {
```
sagikazarmark, line 453: Does this function actually set anything?

sagikazarmark: Also, maybe add a note here that we can remove this after k8s 1.20

src/cluster/autoscale.go
```diff
@@ -445,3 +449,23 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
...
+	k8sVersion, err := getK8sVersion(cluster)
+	if err != nil {
+		log.Error(errors.WrapIfWithDetails(err, "unable to retrieve K8s version of cluster", "clusterID", cluster.GetID()))
```
sagikazarmark, line 456: I'd probably return the value and an error from this function. If the value is empty, the caller doesn't need to do anything. If it's not empty, the caller should add the arch label. If it returns an error, the caller should decide how to handle that.

sagikazarmark (approved):

sagikazarmark (dismissed):
.

src/cluster/autoscale.go
pregnor: Don't forget this.

pregnor (changes requested):

src/cluster/autoscale.go
pregnor: Don't forget this.

src/cluster/autoscale.go
```diff
@@ -445,3 +466,17 @@ func getImageVersion(clusterID uint, cluster interface{}) map[string]string {
 
 	return selectedImageVersion
 }
+
+// ToDo: This need to be removed when we no longer support k8s versions under 1.20.
+func selectArchNodeSelector(cluster CommonCluster) (map[string]string, error) {
```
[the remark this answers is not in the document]
janosSarusiKis: newAMD64ArchNodeSelector is ok for me too.

pregnor (approved):

PR #3421, opened 2021-02-03
Add amd64 architecture node selector to velero (#3411)

### LINKED ISSUE #3411: Add amd64 architecture node selector to velero and velero-plugin-for-aws

In order to be able to use mixed architecture clusters while only single architecture velero-pluging-for-aws images are available, we want to add a node selector to both velero and velero-plugin-for-aws containers which forces the containers onto amd64 nodes.

Add a [kubernetes.io/arch](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch) [node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) set to `amd64` value to the velero and velero-plugin-for-aws components' chart values to force the these containers to be scheduled on amd64 nodes until the multi-architecture ARM support is available for both containers.

Architecture based tainting using an operator.

**Additional context**

One solution is to extend the Velero configuration at [internal/ark/config.go.ValueOverrides](https://github.com/banzaicloud/pipeline/blob/master/internal/ark/config.go#L42) and the inlined [init container configurations](https://github.com/banzaicloud/pipeline/blob/master/internal/ark/config.go#L208) (based on provider) with the necessary node selector (the chart already supports it through the top level `nodeSelector` key).

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3411 

### What's in this PR?
Set node affinity for Velero operator pod to schedule only on amd64 nodes.

### Why?
In order to be able to use mixed architecture clusters while only single architecture Velero and Velero-pluging-for-aws images are available, we want to add a node selector/affinity to Velero operator pod which forces the containers onto amd64 nodes.

### Checklist

### DIFF
--- a/internal/ark/config.go
+++ b/internal/ark/config.go
@@ -15,8 +15,10 @@
 package ark
 
 import (
+	"encoding/json"
 	"fmt"
 
+	"emperror.dev/errors"
 	v1 "k8s.io/api/core/v1"
 
 	"github.com/banzaicloud/pipeline/internal/ark/client"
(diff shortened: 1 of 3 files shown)


bonifaido (approved):

internal/ark/config.go
```diff
@@ -275,6 +278,23 @@ func (req ConfigRequest) Get() (values ValueOverrides, err error) {
...
+									"key":      "kubernetes.io/arch",
+									"operator": "In",
+									"values":   []string{"amd64"},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
pregnor, line 297: **Question**: what is the reason behind favoring node affinity over [node selector](https://github.com/banzaicloud/velero-helm-chart/blob/main/charts/velero/values.yaml#L54)?

As long as the restricting expression is a single node label equaling a single value (and currently the set of values allowed has a single item) my understanding is that they are equivalent solutions (CMIIW), but the affinity is significantly more complex to define, although it is in fact easier to extend in the future should the need arise.

sagikazarmark: Node selector should be enough.

sancyx: In this simple case node selector should be fine, however it's generally is better to use affinity instead since it's newer flexible and node selector might be deprecated in future.

sagikazarmark: Hopefully we can remove this in a couple months.

pregnor: I'm fine with either solution, just wanted to know if there is an objective reason to prefer affinity over selector in case of interchangability.

pregnor (approved):

PR #3424, opened 2021-02-04
Switch the integrated-service-sdk dependency to it's new home, remove local nested module

pepov:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?
Refer to the integrated-service-sdk dependency from it's new dedicated home and remove the local nested modul.

### Why?
The module doesn't belong to this repository solely, but a dependency for both Pipeline and the Integrated Service Operator project.


sagikazarmark (approved):
🎉

lpuskas (approved):

PR #3425, opened 2021-02-04
ci: remove unused snapshot tools

sagikazarmark:
We don't use snapshot versions anymore.


pregnor (approved):

pregnor (commented):
There is also a proxy in the makefile

PR #3443, opened 2021-02-24
Remove ingress controller posthook

janosSarusiKis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | #3444 

### What's in this PR?
- Removed IngressControllerPostHook
- Added InstallIngressControllerActivity to cluster setup workflow

### Why?
To be able to deprecate post hook functionality.

### Checklist


internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -32,13 +33,31 @@ import (
...
+	helmService HelmService,
+) InstallIngressControllerActivity {
+	return InstallIngressControllerActivity{
+		config:      config,
+		helmService: helmService,
+	}
+}
+
+type InstallIngressControllerActivityInput struct {
+	ClusterID, OrgID uint
```
pregnor, line 55: **Optional**: could be opinionated, but IMO multiple field declarations on a single line decrease readability with no benefit (we spared one extra line in a file with 162 other lines).

internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -126,7 +143,7 @@ func (i *IngressControllerPostHook) Do(cluster CommonCluster) error {
 	}
 
 	// TODO: once we move this to an integrated service we must find a way to append tags to user configured annotations
-	if cluster.GetCloud() == pkgCluster.Amazon {
+	if input.Distribution == pkgCluster.EKS || input.Distribution == pkgCluster.PKE {
```
pregnor, line 146: **Required**: IMO this is not an equivalent replacement of the original condition. A PKE cluster theoretically could be on other cloud providers as well, such as Azure which is currently supported in Pipeline. I think we **MUST** keep the original condition here even if the propagation of the cloud provider information makes the code a bit more complex on the calling side.

internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -145,7 +162,19 @@ func (i *IngressControllerPostHook) Do(cluster CommonCluster) error {
 
 	namespace := global.Config.Cluster.Namespace
 
-	return i.helmService.ApplyDeployment(context.Background(), cluster.GetID(), namespace, config.Ingress.Chart, "ingress", valuesBytes, config.Ingress.Version)
+	err = a.helmService.ApplyDeployment(
```
pregnor, line 165: **Question**: are we using simple apply without reusing chart values here because we believe ingress is not deployed at this point opposed to for example the cluster autoscaler solution?

internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -145,7 +162,19 @@ func (i *IngressControllerPostHook) Do(cluster CommonCluster) error {
...
+		context.Background(),
+		input.ClusterID,
+		namespace, config.Ingress.Chart,
+		"ingress",
+		valuesBytes,
+		config.Ingress.Version)
```
pregnor, line 171: **Recommended**: it is more common/idiomatic to add a comma and break a line at the end of the last argument, because it improves readability.

internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -145,7 +162,19 @@ func (i *IngressControllerPostHook) Do(cluster CommonCluster) error {
...
+
+	if err != nil {
+		return errors.WrapIf(err, "failed to merge treafik values with config")
```
pregnor, line 174: **Recommended**: I believe the error message here should rather be `failed to install/deploy [see my other comment] ingress controller` for improved clarity when debugging the operations.

pregnor (changes requested):

internal/cluster/clustersetup/activity_install_ingress_controller.go
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2018 Banzai Cloud
+// Copyright © 2019 Banzai Cloud
```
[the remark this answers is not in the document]
pregnor: Oh I see there are divergences already, because NPLS is also called `install` while the cluster autoscaler is called `deploy`, so this kind of invalidates my comment for now (consolidating these is going to be a bigger refactor because the divergence is not being introduced here which I originally assumed).

You can skip this thread.

cmd/worker/main.go
```diff
@@ -346,6 +346,14 @@ func main() {
```
pregnor, line 346: I think the CI issue is due to `plz tidy` had not been run after the code changes and thus the `cadence-aws-sdk/aws` package which is a new dependency of the cluster setup package has not been added to the package's BUILD.plz file.

internal/cluster/clustersetup/activity_install_ingress_controller.go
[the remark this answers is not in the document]
janosSarusiKis: I will rename it to deploy, since autoscaler is newer and you too preferred that.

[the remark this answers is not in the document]
pregnor: I think it is debatable, because in Helm terminology we are installing deployments and releases and not deploying something, but I'm fine with either names for now as there is going to be at least one outlier.

pregnor (approved):
Well I definitely didn't expect this much hassle for the provider information, but that's the easiest way to propagate, then it's fine.

PR #3448, opened 2021-03-01
Velero sync api fix

### LINKED ISSUE #3449: Fix backup sync for a cluster
SyncARKBackupsOfACluster call is mapped on a wrong path should be /api/v1/orgs/{orgId}/clusters/{id}/backups/sync instead  of /api/v1/orgs/{orgId}/backups/{id}/sync. Same for restore sync per cluster.
Also cluster  backup sync should make sure to delete backups not found on cluster.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3449 

### What's in this PR?
- fix OpenAPI  path for backup / restore sync per cluster
- BackupsSyncService.SyncBackupsForCluster: delete backups not found on cluster. 
- sync backups from bucket only for inactive deployments (backup service disabled or cluster already  deleted)
- add ListForACluster to BackupService to optimise listing of buckets for a cluster

### Why?
Syncing backups both from cluster (Velero Backup CRD) and from objectstorage might cause different problems in case there are discrepancies between them, like backups hanging in InProgress or Deleting mode, when they are not yet created or already deleted in/from objectstorage. To avoid this it's enough to sync backups / restores from cluster (since if backup is enabled Backup/Restore CR's should be in sync with objectstorage) just that currently we only update backups found on cluster, we don't delete backups found in db but not on the cluster anymore. So we had to modify  backup sync to delete these.

### Checklist

### DIFF
--- a/apis/pipeline/pipeline.yaml
+++ b/apis/pipeline/pipeline.yaml
@@ -2689,7 +2689,7 @@ paths:
                 default:
                     $ref: '#/components/responses/Error'
 
-    /api/v1/orgs/{orgId}/backups/{id}/sync:
+    /api/v1/orgs/{orgId}/clusters/{id}/backups/sync:
         put:
             security:
                 - bearerAuth: []
@@ -2931,7 +2931,7 @@ paths:
                 default:
                     $ref: '#/components/responses/Error'
 
-    /api/v1/orgs/{orgId}/restores/{id}/sync:
+    /api/v1/orgs/{orgId}/clusters/{id}/restores/sync:
         put:
             security:
                 - bearerAuth: []
--- a/internal/ark/backups_repository.go
+++ b/internal/ark/backups_repository.go
@@ -56,13 +56,14 @@ func (r *BackupsRepository) FindByClusterID(clusterID uint) (backups []*ClusterB
 		ClusterID:      clusterID,
 	}
 
-	return r.db.
-	    Where(&query).
-	    Preload("Bucket").
-	    Preload("Bucket.Deployment").
-	    Preload("Organization").
-	    Find(&backups).
-	    Error
+	err = r.db.
+		Where(&query).
+		Preload("Bucket").
+		Preload("Bucket.Deployment").
+		Preload("Organization").
+		Find(&backups).
+		Error
+	return
 }
 
 // FindOneByName returns a ClusterBackupsModel instance by name
--- a/internal/ark/backups_repository.go
+++ b/internal/ark/backups_repository.go
@@ -56,9 +56,13 @@ func (r *BackupsRepository) FindByClusterID(clusterID uint) (backups []*ClusterB
 		ClusterID:      clusterID,
 	}
 
-	err = r.db.Where(&query).Preload("Bucket").Preload("Bucket.Deployment").Preload("Organization").Find(&backups).Error
-
-	return
+	return r.db.
+	    Where(&query).
+	    Preload("Bucket").
+	    Preload("Bucket.Deployment").
+	    Preload("Organization").
+	    Find(&backups).
+	    Error
 }
 
 // FindOneByName returns a ClusterBackupsModel instance by name
--- a/internal/ark/backups_repository.go
+++ b/internal/ark/backups_repository.go
@@ -49,6 +49,18 @@ func (r *BackupsRepository) Find() (backups []*ClusterBackupsModel, err error) {
 	return
 }
 
+// Find returns ClusterBackupsModel instances
+func (r *BackupsRepository) FindByClusterID(clusterID uint) (backups []*ClusterBackupsModel, err error) {
+	query := ClusterBackupsModel{
+		OrganizationID: r.org.ID,
+		ClusterID:      clusterID,
+	}
+
+	err = r.db.Where(&query).Preload("Bucket").Preload("Bucket.Deployment").Preload("Organization").Find(&backups).Error
+
+	return
+}
+
 // FindOneByName returns a ClusterBackupsModel instance by name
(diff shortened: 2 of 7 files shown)


apis/pipeline/pipeline.yaml
```diff
@@ -2689,7 +2689,7 @@ paths:
                 default:
                     $ref: '#/components/responses/Error'
 
-    /api/v1/orgs/{orgId}/backups/{id}/sync:
+    /api/v1/orgs/{orgId}/clusters/{id}/backups/sync:
```
pregnor, line 2692: **Note**: this is a public API breaking change which should be communicated accordingly and be updated in "internal" dependents. 
(Which IMO should also exclude the possibility of this change being released in a hotfix release.)

Not sure if this scope is the best to correct this API issue. 
Couldn't we leave the path alone or introduce the new path while leaving the original in place with a deprecation warning for a couple more releases before removing entirely (smoothing out the transition)?

internal/ark/sync/backups_sync_svc.go
```diff
@@ -117,6 +117,7 @@ func (s *BackupsSyncService) SyncBackupsForCluster(cluster api.Cluster) error {
 	if err != nil {
 		return errors.WrapIf(err, "could not list backups")
 	}
+	backupIDS := make([]int, 0)
```
pregnor, line 120: **Note**: typo `backupIDS` -> `backupIDs`

pregnor, line 120: **Recommended**: could be `backupIDs := make([]int, 0, len(backups.Items))` to optimize allocations in general case (and less than half of the items running into error cases) while also keeping the `append()` style filling.

internal/ark/sync/backups_sync_svc.go
```diff
@@ -153,13 +154,21 @@ func (s *BackupsSyncService) SyncBackupsForCluster(cluster api.Cluster) error {
...
 		if err != nil {
 			return errors.WrapIf(err, "could not persist backup")
 		}
+		backupIDS = append(backupIDS, int(syncedBackup.ID))
 
 		log.Debug("backup synced")
 	}
 
+	log := s.logger.WithField("bucket", bucket.Name).WithField("clusterId", bucket.ClusterID)
+	log.Debug("removing backups not found on cluster")
```
pregnor, line 167: **Question**: shouldn't we also log the `backupIDs` content here for debugging purposes? Is it not helpful in any way?

internal/ark/backups_repository.go
```diff
@@ -49,6 +49,18 @@ func (r *BackupsRepository) Find() (backups []*ClusterBackupsModel, err error) {
...
+// Find returns ClusterBackupsModel instances
+func (r *BackupsRepository) FindByClusterID(clusterID uint) (backups []*ClusterBackupsModel, err error) {
+	query := ClusterBackupsModel{
+		OrganizationID: r.org.ID,
+		ClusterID:      clusterID,
+	}
+
+	err = r.db.Where(&query).Preload("Bucket").Preload("Bucket.Deployment").Preload("Organization").Find(&backups).Error
+
+	return
```
pregnor, line 61: **Optional**: IMO explicit error check or direct return is easier to read and less error prone for future extensions here.

```suggestion
	return r.db.
	    Where(&query).
	    Preload("Bucket").
	    Preload("Bucket.Deployment").
	    Preload("Organization").
	    Find(&backups).
	    Error
```

internal/ark/backups_svc.go
```diff
@@ -105,6 +105,23 @@ func (s *BackupsService) List() ([]*api.Backup, error) {
...
 
+// List returns Backup instances for a given cluster
+func (s *BackupsService) ListForACluster(clusterID uint) ([]*api.Backup, error) {
+	backups := make([]*api.Backup, 0)
+
+	items, err := s.repository.FindByClusterID(clusterID)
+	if err != nil {
+		return backups, err
+	}
+
```
pregnor, line 116: **Recommended**: could reinitialize slice here with `backups = make([]int, 0, len(items))` to optimize allocations while also keeping the `append()` style filling.

apis/pipeline/pipeline.yaml
sancyx: Yes, however this never worked. I mean the mapping was incorrect in OpenAPI spec only and as a result it was wrong in generated client, the actual backend path is not changed, just in OpenAPI spec.

pregnor: Well, that invalidates the second part of my comment I believe.

internal/ark/sync/backups_sync_svc.go
sancyx: fixed

internal/ark/backups_svc.go
sancyx: fixed

internal/ark/sync/backups_sync_svc.go
sancyx: fixed

internal/ark/sync/backups_sync_svc.go
sancyx: Would be more interesting to see it there were any deleted items, but we don't return here anything like that.

orymate (approved):

apis/pipeline/pipeline.yaml
pregnor: Okay, now I have finally understood what you wrote, yeah, this is just a patch then.

PR #3460, opened 2021-03-09
Integrate RestoreFromBackup post hook functionality in cluster-setup

### LINKED ISSUE #3435: Integrate RestoreFromBackup post hook functionality in cluster-setup workflow
Move RestoreFromBackup post hook functionality in cluster-setup workflow in a separate activity.
Post hook input params (example see below) will be passed to cluster-setup workflow,, they will be move later into cluster create request.

```
  "postHooks": {
    "RestoreFromBackup": {
      "backupId": 999,
      "serviceAccountRoleARN": "arn:aws:iam::99999999:role/VeleroRole"
    }
  }
```

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3435 

### What's in this PR?
Move RestoreFromBackup post hook functionality in cluster-setup workflow in a separate activity.
Tested with EKS, GKE, AKS, AWS PKE.

### Why?
To be able to deprecate posthooks later.

### Checklist


internal/cluster/clustersetup/workflow.go
```diff
@@ -28,7 +28,10 @@ import (
 // WorkflowName can be used to reference the cluster setup workflow.
 const WorkflowName = "cluster-setup"
 
-const DeployClusterAutoscalerActivityName = "deploy-cluster-autoscaler"
+const (
+	DeployClusterAutoscalerActivityName = "deploy-cluster-autoscaler"
+	RestoreBackupActivityName           = "restore-backup"
+)
```
pregnor, line 34: I forgot to ask last time: why aren't these bundled to the activities themselves, residing in the activity implementation file?
(that was the convention previously)

pregnor (approved):

lpuskas (approved):

internal/cluster/clustersetup/workflow.go
sancyx: Yes, clustersetup pkg can not depend on the pkg of the activity because it would create a circular dependency to cluster pkg.

sancyx: clustersetup pkg is referenced from cluster pkg, while restore and deployAutoscaler activities depend on cluster pkg as well.

pregnor: I think a better approach would have been extracting the common cluster instead, but I understand there is a legacy hell around this, so I see why we chose this path.

PR #3475, opened 2021-04-12
Enable backup running as an integrated service 

### LINKED ISSUE #3474: Enable backup running as an integrated service
Add possibility to run Enable/Disable backup service as an integrated service.
Should `cluster.disasterRecovery.runAsIntegratedServiceV2` config property set to true, Enable/Disable backup service  will attempt to activate/deactivate backup integrated service. In case this property is set to false the Enable/Disable backup service should runs as previously installing / removing backup service using Helm.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3474 

### What's in this PR?
Adds the possibility to run Enable/Disable backup service as an integrated service.
If `cluster.disasterRecovery.runAsIntegratedServiceV2` config property set to true, Enable/Disable backup service tries to activate/deactivate backup as an integrated service. In case this property is set to false the usual flow works as before installing / removing backup service using Helm.
Restore-backup activity running as part of cluster-setup workflow is still using Helm to deploy Velero.

Manually tested use cases: enable, disable, create, list backups, restore to new cluster on EKS with cluster.disasterRecovery.runAsIntegratedServiceV2 set to both true/false.

### Why?
To be able to upgrade/downgrade Velero service independently from Pipeline.

### Checklist


internal/cluster/clustersetup/velero/restore_backup_activity.go
```diff
@@ -44,6 +44,7 @@ const (
 var (
 	nonRestorableNamespaces = []string{
 		"kube-system",
+		"pipeline-system",
```
pregnor, line 47: **Question**: does this have any effect on the restore operation? (sorry, I don't have all the context information)

internal/integratedservices/services/backup/manager.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (m Manager) GetOutput(ctx context.Context, clusterID uint, spec integratedservices.IntegratedServiceSpec) (integratedservices.IntegratedServiceOutput, error) {
+	return map[string]interface{}{
+		"externalDns": map[string]interface{}{
+			"version": m.version,
+		},
+	}, nil
+}
```
pregnor, line 42: **Question**: is this reference right? 
Shouldn't this be related to the output of the backup/velero service? (similar to [internal/integratedservices/services/backup/outputresolver.go:29](https://github.com/banzaicloud/pipeline/pull/3475/files#diff-13a0f2cfdd754f137df8a27a0bbc344157e8147fcbf8f17191c11a9439022898R29) )

pregnor (commented):
Make sure to **thoroughly** test both the helm and the ISv2 version of the backup with all its features (even listing the status) before merging the PR.

internal/cluster/clustersetup/velero/restore_backup_activity.go
sancyx: Yes it affects, resources from pipeline-system are not restored when restoring to a new cluster. It doesn't make sense to restore these, since we install it anyway during cluster setup, it may cause problems in case you have integrated services like backup activated cause the restore process will install Velero from helm.

internal/integratedservices/services/backup/manager.go
sancyx: fixed

pregnor (commented):
LGTM

pregnor (approved):

lpuskas (approved):

PR #3519, opened 2021-06-07
Fixed etcd, kube-apiserver binary sources

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Fixed etcd, kube-apiserver binary sources.

### Why?

The integration test was broken, because the old binary URLs didn't work anymore.

### Checklist


sagikazarmark (approved):
This is probably a good temporary solution for now, but I wonder if we should switch to kind instead. The integration test package is (sadly) deprecated that we use.

sancyx (approved):

PR #3527, opened 2021-06-18
Create and update core-dns as an EKS addon

### LINKED ISSUE #3529: Create and update core-dns as an EKS addon
Add possiblity of creating and updating core-dns as an EKS addon.

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3529 

### What's in this PR?
- EKS bootstrap activity: create `coredns` addon  if configuration property `distribution.eks.enableAddons` is set true and kubernetes version is >= 1.18.
- add eks-update-addon activity to update addon to the latest available version
- add eks-wait-update-addon acitvity to wait for update to finish
- cluster update: upgrade coredns to the latest available version in case it's already present on cluster

### Why?
Add possibility of creating and updating core-dns as an EKS addon.

### Additional information

To manage EKS addons the following privileges has to added to minimum privileges required to provision an EKS cluster with Banzai Cloud Pipeline:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "eks:CreateAddon",
                "eks:UpdateAddon",
                "eks:DescribeAddon",
                "eks:DescribeAddonVersions"
              ],
              "Resource": [
                  "*"
              ]
          }
      ]
  }
```
### Checklist

### DIFF
--- a/config/config.yaml.dist
+++ b/config/config.yaml.dist
@@ -395,6 +395,8 @@ dex:
 #        exposeAdminKubeconfig: true
 #        ssh:
 #            generate: true
+#        # Enable create & update of EKS addons like coredns
+#        enableAddons: true
 #
 #    pke:
 #        amazon:
--- a/config/config.yaml.dist
+++ b/config/config.yaml.dist
@@ -396,7 +396,7 @@ dex:
 #        ssh:
 #            generate: true
 #        # Enable create & update of EKS addons like coredns
-#        enableAddons: true
+#        enableAddons: false
 #
 #    pke:
 #        amazon:
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
@@ -92,7 +92,7 @@ func (a *UpdateAddonActivity) Execute(ctx context.Context, input UpdateAddonActi
 		ClusterName: aws.String(input.ClusterName),
 	}
 	addonOutput, err := eksSvc.DescribeAddon(describeAddonInput)
-	addonNotFoundErrMsg := fmt.Sprintf("No addon: %s found in cluster: %s", input.AddonName, input.ClusterName)
+	addonNotFoundErrMsg := fmt.Sprintf("No %s addon found in cluster: %s", input.AddonName, input.ClusterName)
 	if err != nil {
 		var awsErr awserr.Error
 		if errors.As(err, &awsErr) {
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
@@ -179,7 +179,7 @@ func versionIsCompatible(compatibilities []*eks.Compatibility, kubernetesVersion
 		if c.ClusterVersion == nil {
 			continue
 		}
-		if kubernetesVersion == *c.ClusterVersion {
+		if *c.ClusterVersion == kubernetesVersion {
 			return true
 		}
 	}
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
@@ -1,0 +1,172 @@
+// Copyright © 2019 Banzai Cloud
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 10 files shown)


config/config.yaml.dist
```diff
@@ -395,6 +395,8 @@ dex:
 #        exposeAdminKubeconfig: true
 #        ssh:
 #            generate: true
+#        # Enable create & update of EKS addons like coredns
+#        enableAddons: true
```
pregnor, line 399: **Recommended**: IMO the documented value **SHOULD** reflect the default value used without the configuration being present, which I suppose is `false` IIRC.
```suggestion
#        enableAddons: false
```

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	eksSvc := a.eksFactory.New(session)
+
+	describeAddonInput := &eks.DescribeAddonInput{
+		AddonName:   aws.String(input.AddonName),
+		ClusterName: aws.String(input.ClusterName),
+	}
+	addonOutput, err := eksSvc.DescribeAddon(describeAddonInput)
+	addonNotFoundErrMsg := fmt.Sprintf("No addon: %s found in cluster: %s", input.AddonName, input.ClusterName)
```
pregnor, line 95: **Recommended**: IMO the error message **SHOULD** be improved, the `: %s found in cluster` part is confusing in its current form after the `No addon` part, I propose a more straightforward phrasing.

```suggestion
	addonNotFoundErrMsg := fmt.Sprintf("No %s addon found in cluster: %s", input.AddonName, input.ClusterName)
```

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return latestVersion.Original(), nil
+}
+
+func versionIsCompatible(compatibilities []*eks.Compatibility, kubernetesVersion string) bool {
+	for _, c := range compatibilities {
+		if c.ClusterVersion == nil {
+			continue
+		}
+		if kubernetesVersion == *c.ClusterVersion {
```
pregnor, line 182: **Recommended**: IMO the [Yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions) **SHOULD** be replaced to reduce [cognitive load](https://en.wikipedia.org/wiki/Cognitive_load), because `kubernetesVersion` is a dynamically initialized constant throughout the function and `c` is the iteration variable.

```suggestion
		if *c.ClusterVersion == kubernetesVersion {
```

internal/cluster/distribution/eks/eksworkflow/activity_wait_update_addon.go
```diff
@@ -0,0 +1,107 @@
...
+					err = errors.Combine(err, errors.New(aws.StringValue(e.ErrorMessage)))
+				}
+
+				return errors.WrapIfWithDetails(err, "eks addon update failed", "cluster", input.ClusterName)
+			case eks.UpdateStatusSuccessful:
+				return nil
+			}
+
+		case <-ctx.Done():
+			return nil
```
pregnor, line 104: **Question**: Shouldn't we check here whether the context is `Done`, because it reached a timeout/deadline or was manually cancelled?
(I can see we check the EKS status whether it is cancelled, but I believe the local operation context is independent from that.)

pregnor (commented):
Are we planning to do similar work for Amazon VPC CNI and kube-proxy addons?
(Those are also supported by the addons API and I don't see any reason why shouldn't we.)

internal/cluster/distribution/eks/eksworkflow/activity_wait_update_addon.go
sancyx: Not sure if we want to check it, we are not checking when waiting for cluster update neither: https://github.com/banzaicloud/pipeline/blob/2bb2ce07702cdf7c790b23c1495dab19e3fe7f27/internal/cluster/distribution/eks/eksworkflow/activity_wait_update_version.go#L98

sancyx:
> Are we planning to do similar work for Amazon VPC CNI and kube-proxy addons?
> (Those are also supported by the addons API and I don't see any reason why shouldn't we.)

These two new activities can be used for other addons easily, basically only their name is different, so yes I think we should do this for kube-proxy and VPC CNI as it would simplify their update code, however it's only available from 1.18.

internal/cluster/distribution/eks/eksworkflow/activity_wait_update_addon.go
pregnor: Ok, then I let it go.

pregnor:
> These two new activities can be used for other addons easily, basically only their name is different, so yes I think we should do this for kube-proxy and VPC CNI as it would simplify their update code, however it's only available from 1.18.

Yeah, I'm aware of that, the same is true for CoreDNS.

If possible, I would also investigate whether we can reuse the same activities and parametrize the addon name if that is enough and structurally everything works the same.

pregnor (approved):

PR #3544, opened 2021-07-29
Updated Cadence client 0.16.0->0.17.0, updated code

pregnor:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Updated Cadence client 0.16.0->0.17.0 (also cadence-aws-sdk 0.3.0->0.4.0).

Due to a breaking change in the client [version changeset](https://github.com/uber-go/cadence-client/releases/tag/v0.17.0) (namely the split of `EntityNotExistsError` into 2, itself and `WorkflowExecutionAlreadyCompletedError`) I had to update the code at a couple locations where `EntityNotExistsError` was used directly.

1. I removed the seemingly unused `scrape-shared-spotguides` workflow cancellation from the Pipeline process.
2. Added a new `AlreadyCompletedError` error type to the Pipeline `WorkflowService` implementation and used in `CancelProcess` and `SignalProcess`, seemingly nothing depended on the old error used `NotFoundError` to be returned in case of completed workflows as well.
3. Updated the Expiry IS's CancelExpiry logic's error handling accordingly.
Previously not existing expiry workflow and any closed status expiry workflow was handled as success.
From now on `completed` and `continued-as-new` closed status expiry workflows will return an error from the cancellation attempt.
IMO the `completed` closed status workflow should not be counted as successfully cancelled expiry workflow, because the expiry workflow has run and finished thus was not cancelled successfully.
IMO the `continued-as-new` closed status expiry workflow should not be counted as successfully cancelled expiry workflow, because another expiry workflow is about to start and that should be cancelled to avoid deleting the cluster due to expiration.
So I changed the logic to handle these two cases differently than before, namely return an error explaining the completion in the first case and another error instructing the caller to retry the cancellation in the second case. 
4. Updated the PKE `node-ready` workflow signal error handling branch to use `WorkflowExecutionAlreadyCompletedError` to handle an already completed workflow instead of using `EntityNotExistsError`.
The logic was already checking for the workflow to be the completed case and not the never existed case before.

### Why?

To use the latest version.

### Additional context

Cadence server 0.21 enabled us to do the update.

Local tests were green.

### Checklist


sagikazarmark (approved):
LGTM

PR #3563, opened 2021-08-10
Vpc-cni addon should be updated only one minor version at a time

sancyx:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | fixes #3652

### Why?
Cluster update fails when updating vpc-cni plugin.

### Checklist

- [ ] Code meets the [Developer Guide](https://github.com/banzaicloud/developer-guide)
- [ ] OpenAPI and Postman files updated (if needed)

### DIFF
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
@@ -17,6 +17,7 @@ package eksworkflow
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"emperror.dev/errors"
 	"github.com/Masterminds/semver/v3"
@@ -49,13 +50,15 @@ type UpdateAddonActivityInput struct {
 	ClusterName       string
 	KubernetesVersion string
 
-	AddonName string
+	AddonName                 string
+	UpgradeToNextMinorVersion bool
 }
 
 // UpdateAddonActivityOutput holds the output data of the UpdateAddonActivityOutput
 type UpdateAddonActivityOutput struct {
 	UpdateID          string
 	AddonNotInstalled bool
+	LatestVersion     bool
 }
 
 // NewUpdateAddonActivity instantiates a new EKS addon version update
@@ -119,7 +122,7 @@ func (a *UpdateAddonActivity) Execute(ctx context.Context, input UpdateAddonActi
 		return nil, errors.WrapIfWithDetails(err, "failed to retrieve addon versions", "cluster", input.ClusterName, "addon", input.AddonName)
 	}
 
-	selectedVersion, err := selectLatestVersion(addonVersionsOutput, currentVersion, input.KubernetesVersion)
+	selectedVersion, latestVersion, err := selectNextVersion(addonVersionsOutput, currentVersion, input.KubernetesVersion, input.UpgradeToNextMinorVersion)
 	if err != nil {
 		return nil, errors.WrapIfWithDetails(err, "error selecting new version", "cluster", input.ClusterName, "addon", input.AddonName)
 	}
@@ -143,33 +146,63 @@ func (a *UpdateAddonActivity) Execute(ctx context.Context, input UpdateAddonActi
 		}
 		return nil, errors.WrapIfWithDetails(err, "failed to update addon", "cluster", input.ClusterName, "addon", input.AddonName)
 	}
-	output := UpdateAddonActivityOutput{UpdateID: aws.StringValue(updateAddonOutput.Update.Id)}
+	output := UpdateAddonActivityOutput{UpdateID: aws.StringValue(updateAddonOutput.Update.Id), LatestVersion: latestVersion}
 
 	return &output, nil
 }
 
-func selectLatestVersion(addonVersions *eks.DescribeAddonVersionsOutput, currentVersion string, kubernetesVersion string) (string, error) {
+// return latest available version which might be the currentVersion. In case selectNextMinor = true tries to return next available minor version.
+// Beisde the next version it returns whether thr returned version is the latest one.
+func selectNextVersion(addonVersions *eks.DescribeAddonVersionsOutput, currentVersion string, kubernetesVersion string, selectNextMinor bool) (string, bool, error) {
 	currentVersionSemver, err := semver.NewVersion(currentVersion)
 	if err != nil {
-		return "", err
+		return "", false, err
 	}
 	latestVersion := currentVersionSemver
 
+	versions := make([]*semver.Version, 0, 5)
 	for _, addon := range addonVersions.Addons {
 		for _, version := range addon.AddonVersions {
 			if !versionIsCompatible(version.Compatibilities, kubernetesVersion) {
 				continue
 			}
-			newVersion, err := semver.NewVersion(*version.AddonVersion)
+			version, err := semver.NewVersion(*version.AddonVersion)
 			if err != nil {
-				return "", err
+				return "", false, err
 			}
-			if newVersion.GreaterThan(latestVersion) {
-				latestVersion = newVersion
+			if version.GreaterThan(latestVersion) {
+				versions = append(versions, version)
 			}
 		}
 	}
-	return latestVersion.Original(), nil
+
+	if len(versions) == 0 {
+		return latestVersion.Original(), true, nil
+	}
+
+	sort.Slice(versions, func(i, j int) bool {
+		if versions[i].Compare(versions[j]) == -1 {
+			return true
+		}
+		return false
+	})
+
+	if selectNextMinor {
+		for i, version := range versions {
+			if !isSameMinorVersion(version, latestVersion) {
+				return version.Original(), i+1 == len(versions), nil
+			}
+		}
+	}
+
+	return versions[len(versions)-1].Original(), true, nil
+}
+
+func isSameMinorVersion(newVersion *semver.Version, currentVersion *semver.Version) bool {
+	if newVersion.Major() == currentVersion.Major() && newVersion.Minor() == currentVersion.Minor() {
+		return true
+	}
+	return false
 }
 
 // errorMessageAWSAddonNotFound is the error message returned by AWS when a
--- a/internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
+++ b/internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
@@ -61,26 +61,82 @@ func TestSelectLatestVersion(t *testing.T) {
 							},
 						},
 					},
+					{
+						AddonVersion: aws.String("v1.8.5-eksbuild.1"),
+						Compatibilities: []*eks.Compatibility{
+							{
+								ClusterVersion: aws.String("1.20"),
+							},
+						},
+					},
+					{
+						AddonVersion: aws.String("v1.8.8-eksbuild.1"),
+						Compatibilities: []*eks.Compatibility{
+							{
+								ClusterVersion: aws.String("1.20"),
+							},
+						},
+					},
+					{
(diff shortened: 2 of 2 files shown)


internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -49,13 +50,15 @@ type UpdateAddonActivityInput struct {
 	ClusterName       string
 	KubernetesVersion string
 
-	AddonName string
+	AddonName                 string
+	UpgradeToNextMinorVersion bool
```
pregnor, line 54: **Recommended**: i think we **SHOULD** rename the field `UpgradeToNextMinorVersion` to `UpgradeAtMostOneMinorVersion`, because we are using it (the true value) in the  context where we also allow patch upgrades to happen with no minor change which makes it slightly misleading as no minor update happens then and the original name instructs to upgrade exactly one minor version. Renaming would decrease cognitive load of remembering the exact semantics.

```suggestion
	UpgradeAtMostOneMinorVersion bool
```

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -143,33 +146,63 @@ func (a *UpdateAddonActivity) Execute(ctx context.Context, input UpdateAddonActi
...
+	if selectNextMinor {
+		for i, version := range versions {
+			if !isSameMinorVersion(version, latestVersion) {
+				return version.Original(), i+1 == len(versions), nil
+			}
+		}
+	}
+
+	return versions[len(versions)-1].Original(), true, nil
+}
```
pregnor, line 199: **Required**: we **MUST** adhere to the Golang comment [guidelines](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments) and [start the doc comments of entities with their name](https://golang.org/doc/effective_go#commentary).

**Recommended**: I think we **SHOULD** rename the parameter `selectNextMinor` and reword the function comment, because it is misleading the same way `UpgradeToNextMinorVersion` is, namely it falsely implies when set to true it will result in either no update or a minor version update which is not true, because we also allow patch only updates to happen in cases only those are available, so I suggest renaming it to `selectNextAtMostOneMinorVersion` and update the comment to reflect that.

**Recommended**: I think we **SHOULD** do this in a less complicated, more clean way, because while it is a series of simple steps, the number of serialized steps (and number of loops used) still make it less straightforward then a `selectLatestOrNextAtMostOneMinorVersion` operation should be and thus increases the cognitive complexity.

Suggestion regarding the comments above:
```golang
// selectNextVersion returns the next version the addon can be upgraded to based on the specified current and Kubernetes versions and possibly limiting the upgrade to at most a minor version when the provided selectNextAtMostMinorVersion is set to true or alternatively an error. 
//
// When no version is available, the function returns the current version with the latest compatible version indicator set to true.
func selectNextVersion(addonVersions *eks.DescribeAddonVersionsOutput, currentVersion string, kubernetesVersion string, selectNextAtMostOneMinorVersion bool) (string, bool, error) {
	currentVersionSemver, err := semver.NewVersion(currentVersion)
	if err != nil {
		return "", false, errors.WrapWithDetails(err, "invalid current version, not a semantic version", "currentVersion", currentVersion)
	}

	latestVersion := currentVersionSemver
    nextAtMostOneMinorVersion := currentVersionSemver

    for _, addon := range addonVersions.Addons {
        for _, version := range addon.AddonVersions {
            if version == nil ||
                !versionIsCompatible(version.Compatibilities, kubernetesVersion) {
                continue
            }

            addonVersion, err := semver.NewVersion(*version.AddonVersion)
            if err != nil {
                return "", false, errors.WrapWithDetails(err. "invalid addon version, not a semantic version", "addonVersion", *version.AddonVersion)
            }

        if addonVersion.Major() == currentVersionSemver.Major() && // Note: only consider same major version versions (multiple minor version updates at most).
            addonVersion.Minor() <= currentVersionSemver.Minor()+1 &&  // Note: only consider +1 minor version versions.
            addonVersion.GreaterThan(nextAtMostOneMinorVersion) { // Note: only set nextAtMostOneMinorVersion to higher version than it was before, initialized to currentVersionSemver.
            nextAtMostOneMinorVersion := addonVersion
        }

        if addonVersion.GreaterThan(latestVersion) {
            latestVersion := addonVersion
        }
    }
}

if selectNextAtMostOneMinorVersion {
    return nextAtMostOneMinorVersion.Original(), nextAtMostOneMinorVersion.Equal(latestVersion), nil
}

return latestVersion, true, nil
```
(GH suggestions are "not available for deleted lines", it says, that's why this is only a code block.)

**Note**: you **MUST** retest the functionality if you apply this change, because I wrote the code from heart based on the file content seen during the review.

Also `isSameMinorVersion()` is not required with this approach and can be removed.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -143,33 +146,63 @@ func (a *UpdateAddonActivity) Execute(ctx context.Context, input UpdateAddonActi
...
+
+func isSameMinorVersion(newVersion *semver.Version, currentVersion *semver.Version) bool {
+	if newVersion.Major() == currentVersion.Major() && newVersion.Minor() == currentVersion.Minor() {
+		return true
+	}
+	return false
```
pregnor, line 205: **Recommended**: this function could be simplified into the following to make it more clean/decrease the cognitive complexity.
```suggestion
	return newVersion.Major() == currentVersion.Major() && 
	    newVersion.Minor() == currentVersion.Minor()
```

**Note**: also probably this condition is not worth the trouble to be put into a function (unless building a semver library), because it is easier to understand the direct 2 part condition than to bear the cognitive load of `isSameMinorVerison() != (x.Minor() == y.Minor())`.

**Note**: with an earlier suggestion implemented this function can even be removed.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
```diff
@@ -61,26 +61,82 @@ func TestSelectLatestVersion(t *testing.T) {
...
+						},
+					},
+					{
+						AddonVersion: aws.String("v1.9.1-eksbuild.1"),
+						Compatibilities: []*eks.Compatibility{
+							{
+								ClusterVersion: aws.String("1.20"),
+							},
+						},
+					},
```
pregnor, line 95: **Question**: shouldn't these be part of the `coredns` addon's `AddonVersions`? Or if we deliberately add a new one, at least we should give it a name as well regardless of what it is (the input is not used).

If you opt for merging, the 1.8.5 version should be merged with the existing 1.8.5 verison with 2 compatibilities entry for k8s 1.19 and 1.20.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
```diff
@@ -61,26 +61,82 @@ func TestSelectLatestVersion(t *testing.T) {
...
+		assert.True(t, isLatestVersion)
+	})
+
+	t.Run("next minor version is selected for 1.18", func(t *testing.T) {
+		version, isLatestVersion, err := selectNextVersion(addonVersions, "v1.7.0-eksbuild.1", "1.18", true)
+		require.NoError(t, err)
+		assert.Equal(t, "v1.8.0-eksbuild.1", version)
+		assert.False(t, isLatestVersion)
+	})
+
```
pregnor, line 121: **Required**: we **MUST** include a test case which tests for patch upgrades, but not to the latest version, because otherwise the functionality is not assured.

```suggestion

    t.Run("next compatible patch version is selected for 1.19", func(t *testing.T){
        version, isLatestCompatibleVersion, err := selectNextVersion(addonVersions, "v1.8.3-eksbuild.1", "1.19", true)
        require.NoError(t, err)
        assert.Equal(t, "v1.8.5-eksbuild.1", version)
        assert.False(t, isLatestVersion)
    })

```

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
```diff
@@ -61,26 +61,82 @@ func TestSelectLatestVersion(t *testing.T) {
...
+		assert.Equal(t, "v1.9.1-eksbuild.1", version)
+		assert.True(t, isLatestVersion)
 	})
 
 	t.Run("no available new version", func(t *testing.T) {
-		version, err := selectLatestVersion(addonVersions, "v1.7.0-eksbuild.1", "1.20")
+		version, isLatestVersion, err := selectNextVersion(addonVersions, "v1.7.0-eksbuild.1", "1.21", false)
 		require.NoError(t, err)
 		assert.Equal(t, "v1.7.0-eksbuild.1", version)
+		assert.True(t, isLatestVersion)
```
pregnor, line 140: **Recommended**: I think the wording of the test cases `latest/next minor version` and the variable names `isLatestVersion` are slightly misleading, because they are supposed to be `latest/next compatible [minor] verison` and `isLatestCompatibleVersion`.
We **SHOULD** rename them to decrease the cognitive load of remembering it is about compatible versions only.
This is especially prominent in the last case which is the no compatible update is available case.

pregnor (changes requested):

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/pull/3563/commits/22da7696190597e681af96bcbc1fc21ee85163dc.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/pull/3563/commits/22da7696190597e681af96bcbc1fc21ee85163dc.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
pregnor: Fixed in https://github.com/banzaicloud/pipeline/pull/3563/commits/5787ff4f1ff2e689ae2abcb4195e395b8ba06fb6.

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
```diff
@@ -49,13 +50,15 @@ type UpdateAddonActivityInput struct {
...
-	AddonName string
+	AddonName                 string
+	UpgradeToNextMinorVersion bool
 }
 
 // UpdateAddonActivityOutput holds the output data of the UpdateAddonActivityOutput
 type UpdateAddonActivityOutput struct {
 	UpdateID          string
 	AddonNotInstalled bool
+	LatestVersion     bool
```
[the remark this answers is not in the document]
sancyx: fixed

internal/cluster/distribution/eks/eksworkflow/activity_update_addon.go
sancyx: refactored

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
sancyx: fixed

internal/cluster/distribution/eks/eksworkflow/activity_update_addon_test.go
sancyx: fixed

pregnor (approved):

PR #3580, opened 2021-09-24
docs: update developer documentation

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no

### What's in this PR?

Updates in the developer docs to reflect the change from `make` to `please` as the build system and also some additional information.

### Checklist

- [x] User guide and development docs updated

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -24,13 +24,13 @@ At least one of the followings has to be configured:
 To spin up the development environment with every dependencies just run the following command:
 
 ```bash
-$ make up
+$ plz up
 ```
 
 The inverse of that command is of course:
 
 ```bash
-$ make down
+$ plz down
 ```
 
 which removes everything.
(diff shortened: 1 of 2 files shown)


docs/developer.md
```diff
@@ -61,16 +61,15 @@ The development environment uses Docker Compose to create an isolated area for P
...
+$ plz start
 ```
 
 This will create a `mysql` and `vault` container:
  - Vault GUI: http://localhost:8200 login with token found in `cat ~/.vault-token`
 
-**Note:** If you want to customize mount points and port mappings, create a `docker-compose.override.yml` file via
-`make docker-compose.override.yml` and edit the file manually. Please note that you might need to edit the application
-configuration file as well.
+**Note:** If you want to customize mount points and port mappings, create a `docker-compose.override.yml` file via `make docker-compose.override.yml` and edit the file manually. Please note that you might need to edit the application configuration file as well.
```
pregnor, line 70: **Question**: is this rewrapping intentional?
What is the reason behind?
(We don't have any strict wrapping rules yet, but if it was automatically made by IDE, we should sync up, because I use a wrap-to-80-chars-long-lines extension for Markdown text.)

pregnor (approved):

docs/developer.md
akijakya: I haven't noticed that wrap rule so I just assumed it was wrapped that way for no reason and removed it. Now I checked and most of the lines don't break at 80 characters, but it is a good idea to make the line wrapping consistent.

pregnor: Yeah, I haven't "refactored" all instances to incorporate this change on any existing document, it just happened to touched/new content by automation on my end.

Issue #498, opened 2018-05-23
Handle in-progress ClusterCreation


bonifaido:
When a cluster creation is in Progress with name `cluster1` and another cluster creation request arrives with the same name `cluster1` the second request shouldn't fail but return `HTTP 20x` marking that the creation is in-progress. This change should be made in correlation with https://github.com/banzaicloud/drone-plugin-pipeline-client

sagikazarmark:
Sounds like 201 would be a perfect candidate. I would also handle the case when the cluster is being deleted when a creation request is sent: that should probably be 409 (Conflict)

Issue #2576, opened 2019-12-02
Better cluster delete flows


siliconbrain:
Make cluster deletion flows:
- more robust
- idempotent

Issue #1323, opened 2018-11-07
Deleting cluster fails if it was still being created


orymate:
**Describe the bug**
If I create a cluster, but delete it before it becomes ready, the transitions into ERROR: "There's no secret with this ID" status.

**Steps to reproduce the issue:**

* create cluster
* delete cluster immediately

**Expected behavior**
Either
* http400 for the deletion if cluster is in creating state, or
* cancel the creation somehow

Also
* store the used secret id initially (when the db record is created)

akijakya:
The issue has been resolved in #3587 by returning 409 Conflict error if deletion is called upon a cluster in CREATING status, however, it can be overwritten by force deleting it (adding the -f flag in banzai CLI or uncheck the graceful delete option on the GUI).

PR #3587, opened 2021-11-05
Fix cluster status response codes

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related issues | fixes https://github.com/banzaicloud/pipeline/issues/498, https://github.com/banzaicloud/pipeline/issues/1323, partially https://github.com/banzaicloud/pipeline/issues/2576

### What's in this PR?

Utilize more diverging response codes in some cases:
- only allow cluster update when the status of the cluster is either Running, Warning or Error, return 409 Conflict otherwise
- only allow cluster delete when the status of the cluster is either Running, Warning or Error and the force flag isn't used, return 409 Conflict otherwise
- return 201 Created on creating a cluster that is already being created
- return 409 Conflict on creating a cluster that is already being deleted

### Why?

Updating or deleting resources on a cluster that is in creating, updating or deleting status can cause issues, race conditions resulting in a broken cluster that needs to be fixed or deleted manually.

In the case of cluster creation, these should provide a clearer feedback on the cluster.

### Checklist


pregnor:
(Pre-review) Are we also handling delete being called on a deleting cluster to give an OK-ish response (2xx) and letting the running operation finish?
It could be a nice addition to handle double deletes, etc. in an easy to understand and non-erroring way.

akijakya:
> (Pre-review) Are we also handling delete being called on a deleting cluster to give an OK-ish response (2xx) and letting the running operation finish? It could be a nice addition to handle double deletes, etc. in an easy to understand and non-erroring way.

Great idea!

pregnor:
202 is probably a good answer for double deletes.

src/api/cluster_create.go
```diff
@@ -236,22 +261,39 @@ func (a *ClusterAPI) createCluster(
...
+
+		if isInvalid(err) {
+			logger.Debugf("invalid cluster creation: %s", err.Error())
+
+			return nil, badRequestErrorResponse
+		}
+
+		logger.Errorf("error during cluster creation: %s", err.Error())
+
+		return nil, internalServerErrorResponse
```
pregnor, line 296: **Recommendation**: there are a couple of things IMO we **SHOULD** do differently here. I'm going to provide a code suggestion at the end (but not as a GH suggestion, cause that doesn't work on deleted lines).

First I would advise never returning both a value and an error as on line 285 `return existingCluster, badRequestErrorResponse`, because it is very un-idiomatic in Go and would surprise a lot of Go engineers. When there is an error, the cluster shouldn't be created and thus returned, and where it is returned, the error should be nil.
Instead the raw already exists response should be returned and the cluster should be queried by the caller in case of an already existing error for the additional info and deciding the response.

Secondly, I would say the logical branching is a little convoluted here, with multiple branches ending up in similar result (`return nil, internalServerErrorResponse`, `return x, badRequestErrorResponse`), those could be straightened out and simplified in my opinion, I'm going to give it a try.

While I was trying to write this code I was struck by the issue of trying to return an error response from this function but also handling a non-error case here. IMO the best approach would be to query the cluster by name before calling this function and handling create on creating case before arriving here, that way this part can be simplified into handling the already exists case as having non-compatible status and thus returning conflict instead. 
What do you think? 
(an alternative I considered is adding an interface which is either a success response or an error response optionally or using an error response to encode a successful status code response, because 1 level upward the c.JSON() call seems to be very similar for non-error and error cases, but this short-cutting idea on the `CreateCluster()` function seems to be the easiest/best approach and it also sounds like the most most performant one as it doesn't rely try and fail logic for short-cutting the already existing and creating case).

internal/cluster/service.go
```diff
@@ -288,6 +289,14 @@ func (s service) DeleteCluster(ctx context.Context, clusterIdentifier Identifier
 		return false, err
 	}
 
+	if !clusterIsReady(c.Status) && !options.Force {
+		return false, NotReadyError{
+			OrganizationID: c.OrganizationID,
+			ID:             c.ID,
+			Name:           c.Name,
+		}
+	}
```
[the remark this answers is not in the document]
akijakya: I think it would be enough to extend the part above this code like this:

```
	if IsNotFoundError(err) || c.Status == Deleting {
		// Already deleted or being deleted
		return true, nil
	} else if err != nil {
		return false, err
	}
```

What do you think? Unless you want to be able to force delete a cluster that is already in DELETING status.

internal/cluster/service_node_pool.go
```diff
@@ -302,7 +302,11 @@ func (s service) DeleteNodePool(ctx context.Context, clusterID uint, nodePoolNam
 
 func (s service) checkCluster(cluster Cluster) error {
 	if cluster.Status != Running && cluster.Status != Warning {
-		return errors.WithStack(NotReadyError{ID: cluster.ID})
+		return errors.WithStack(NotReadyError{
+			OrganizationID: cluster.OrganizationID,
+			ID:             cluster.ID,
+			Name:           cluster.Name,
+		})
```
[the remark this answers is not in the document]
akijakya: If this can have any effect on something, I agree to put this change aside for now!

src/api/cluster_create.go
akijakya: If we check whether the cluster exists or not even before calling `clusterCreate`, wouldn't it make sense to handle all "cluster already exists" cases there, and delete all code that checks for cluster existence within `clusterCreate`?

internal/cluster/service.go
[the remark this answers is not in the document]
pregnor: 1. We don't want to force delete an already deleting cluster (unless it is stuck, but that is a special case).

2. This could work, I had similar thoughts in mind.

However it is possible this case is already handled elsewhere, because on my tests when I called a delete on my already deleting cluster it ended up returning 202 through the banzai CLI, but it is possible the case is handled either in the CLI or in the API on a higher level. This should be investigated, because it would be better if this was handled in the API than the CLI.

Example:
```
pregnor _ 99895 : PREGNOR-M-82X9 @ 84.3.170.99 . /Users/pregnor [ alpha-cluster / banzaicloud ] 
-> banzai cluster delete --cluster-name pregnor-uv0n4djkvt8eh

# Begin: 2021-11-09T09:39:51.526874000+01:00

Id    Name                   Distribution  Location      Version  CreatorName  CreatedAt             Status    StatusMessage             
6540  pregnor-uv0n4djkvt8eh  eks           eu-central-1  1.18.16  pregnor      2021-11-08T14:32:50Z  DELETING  Termination is in progress
? Do you want to DELETE the cluster? Yes
INFO[0002] Deleting cluster &{202 Accepted 202 HTTP/1.1 1 1 map[Connection:[keep-alive] Content-Length:[0] Date:[Tue, 09 Nov 2021 08:39:53 GMT] Vary:[Accept-Encoding]] 0xc000474d00 0 [] false false map[] 0xc0001b2800 0xc0002cda20} 
Id    Name                   Distribution  Location      Version  CreatorName  CreatedAt             Status    StatusMessage             
6540  pregnor-uv0n4djkvt8eh  eks           eu-central-1  1.18.16  pregnor      2021-11-08T14:32:50Z  DELETING  Termination is in progress

# Executed command: banzai cluster delete --cluster-name pregnor-uv0n4djkvt8eh
# End: 2021-11-09T09:39:53.849148000+01:00
# Duration: 2s322ms274us
# Exit code: 0
```

src/api/cluster_create.go
pregnor: It most definitely would, good idea.
It is worth checking the cluster existence beforehand then and if it exists with non-creating status then return the already exists error, if it exists with creating status, return the accepted success and if it doesn't exist then execute the cluster creation.

internal/cluster/service.go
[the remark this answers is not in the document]
akijakya: I looked into it and it seems nothing prevents double or any time of deletes, the 202 Accepted gets returned because the cluster delete workflow started again. I just started 3 cluster delete workflows on a cluster, they are listed separately in Cadence.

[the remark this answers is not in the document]
pregnor: Then IMO we should first guard against multiple deletes probably around here, returning 202 if the cluster is already deleting (and if it fails, we let it fail and the user needs to instruct Pipeline to delete the cluster again after the cluster transitions into an error state).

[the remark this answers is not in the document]
akijakya: The small modification I proposed in my first response in this thread solves this together with another minor change in internal/cluster/clusterdriver/transport_http.go

src/api/cluster_create.go
akijakya: I went with checking the cluster existence before calling cluster create as we discussed, also removed the existence checks after this point 👍

internal/cluster/clusterdriver/transport_http.go
```diff
@@ -145,8 +145,8 @@ func getOrgID(req *http.Request) (uint, error) {
 }
 
 func encodeDeleteClusterHTTPResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {
-	resp, ok := response.(DeleteClusterResponse)
-	if !ok || resp.Deleted {
+	_, ok := response.(DeleteClusterResponse)
+	if !ok {
```
pregnor, line 149: **Question**: is the `deleted` flag used when the cluster is still in deleting state?
(I'm arguing myself whether this is a good thing or not, I get it we mean the cluster has started to be deleted, but I want to double check whether the post-delete behavior changes because of this, I would think so.)

internal/cluster/clusterdriver/transport_http_test.go
```diff
@@ -40,7 +40,7 @@ func TestRegisterHTTPHandlers_DeleteCluster(t *testing.T) {
 			endpointFunc: func(ctx context.Context, request interface{}) (response interface{}, err error) {
 				return DeleteClusterResponse{Deleted: true}, nil
 			},
-			expectedStatusCode: http.StatusNoContent,
+			expectedStatusCode: http.StatusAccepted,
```
pregnor, line 43: **Note**: yeah, this was what I meant with my earlier comment, I would expect the already deleted (as in finished the delete successfully) to be `StatusNoContent` instead, because if there is nothing to do after a finished delete, it should remain the pre-existing value. However if the delete is still in progress and could run into error which would stop the delete process then it should be `StatusAccepted` indeed.

This was what I meant to split the two cases instead of merging them and changing the previous behavior, we need to handle them individually.

src/api/cluster_create.go
```diff
@@ -110,9 +110,44 @@ func (a *ClusterAPI) CreateCluster(c *gin.Context) {
 			createClusterRequest.SecretId = secret.GenerateSecretIDFromName(createClusterRequest.SecretName)
 		}
 
-		commonCluster, err := a.createCluster(ctx, &createClusterRequest, orgID, userID, createClusterRequest.PostHooks)
+		existingCluster, err := a.clusterManager.GetClusterByName(ctx, orgID, createClusterRequest.Name)
 		if err != nil {
-			c.JSON(err.Code, err)
+			c.JSON(http.StatusInternalServerError, err)
```
pregnor, line 115: **Question**: shouldn't we also handle not found error here in case no cluster exists at this point? Or does it not return an error, but only a `nil, nil`?

internal/cluster/service.go
```diff
@@ -281,13 +282,21 @@ func (s service) DeleteCluster(ctx context.Context, clusterIdentifier Identifier
 		c, err = s.clusters.GetCluster(ctx, clusterIdentifier.ClusterID)
 	}
 
-	if IsNotFoundError(err) {
-		// Already deleted
+	if IsNotFoundError(err) || c.Status == Deleting {
+		// Already deleted or being deleted
```
pregnor, line 286: **Note**: I mean we should split the logic here into two by the two conditions and handle them differently, `IsNotFoundError(err)` should be `204 NoContent` and `c.Status == Deleting` case should be `202 Accepted`. What do you think?

internal/cluster/service.go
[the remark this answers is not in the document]
pregnor: Let's discuss it under my further comments.

src/api/cluster_create.go
akijakya: Thanks for bringing my attention to this part again, we shouldn't return the error here since the expected behavior when creating a cluster is that the cluster is not existing yet. I think the error should be ignored, the check whether the`existingCluster` is `nil` in the next part should be enough.

internal/cluster/service.go
akijakya: Agreed, it feels a little hacky but it can be solved by returning `false, nil` if the status is `DELETING`...

pregnor: Yeah it does, but unfortunately anything else seems levels of complexity higher regarding the needed effort, so I will let it fly temporarily, we need to refactor this part into a more clear behavior/logic in the future.

pregnor: Actually I take it back, it's also `false,nil` when the first delete is initiated, and we want to make it idempotent for subsequent deletes, so I think it fits right into the current logic.
(Not saying it couldn't be better/more clear in an ideal scenario, but it feels more than acceptable in the current circumstances.)

src/api/cluster_create.go
```diff
@@ -110,9 +110,41 @@ func (a *ClusterAPI) CreateCluster(c *gin.Context) {
 			createClusterRequest.SecretId = secret.GenerateSecretIDFromName(createClusterRequest.SecretName)
 		}
 
-		commonCluster, err := a.createCluster(ctx, &createClusterRequest, orgID, userID, createClusterRequest.PostHooks)
-		if err != nil {
-			c.JSON(err.Code, err)
+		existingCluster, _ := a.clusterManager.GetClusterByName(ctx, orgID, createClusterRequest.Name)
```
pregnor, line 113: **Question**: wouldn't it be better if we check the returned error here instead of assuming `if existingCluster == nil then GetClusterByName() returned nil, NotFoundError{}`?
I see some benefit and safety if we handled only the `NotFoundError` (or whatever is returned in that case) as not existing and proceeding and for every other error I would immediately return 500[1], but if the cluster is successfully retrieved then I would do the following if branch's logic.

Does it make sense what I'm trying to say? What do you think?

[1]: This is the case which is handled as not existing cluster right now and the cluster creation is tried again which of course is going to fail if there is a cluster we just couldn't retrieve it this time by name.

akijakya: Unfortunately I could only check the error message, not the type of the error it is bundled up with other error messages, but it works!

src/api/cluster_create.go
```diff
@@ -110,9 +110,53 @@ func (a *ClusterAPI) CreateCluster(c *gin.Context) {
...
+				a.logger.Info("cluster does not exist yet")
+			} else {
+				c.JSON(http.StatusInternalServerError, pkgCommon.ErrorResponse{
+					Code:    http.StatusInternalServerError,
+					Message: "check if the cluster already exists failed",
+					Error:   err.Error(),
+				})
+				return
+			}
+		}
```
pregnor, line 125: **Required**: matching exact error messages is very fragile, we should either use error behavior interfaces when provided or at least error type matching when not, because it makes the check more resilient to changes.

For example if we change the error message into "cluster %cluster_name% not found" then this check is going to be broken.

This can be done by utilizing the [Go 1.13 error functions](https://go.dev/blog/go1.13-errors), namely [As](https://pkg.go.dev/emperror.dev/errors#As) (fortunately emperror provides a proxy function for it).

```golang
// src/api/error.go

type notFoundError interface {
		NotFound() bool
}

// isNotFoundError returns true if the error implements the NotFound behavior and it returns true.
func isNotFoundError(err error) bool {
    var err notFoundError

	return errors.As(err, &err) && 
        err.NotFound()
}
```

```golang
// src/api/cluster_create.go

    if err != nil && 
        !isNotFoundError(err) {
        c.JSON(http.StatusInternalServerError, pkgCommon.ErrorResponse{
		    Code:    http.StatusInternalServerError,
		    Message: "check if the cluster already exists failed",
		    Error:   err.Error(),
		})

		return
    } else if err != nil && 
        isNotFoundError(err) {
        a.logger.Info("cluster does not exist yet, creating")
    }
```

pregnor (changes requested):

src/api/cluster_create.go
akijakya: Thanks for the tip!

pregnor: Updated the example, because I messed up the naming.

src/api/cluster_create.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/banzaicloud/pipeline/.gen/pipeline/pipeline"
+	"github.com/banzaicloud/pipeline/internal/cluster/clusteradapter"
```
pregnor, line 28: **Required**: we **MUST NOT** create this dependency chain here, I intentionally wrote the example I provided in my last comment, because I do not want this package to depend on the error type or interface directly.
Not found errors are defined in a lot of places, we only care about the interface, we should redefine the interface in src/api/error.go as I wrote in my last comment with the helper function and then we can user that.

pregnor (changes requested):

src/api/cluster_create.go
akijakya: Sorry I did not realize your point, fixed!

pregnor (approved):

Issue #1684, opened 2019-01-30
Log pending spot request states while waiting for ASG healthiness


waynz0r:
We should log pending spot request states while waiting for ASG healthiness, so we could investigate if the ASG doesn't get into healthy state in time.

akijakya:
More informative logging of ASG failures has been added in #3591, in this case the logs now include the last ASG activity with the status, cause and description of the problem.

PR #3591, opened 2021-11-12
Fix logging asg failures

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related issues | https://github.com/banzaicloud/pipeline/issues/1684

### What's in this PR?

Add meaningful worker log message when the AWS autoscaling group creation fails during cluster creation and node pool update.

- ASGs can be determined either by launch configuration or launch template. If the former is not found, handle the existence of the latter before returning a "could not find launch configuration for ASG" error message.
- Include info (status, cause, description) of the last ASG activity in the error response when the ASG creation times out.

### Why?

When instances in the ASG cannot be created, e.g. because of low spot price in the config, cluster creation time is going to get really long and eventually time out without providing any feedback on the cause.

### Additional context

This is the first step to make an easy to follow cluster creation process.

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
+++ b/internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"emperror.dev/errors"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/sirupsen/logrus"
@@ -80,7 +81,24 @@ func WaitForASGToBeFulfilled(
 					return nil
 				}
 			} else {
-				return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s", waitAttempts, waitInterval)
+				asGroup, err := m.GetAutoscalingGroupByStackName(asgName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get ASG in stack: %s", waitAttempts, waitInterval, asgName)
+				}
+
+				lastActivity, err := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get last ASG activity in stack: %s", waitAttempts, waitInterval, asgName)
+				}
+
+				return errors.Errorf(
+					"waiting for ASG to be fulfilled timed out after %d x %s, status: %s, cause: %s, description: %s",
+					waitAttempts,
+					waitInterval,
+					aws.StringValue(lastActivity.StatusCode),
+					aws.StringValue(lastActivity.Cause),
+					aws.StringValue(lastActivity.Description),
+				)
 			}
 		case <-ctx.Done(): // wait for ASG fulfillment cancelled
 			return nil
--- a/internal/cluster/distribution/eks/eksprovider/workflow/amazon.go
+++ b/internal/cluster/distribution/eks/eksprovider/workflow/amazon.go
@@ -483,15 +483,31 @@ func WaitForASGToBeFulfilled(
 					if autoscaling.IsErrorFinal(err) {
 						return errors.WithDetails(err, "nodePoolName", nodePoolName, "stackName", aws.StringValue(asGroup.AutoScalingGroupName))
 					}
-					// log.Debug(err)
 					continue
 				}
 				if ok {
-					// log.Debug("ASG is healthy")
+					logger.Debug("ASG is healthy")
 					return nil
 				}
 			} else {
-				return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s", asgFulfillmentWaitAttempts, asgFulfillmentWaitInterval)
+				asGroup, err := m.GetAutoscalingGroupByStackName(stackName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get ASG in stack: %s", asgFulfillmentWaitAttempts, asgFulfillmentWaitInterval, stackName)
+				}
+
+				lastActivity, err := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get last ASG activity in stack: %s", asgFulfillmentWaitAttempts, asgFulfillmentWaitInterval, stackName)
+				}
+
+				return errors.Errorf(
+					"waiting for ASG to be fulfilled timed out after %d x %s, status: %s, cause: %s, description: %s",
+					asgFulfillmentWaitAttempts,
+					asgFulfillmentWaitInterval,
+					aws.StringValue(lastActivity.StatusCode),
+					aws.StringValue(lastActivity.Cause),
+					aws.StringValue(lastActivity.Description),
+				)
 			}
 		case <-ctx.Done(): // wait for ASG fulfillment cancelled
 			return nil
--- a/pkg/providers/amazon/autoscaling/autoscaling.go
+++ b/pkg/providers/amazon/autoscaling/autoscaling.go
@@ -99,10 +99,12 @@ func (group *Group) getSpotRequests() ([]*ec2.SpotInstanceRequest, error) {
 	}
 
 	lc, err := group.getLaunchConfiguration()
-	if err == nil && lc == nil {
-		err = errors.NewWithDetails("could not find launch configuration for ASG", "asg", group.getName())
-	}
 	if err != nil {
+		if _, ok := err.(errLaunchConfigurationNotFound); ok {
+			if group.LaunchTemplate != nil {
+				return nil, nil
+			}
+		}
 		return nil, err
 	}
 
@@ -128,7 +130,7 @@ func (group *Group) getLaunchConfiguration() (*autoscaling.LaunchConfiguration, 
 	asgName := group.getName()
 
 	if group.LaunchConfigurationName == nil {
-		return nil, errors.NewWithDetails("could not find launch configuration for ASG", "asg", asgName)
+		return nil, newLaunchConfigurationNotFoundError(asgName)
 	}
 
 	input := &autoscaling.DescribeLaunchConfigurationsInput{
@@ -144,6 +146,10 @@ func (group *Group) getLaunchConfiguration() (*autoscaling.LaunchConfiguration, 
 
 	if len(result.LaunchConfigurations) != 1 {
 		return nil, errors.WrapIfWithDetails(errors.WithDetails(errors.New("invalid response count"), "count", len(result.LaunchConfigurations)), "could not get launch configuration for ASG", "asg", asgName)
+	}
+
+	if len(result.LaunchConfigurations) == 0 {
+		return nil, newLaunchConfigurationNotFoundError(asgName)
 	}
 
 	return result.LaunchConfigurations[0], nil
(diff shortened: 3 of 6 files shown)


pregnor:
> When the ASG cannot be created, e.g. because of low spot price in the config, cluster creation time is going to get really long and eventually time out without providing any feedback on the cause.

Just a note: technically in this case the ASG is created, only the instances are not created because of the low spot price (just phrasing clarification).

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
```diff
@@ -80,7 +81,21 @@ func WaitForASGToBeFulfilled(
...
+				if lastActivity != nil {
+					return errors.Errorf(
+						"waiting for ASG to be fulfilled timed out after %d x %s, status: %s, cause: %s, description: %s",
+						waitAttempts,
+						waitInterval,
+						aws.StringValue(lastActivity.StatusCode),
+						aws.StringValue(lastActivity.Cause),
+						aws.StringValue(lastActivity.Description),
+					)
+				}
```
pregnor, line 98: **Question**: is there no error in the spot price timeout case in the `GetAutoscalingGroupByStackName()` function?

pkg/providers/amazon/autoscaling/autoscaling.go
```diff
@@ -99,10 +99,12 @@ func (group *Group) getSpotRequests() ([]*ec2.SpotInstanceRequest, error) {
 	}
 
 	lc, err := group.getLaunchConfiguration()
-	if err == nil && lc == nil {
-		err = errors.NewWithDetails("could not find launch configuration for ASG", "asg", group.getName())
-	}
 	if err != nil {
+		if _, ok := err.(errLaunchConfigurationNotFound); ok {
+			if group.LaunchTemplate != nil {
+				return nil, nil
```
pregnor, line 105: **Question**: in this case, couldn't (shouldn't?) we check the launch template for the same things we check the launch configuration for during the logic following this point in the function (spot price, spot instance requests)?

Are they so different we can't do anything here?
Because if we could, I would sleep better knowing we do check it otherwise in EKS cases because of the launch template we are going to skip those checks entirely and silently with this shortcut.

I have a feeling spot requests should be query-able from launch templates as well (maybe through some other resources created from launch templates if not directly).

To clarify, I just don't like returning `nil, nil` from a `getSpotRequests()` function when there is a launch template with potential spot requests to check (first because it is less idiomatic, but this is the least of my problems, second because of the not checked/verified spot requests).
Maybe if we could properly retrieve the launch template spot requests we wouldn't need to be manually checking for spot errors in the `WaitForASGToBeFulfilled()` function, because we would already got our error from the spot request checking in the ASG somewhere around here.

pkg/providers/amazon/autoscaling/manager.go
```diff
@@ -129,3 +131,31 @@ func (m *Manager) GetAutoscalingGroupByStackName(stackName string) (*Group, erro
 
 	return NewGroup(m, describeAutoScalingGroupsOutput.AutoScalingGroups[0]), nil
 }
+
+// GetLastAutoscalingActivity returns the last ASG activity recorded.
+func (m *Manager) LastAutoscalingActivity(asgName *string) (*autoscaling.Activity, error) {
+	ctx, cancelFunction := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancelFunction()
```
pregnor, line 138: **Optional**: it would be more proper & idiomatic to take the context from the caller through a parameter in a cascadic fashion and only pass it here, sorry this was a quick solution during the pair programming when I wanted to check whether we can query the last ASG activity for failure reasons.

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
akijakya: There is no error but it can be inferred there is a problem by checking the length of the Instances slice (if the instances cannot come up this is empty).

pkg/providers/amazon/autoscaling/autoscaling.go
akijakya: I checked and I only found spot requests related to PKE clusters, so I thought things work differently when launch templates are being used, but I couldn't yet figure out exactly how.

There is another `nil, nil` return a bit downwards, that is way I went with this. In the function that calls this one, there is a check for this scenario, but it would be nice to handle things a different way for sure.

pkg/providers/amazon/autoscaling/manager.go
akijakya: Nice catch, done!

pkg/providers/amazon/autoscaling/autoscaling.go
pregnor: Okay, well done then, I'm going to double check just to make sure (stand by) but if we can't do it any better way, this will have to suffice.

pregnor: The only thing I found is that the spot price is part of the launch template **version's** description, it could be queried from there in case of an LT. This can be a substitute in the `getSpotRequests()` function in place of the launch configuration's spot price, but as it is already part of the ASG last activity error, there is no need for further development.

The spot requests can be queried from the EC2 API's DescribeSpotInstanceRequests, but it seems like even without filtering I could not find the LT's spot requests, it is possible the ASG requests instances directly so we cannot build pretty requests like in the case of PKE.

Because of this I'm all for letting the spot requests go and use the ASG activity for informational purposes.

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
```diff
@@ -80,7 +81,21 @@ func WaitForASGToBeFulfilled(
 					return nil
 				}
 			} else {
-				return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s", waitAttempts, waitInterval)
+				asGroup, err := m.GetAutoscalingGroupByStackName(asgName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get ASG in stack: %s", waitAttempts, waitInterval, asgName)
+				}
+				lastActivity, _ := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
+				if lastActivity != nil {
```
pregnor, line 89: **Required**: I would check for error here (and in the next file as well at the same point in the logic), because if we don't have the last activity, we should return the generic error message like above, otherwise proceed with the detailed error in the current if branch.

While we are here it is already an error case (timeout) so I would put the error checking into an if branch and let the detailed error be returned in case the activity is successfully retrieved.

pregnor (changes requested):
Sorry, I didn't catch this missing error checking before.

Otherwise the PR is 💯 , nothing else to do right now.

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
akijakya: You are totally right, sorry for not thinking that through!

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
```diff
@@ -80,7 +81,24 @@ func WaitForASGToBeFulfilled(
...
-				return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s", waitAttempts, waitInterval)
+				asGroup, err := m.GetAutoscalingGroupByStackName(asgName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get ASG in stack: %s", waitAttempts, waitInterval, asgName)
+				}
+
+				lastActivity, _ := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
+				if err != nil {
+					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get last ASG activity in stack: %s", waitAttempts, waitInterval, asgName)
+				}
```
pregnor, line 92: **Required**: you forgot to reassign the err variable before the check.

```suggestion
				lastActivity, err := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
				if err != nil {
					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get last ASG activity in stack: %s", waitAttempts, waitInterval, asgName)
				}
```

pregnor (changes requested):

internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go
akijakya: Oh my! Good catch, thanks!

internal/cluster/distribution/eks/eksprovider/workflow/amazon.go
```diff
@@ -483,15 +483,28 @@ func WaitForASGToBeFulfilled(
...
+				if lastActivity != nil {
+					return errors.Errorf(
+						"waiting for ASG to be fulfilled timed out after %d x %s, status: %s, cause: %s, description: %s",
+						asgFulfillmentWaitAttempts,
+						asgFulfillmentWaitInterval,
+						aws.StringValue(lastActivity.StatusCode),
+						aws.StringValue(lastActivity.Cause),
+						aws.StringValue(lastActivity.Description),
+					)
+				}
```
pregnor, line 507: **Question**: shouldn't we do the same here as [there](https://github.com/banzaicloud/pipeline/pull/3591/files#diff-3ee9d719848b4882a759b0611d32bbf3df439588b845ba3f528f1819c5c4b271R89-R101)?

akijakya: Again, thanks for noticing! 🙂

internal/cluster/distribution/eks/eksprovider/workflow/amazon.go
```diff
@@ -483,15 +483,28 @@ func WaitForASGToBeFulfilled(
...
```
pregnor, line 507: **Recommended**: If there is no reason to make the `internal/cluster/distribution/eks/ekscluster/nodepools/wait_for_asg.go` and `internal/cluster/distribution/eks/eksprovider/workflow/amazon.go` versions different, I would propose to use the exact same code for the 2.

(I'm not sure why there are 2 different implementations, I want to deduplicate those in the future if possible in a different PR, but until it's done I would prefer keeping them in sync unless required not to.)

```suggestion
				asGroup, err := m.GetAutoscalingGroupByStackName(asgName)
				if err != nil {
					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get ASG in stack: %s", waitAttempts, waitInterval, asgName)
				}

				lastActivity, err := m.LastAutoscalingActivity(ctx, asGroup.AutoScalingGroupName)
				if err != nil {
					return errors.Errorf("waiting for ASG to be fulfilled timed out after %d x %s, could not get last ASG activity in stack: %s", waitAttempts, waitInterval, asgName)
				}

				return errors.Errorf(
					"waiting for ASG to be fulfilled timed out after %d x %s, status: %s, cause: %s, description: %s",
					waitAttempts,
					waitInterval,
					aws.StringValue(lastActivity.StatusCode),
					aws.StringValue(lastActivity.Cause),
					aws.StringValue(lastActivity.Description),
				)
```

pregnor (changes requested):

internal/cluster/distribution/eks/eksprovider/workflow/amazon.go
pregnor: Apologies for not being clear enough in my previous comment, I thought it is obvious I meant to make all the changes here as well as done there, not just the `err` assignment.

akijakya: Sorry, I just keep forgetting that this code is present in two places...

pregnor: Yeah, I advise you to not rush the different tasks, because context switching is difficult to be done right.

Take your time, this is the chance to learn how things work, don't rush it, there is no need.

pregnor (approved):

Issue #2622, opened 2020-01-03
Validate node pool name


sagikazarmark:
**Is your feature request related to a problem? Please describe.**
Node pool names should be validated:

```
poolX": a DNS-1123 subdomain must consist of lower case alphanumeric characters
```

**Describe the solution you'd like to see**
Add a validation rule similar to the rest of node pool fields

PR #3592, opened 2021-11-15
feat: add node pool name validation

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related issues | fixes https://github.com/banzaicloud/pipeline/issues/2622

### What's in this PR?

Adding validation for node pool names (only accept lowercase characters and numbers) for clusters created on AWS.

### Why?

The `eks-create-asg` Cadence workflow errors if there are special characters because the node pool name gets attached to the stack name; also the `configure-nodepool-labels` workflow errors if there are uppercase characters in the nood pool name resulting in unsuccessful cluster creation, so it is useful if a validator catches these cases even before the cluster creation begins.

### Additional context

Example config.yaml:

```
{
  "cloud": "amazon",
  "location": "eu-west-1",
  "name": "test-validation",
  "properties": {
    "eks": {
      "nodePools": {
        "pool1X": {
          "autoscaling": true,
          "count": 5,
          "instanceType": "t3a.small",
          "maxCount": 10,
          "minCount": 2
        }
      },
      "version": "1.20"
    }
  },
  "secretName": "secret_name"
}
```

Response (banzai CLI):

```
ERRO[0013] failed to create cluster: 400 Bad Request (err pipeline.GenericOpenAPIError, request=map[cloud:amazon location:eu-west-1 name:test-validation properties:map[eks:map[nodePools:map[pool1X:map[autoscaling:true count:5 instanceType:t3a.small maxCount:10 minCount:2]] version:1.20]] secretName:secret_name], response={"code":400,"message":"node pool name 'pool1X' invalid: a DNS-1123 subdomain must consist of lower case alphanumeric characters","error":"node pool name 'pool1X' invalid: a DNS-1123 subdomain must consist of lower case alphanumeric characters"}
)
Error: failed to create cluster: 400 Bad Request
failed to create cluster: 400 Bad Request
```

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/ekscluster/eks.go
+++ b/internal/cluster/distribution/eks/ekscluster/eks.go
@@ -182,6 +182,11 @@ const (
 
 // Validate checks Amazon's node fields
 func (a *NodePool) Validate(npName string) error {
+	// ---- [ Node pool name validation ] ---- //
+	if err := pkgCommon.ValidateNodePoolName(npName); err != nil {
+		return err
+	}
+
 	// ---- [ Node instanceType check ] ---- //
 	if len(a.InstanceType) == 0 {
 		return pkgErrors.ErrorInstancetypeFieldIsEmpty
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -17,6 +17,7 @@ package common
 import (
 	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
@@ -71,6 +72,14 @@ const (
 	SpotConfigMapKey = "spot-deploy-config"
 )
 
+// nodePoolNameRegexp is matching a valid node pool name
+// (only lower case alphanumeric characters and single dashes within the string)
+//
+// Regular expression: https://regex101.com/r/OrRrlw/1
+var (
+	nodePoolNameRegexp = regexp.MustCompile(`^[a-z0-9]([a-z0-9]*[a-z0-9])?(-[a-z0-9]([a-z0-9]*[a-z0-9])?)*$`) // nolint:gochecknoglobals
+)
+
 // ErrorResponseWithStatus aborts the http request with a JSON error response with the given status code and error
 func ErrorResponseWithStatus(c *gin.Context, status int, err error) {
 	if c.Writer.Status() != http.StatusOK {
@@ -102,4 +111,16 @@ func ValidateNodePoolLabels(nodePoolName string, labels map[string]string) error
 	}
 
 	return nil
+}
+
+// ValidateNodePoolName checks the length and validity of the node pool name
+func ValidateNodePoolName(nodePoolName string) error {
+	if len(nodePoolName) > 63 {
+		return errors.New(fmt.Sprintf("node pool name '%s' too long: over 63 characters", nodePoolName))
+	}
+
+	if nodePoolNameRegexp.MatchString(nodePoolName) {
+		return nil
+	}
+	return errors.New(fmt.Sprintf("node pool name '%s' invalid: must consist of lower case alphanumeric characters and single dashes within the string", nodePoolName))
 }


pkg/common/common.go
```diff
@@ -71,6 +72,11 @@ const (
 	SpotConfigMapKey = "spot-deploy-config"
 )
 
+// Constant for node pool name validation
+const (
+	nodePoolNameValidationRegexp = "^[a-z0-9]*$"
```
pregnor, line 77: **Required**: this is more restrictive than it should be, we **MUST** relax it a bit to allow previously allowed names as well. 

According to the [AWS CloudFormation naming](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html) (in the end we create CF stacks to manage most resources), dashes should definitely also be allowed, however they cannot begin or end the string or have 2 consecutive dashes.

Also according to my experience (supportive [link](https://gist.github.com/christianklotz/468a240a75f4ae3e73bd52722f926acb)) there is a character count limit on the AWS CF stack names (63 characters IIRC) which we should employ in order to prefilter the node pool names.
This limit should be lowered for the actual node pool name because we are stitching together the cluster name, node pool name and other IDs for the final node pool stack name, but it will have to suffice for now, at least we are guarding the most obvious mistakes.

Also it might be worth to use the [original Kubernetes DNS 1123 error message](https://github.com/kubernetes/kubernetes/pull/39675/files) or parts of it. This also states dots as allowed characters, I would give it a try whether they work just to make sure we don't overly limit the naming schema and also tailor the resulting error message accordingly.

P.S.: In the long run it would probably be a good idea to switch to using (short) unique IDs instead of names in stack naming, but that's a very long road to walk on.

pregnor: **Note/Recommended**: it is a good idea to include a comment reference next to the in-code raw regexes to some regex testing sites with saved examples (I often use regex101.com) to provide easy access to immediate verification with pre-saved test cases to see the regex in working. It is more prominent for complex regexes, but for newcomer's sake it is a good practice IMO.

pkg/common/common.go
```diff
@@ -103,3 +109,14 @@ func ValidateNodePoolLabels(nodePoolName string, labels map[string]string) error
...
 	return nil
 }
+
+// ValidateNodePoolName checks the validity of the node pool name
+func ValidateNodePoolName(nodePoolName string) error {
+	if matched, err := regexp.MatchString(nodePoolNameValidationRegexp, nodePoolName); !matched {
+		return errors.New(fmt.Sprintf("node pool name '%s' invalid: a DNS-1123 subdomain must consist of lower case alphanumeric characters", nodePoolName))
+	} else if err != nil {
+		return errors.WrapIf(err, "couldn't validate node pool name")
+	}
```
pregnor, line 119: **Recommended**: it is general practice to check error first and matching later, because errors can screw with results. In this example when the regex returns an error the matched value is probably left untouched which means the it will be the default value which is false for boolean so the first if branch will always be executed even when there is an error.

Also for static regexes it might be worth saving them as a compiled regex with `regexp.MustCompile()` and use their member functions for matches, because that cannot generate any error so the logic gets easier. The must compilation can fail and it throws a panic on call (initialization time for globals) which is generally avoided, that's why I advise it only for static regexes (where any issue would be brought to light during initial testing).

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -232,6 +233,11 @@ func (a *NodePool) Validate(npName string) error {
 		return err
 	}
 
+	// ---- [ Node pool name validation ] ---- //
+	if err := pkgCommon.ValidateNodePoolName(npName); err != nil {
+		return err
+	}
+
```
pregnor, line 240: **Optional**: I personally would start with the name validation, but this is entirely optional, just a thought.

pregnor (changes requested):

pkg/common/common.go
akijakya: When I was checking the behavior with different node pool names, I encountered two different error messages (because the node pool name gets attached to the stack name and that has another constraints):

```
NodePoolLabelSet.labels.banzaicloud.io \"poolX\" is invalid: metadata.name: Invalid value: \"poolX\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```
```
ValidationError: 1 validation error detected: Value 'pipeline-eks-nodepool-test-validation-pool1X'' at 'stackName' failed to satisfy constraint: Member must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]*
```

So it seemed to me that is why the wording in the [issue](https://github.com/banzaicloud/pipeline/issues/2622) says "must consist of lower case alphanumeric characters", to satisfy both of these regexes. You are right about the `-` though and the character limit is also a good idea, I am going to refine the regexp!

internal/cluster/distribution/eks/ekscluster/eks.go
akijakya: I put it there first but then thought it would look better together with the label validation at the end, but definitely makes more sense in to start with that 🙂

pkg/common/common.go
akijakya: Thanks, will do!

pkg/common/common.go
akijakya: I crafted a regexp that allows single dashes within the string and also added node pool name verification.

pregnor: Then you are right, the correct approach is to use the top regex + the length constraint.

internal/cluster/distribution/eks/ekscluster/eks.go
pregnor: I meant just before the label validation, not at the start of the function, but it is good as it is (whichever it is right now), not worth cycles over it :) .

internal/cluster/distribution/eks/ekscluster/eks.go
```diff
@@ -178,10 +178,16 @@ const (
 	DEFAULT_VPC_CIDR     = "192.168.0.0/16"
 	DEFAULT_SUBNET0_CIDR = "192.168.64.0/20"
 	DEFAULT_SUBNET1_CIDR = "192.168.80.0/20"
+	// nodePoolNameValidationRegexp = "^[a-z0-9]*$"
```
pregnor, line 181: **Question**: is this comment left in intentionally? What purpose does it serve? (If nothing, I would propose to remove it.)

pregnor (approved):

internal/cluster/distribution/eks/ekscluster/eks.go
akijakya: I forgot to remove it, thanks for noticing!

pregnor (approved):

Issue #3162, opened 2020-09-09
chore(term): consolidate EBS volume size measurement units to GiB


pregnor:
**Is your feature request related to a problem? Please describe.**

GB unit is used throughout the code base for description, while the [AWS EC2 instance EBS VolumeSize definition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html) states the unit is GiB instead.

**Describe the solution you'd like to see**

Replace the GB references with GiB.
Note: the values are transferred transparently through integer holders so this does not have an impact on code execution, it is merely for clarifying the documentation of the corresponding variables and meeting the end user expectations and relations between the value put in and the resulting volume size.

**Describe alternatives you've considered**

-

**Additional context**

-

akijakya:
This issue got resolved in PR #3593

PR #3593, opened 2021-11-16
Consolidate EBS volume size units to GiB

akijakya:
GB unit is used throughout the code base for description, while the AWS EC2 instance EBS VolumeSize definition states the unit is GiB instead.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related issues | https://github.com/banzaicloud/pipeline/issues/3162

### What's in this PR?

Consolidate EBS volume size measurement units to GiB

### Why?

GB unit is used throughout the code base for description, while the AWS EC2 instance EBS VolumeSize definition states the unit is GiB instead.

### Additional context

The values are transferred transparently through integer holders so this does not have an impact on code execution, it is merely for clarifying the documentation of the corresponding variables and meeting the end user expectations and relations between the value put in and the resulting volume size.

### Checklist

### DIFF
--- a/internal/cluster/distribution/eks/eksprovider/workflow/select_volume_size_activity.go
+++ b/internal/cluster/distribution/eks/eksprovider/workflow/select_volume_size_activity.go
@@ -70,7 +70,7 @@ func (a *SelectVolumeSizeActivity) Execute(ctx context.Context, input SelectVolu
 
 	if output.VolumeSize < input.AMISize {
 		return nil, errors.New(fmt.Sprintf(
-			"selected volume size of %d GB (source: %s) is less than the AMI size of %d GB",
+			"selected volume size of %d GiB (source: %s) is less than the AMI size of %d GiB",
 			output.VolumeSize, valueSource, input.AMISize,
 		))
 	}
--- a/.gen/pipeline/api/openapi.yaml
+++ b/.gen/pipeline/api/openapi.yaml
@@ -19035,7 +19035,7 @@ components:
         volumeEncryption:
           $ref: '#/components/schemas/EKSNodePoolVolumeEncryption'
         volumeSize:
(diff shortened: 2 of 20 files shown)


internal/cluster/distribution/eks/eksprovider/workflow/select_volume_size_activity.go
```diff
@@ -70,7 +70,7 @@ func (a *SelectVolumeSizeActivity) Execute(ctx context.Context, input SelectVolu
 
 	if output.VolumeSize < input.AMISize {
 		return nil, errors.New(fmt.Sprintf(
-			"selected volume size of %d GB (source: %s) is less than the AMI size of %d GB",
+			"selected volume size of %d GiB (source: %s) is less than the AMI size of %d GiB",
```
pregnor, line 73: **Note**: we need to add this to the release changes as a patch, because people could be depending on the returned error string (even though not advised) which changes here, so technically there is a (breaking) behavior change with this PR (but not a breaking API change which is the main thing tracked in the semantic versioning). (This is more of a self note for release than a review comment, but the thought process is worth sharing.)

If this would be a named/marked/highlighted/individualized/interfaced error (which is encouraged to be built upon by downstream users) then I would even put it into the BREAKING CHANGES release category.
(But if implicit behavior change would cause BREAKING CHANGES then almost every release would be a BREAKING CHANGE which 95% of the time wouldn't require code modifications from downstream users which would be noisy and would defeat the purpose of using the BREAKING CHANGES indicator., so that's why I'm saying it is fine to be a patch IMO - everyone's flavor may differ.)

pregnor (approved):
You are now officially a hero of mine for getting this done.
These things irk me more than anything.

PR #3594, opened 2021-11-29
Spot price validation

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Added spotPrice validation to the request validation layer. The validator pulls the information via Cloudinfo on current prices in the AWS region for the instance type that is specified for each node group and returns a 400 Bad request response if the specified price is lower than the highest zone price.

### Why?

Giving a lower spotPrice than the minimum price results in the elongation and eventually the failure of cluster creation workflow which can be prevented this way. 

### Checklist


src/api/cluster_create.go
```diff
@@ -110,6 +110,8 @@ func (a *ClusterAPI) CreateCluster(c *gin.Context) {
 			createClusterRequest.SecretId = secret.GenerateSecretIDFromName(createClusterRequest.SecretName)
 		}
 
+		createClusterRequest.ProductDetailsGetter = a.productDetailsGetter
```
pregnor, line 113: **Question**: would it be much more complex if instead of proxying the product details getter here, we would retrieve the details (below the existing cluster checking part) and pass the static details to the validation function?

pkg/errors/errors.go
```diff
@@ -73,3 +76,35 @@ type ValidationBehavior struct{}
...
+		spotPrice:        spotPrice,
+		currentSpotPrice: currentSpotPrice,
+		location:         location,
+		instanceType:     instanceType,
+	}
+}
+
+func (e SpotPriceError) Error() string {
+	return fmt.Sprintf(
+		"spotPrice '%.4f' is too low, current minimum price in region '%s' for instance '%s' is '%.4f'",
```
pregnor, line 98: **Question**: wouldn't the error message be better to use location instead of region in case the user specifies a zone?

```suggestion
		"spotPrice '%.4f' is too low, current minimum price for location '%s' and instance '%s' is '%.4f' (zone price for zone locations, maximum of zone prices for regions)",
```

pkg/cluster/base.go
```diff
@@ -74,14 +80,15 @@ const (
...
-	Properties *CreateClusterProperties `json:"properties" yaml:"properties" binding:"required"`
+	Name                 string                   `json:"name" yaml:"name" binding:"required"`
+	Location             string                   `json:"location" yaml:"location"`
+	Cloud                string                   `json:"cloud" yaml:"cloud" binding:"required"`
+	SecretId             string                   `json:"secretId" yaml:"secretId"`
+	SecretIds            []string                 `json:"secretIds,omitempty" yaml:"secretIds,omitempty"`
+	SecretName           string                   `json:"secretName" yaml:"secretName"`
+	PostHooks            PostHooks                `json:"postHooks" yaml:"postHooks"`
+	Properties           *CreateClusterProperties `json:"properties" yaml:"properties" binding:"required"`
+	ProductDetailsGetter cloudinfo.ProductDetailsGetter
```
pregnor, line 91: **Note**: This feels like being in the wrong place here.
It should not be part of the request, because the request is supposed to be a descriptive type of the sent in request.
Let me think a bit about alternatives.

TODO: @pregnor

pkg/cluster/base.go
```diff
@@ -364,6 +374,86 @@ func (r *CreateClusterRequest) validateMainFields() error {
...
+		}
+	}
+
+	return errors.Combine(errs...)
+}
+
+func (r *CreateClusterRequest) getSpotPriceMap(spotPrice, cloud, service, instanceType string) (map[string]float64, error) {
+	spotPriceFloat, err := strconv.ParseFloat(spotPrice, 64)
+	if err != nil {
+		return nil, errors.WrapWithDetails(err, "could not parse spotPrice string to float64")
```
pregnor, line 436: **Required**: we **MUST** add the raw spot price string to the error details here for easier debugging.

pkg/cluster/base.go
```diff
@@ -364,6 +374,86 @@ func (r *CreateClusterRequest) validateMainFields() error {
...
+	}
+
+	prices := make(map[string]float64)
+	prices[desiredSpotPrice] = spotPriceFloat
```
pregnor, line 440: **Question**: why are we adding the desired spot price to the map of spot prices here?

It is available from earlier, feels like unnecessary addition.
Otherwise the `maxZonePrice` could be replaced by the region name and then we wouldn't need globals, because the region is also available in case the request is region based thus we don't need extra keys for these things.
Also if we do use global extra keys they should be named appropriately (maxZonePrice as a global makes no sense to be set to the string `maxZonePrice`, it should be `maxZonePriceKey` or `maxZonePriceMapKey` to disambiguate, but I would only go for globals if I have to and this case it feels like we are only using these from 1 function here, and maybe where the map is processed, but there the necessary data such as desired spot price or the region in case of my proposal are already available, so no need for extra keys, region can be passed in to this function and the desired should not be put in the spot prices map, the spot prices map should contain the actual values, excluding the manual desired one IMO).

pkg/cluster/base.go
```diff
@@ -364,6 +374,86 @@ func (r *CreateClusterRequest) validateMainFields() error {
...
+			spotPriceMap, err := r.getSpotPriceMap(np.SpotPrice, cloud.Amazon, EKS, np.InstanceType)
+			if err != nil {
+				return err
+			}
+
+			if spotPriceMap[desiredSpotPrice] < spotPriceMap[maxZonePrice] {
+				errs = append(errs, pkgErrors.NewSpotPriceError(spotPriceMap[desiredSpotPrice], spotPriceMap[maxZonePrice], r.Location, np.InstanceType))
+			}
+		}
+	}
```
pregnor, line 428: **Required**: as I mentioned in private, this very big 2 if branch can be simplified into a single for loop, let's discuss it offline.

pregnor (changes requested):

src/api/cluster_create.go
akijakya: I think we can only do it at the nodepool level because the detail getter method needs the exact instance type, zone and spotPrice and this can be differ between node pools within the same cluster.

pkg/errors/errors.go
akijakya: Ok, it is clearer this way!

pkg/cluster/base.go
akijakya: I agree this is not the perfect place, the easy alternative would be to pass down the `productDetailsGetter` as function parameter all the way where we would use it...

pkg/cluster/base.go
pregnor: Discussed, for the time being there is no easier alternative because of a bunch of different types.

PR #3602, opened 2022-01-26
Updated 1.20.9,0.21.3->0.21.9

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Updated k8s lib 1.20.9,0.21.3->0.21.9.

### Why?

To consolidate versions that have been partially bumped by a recent helmp update (#3599).

### Checklist


go.mod
```diff
@@ -127,31 +127,31 @@ replace (
...
-	k8s.io/kubectl => k8s.io/kubectl v0.21.3
-	k8s.io/kubelet => k8s.io/kubelet v0.21.3
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.21.3
-	k8s.io/metrics => k8s.io/metrics v0.21.3
-	k8s.io/mount-utils => k8s.io/mount-utils v0.21.3
-	k8s.io/node-api => k8s.io/node-api v0.21.3
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.21.3
-	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.21.3
-	k8s.io/sample-controller => k8s.io/sample-controller v0.21.3
+	k8s.io/api => k8s.io/api v0.21.9
```
pregnor, line 130: Question: why are we replacing to the same version as defined in the requirement?
Is there any reason not to drop the replacement where the required version is already the same?

akijakya: Well, as I could see in previous versions this was always the case so I looked at it as a collection of all k8s libs to ensure they are all on the same version, without checking what they are replacing (easier to maintain). But that is just something I assumed, do you think it would be cleaner to drop the ones that are already on the same version?

pregnor: I think we are double booking now. Once at the require and secondly at the replace. I concur it sometimes more clear to see the replacements be on the same version than the implicit requirements so we can leave it in, thanks.

pregnor (approved):

PR #3612, opened 2022-03-31
Fix integrated service tests

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

- Fixed `hashFiles` function in is-test workflows.
- Updated versions of services in test environment to be in sync with what is being used in the dev env.
- Kubernetes version 1.21.1 has been specified in the Makefile to be used by kind when creating a cluster to run tests against.

### Why?

In a recent change hashfiles function throws an error instead of failing silently causing these is-test workflows to fail.
Also the latest Ubuntu images that are used in the Github actions use the latest Kind version which starts kubernetes v1.23.4 version clusters by default, causing the integrated services v2 test to fail. 

### Checklist

### DIFF
--- a/.github/workflows/is-test-v2.yml
+++ b/.github/workflows/is-test-v2.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/.cache/go-build
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/*.go', '**/go.mod', '**/go.sum') }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('*.go', 'go.mod', 'go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-build-
 
--- a/.github/workflows/is-test.yml
+++ b/.github/workflows/is-test.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/.cache/go-build
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/*.go', '**/go.mod', '**/go.sum') }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('*.go', 'go.mod', 'go.sum') }}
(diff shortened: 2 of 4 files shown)


.github/workflows/is-test-v2.yml
```diff
@@ -22,11 +22,17 @@ jobs:
...
+        run: |
+          echo "::set-output name=date::$(/bin/date -u "+%Y-%m")"
+        shell: bash
+
       - name: Go build cache
         uses: actions/cache@v2
         with:
           path: ~/.cache/go-build
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/*.go', '**/go.mod', '**/go.sum') }}
+          key: ${{ runner.os }}-go-build-${{ steps.get-date.outputs.date }}
```
pregnor, line 35: I think these changes are going to implement a different kind of behaviors than expected.

The old behavior was supposed to create files based on the source code and go dependency files and update them if they change, otherwise use the cached version when they are similar to thefiles used in the previous runs thus not downloading the same dependency packages again, instead using them from cache.

The new behavior seems to replace the content-based hash with a date which is the current date in month if I see it correctly which is always going to use the current year and month for caching purposes.

That means,
1. every month the cache will be invalidated and the dependencies will be downloaded again (smaller issue), and
2. every time the source code or the go dependency files change, the dependencies are **NOT** going to be updated, until a new month begins (because of the date based caching), thus the dependency changes will not take effect.

I think this is going to cause further issues down the line - unless I misunderstood something about the caching here - so I think we should reconsider our approach here based on what problem we are trying to solve.

akijakya: I agree, it was only one of my many attempts to fix the original problem which is clearly not solved yet so it is a work in progress 🙂

pregnor (approved):

PR #3627, opened 2022-08-10
refactor: TestBlockingTags

Abirdcfly:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | 

### What's in this PR?

refactor the TestBlockingTags

### Why?

Initially, this test was found to be faulty because the subsequent code was unreachable because of `t.FailNow()` in [internal/secret/restricted/store_test.go](https://github.com/banzaicloud/pipeline/pull/3627/files/5c370cf8013765d9e0f195925611e5e932ab71cd#diff-1146d372385f5b72c7c361f87ae0958d6bfd62ad5232eed7439b9fb7165d54bd) line 55, and during the modification process, it was found that the test should be refactored directly.

### Checklist

### DIFF
--- a/internal/secret/restricted/store_test.go
+++ b/internal/secret/restricted/store_test.go
@@ -28,11 +28,12 @@ const (
 
 func TestBlockingTags(t *testing.T) {
 	cases := []struct {
-		name    string
-		request *secret.CreateSecretRequest
+		name      string
+		request   *secret.CreateSecretRequest
+		expectErr error
 	}{
-		{name: "readonly", request: &requestReadOnly},
-		{name: "forbidden", request: &requestForbidden},
+		{name: "readonly", request: &requestReadOnly, expectErr: ReadOnlyError{SecretID: secret.GenerateSecretID(&requestReadOnly)}},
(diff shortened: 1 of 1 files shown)


internal/secret/restricted/store_test.go
```diff
@@ -45,28 +46,21 @@ func TestBlockingTags(t *testing.T) {
...
 			if err != nil {
-				t.Errorf("error during storing readonly secret: %s", err.Error())
-				t.FailNow()
+				t.Fatalf("error during storing readonly secret: %s", err.Error())
 			}
 
 			err = store.Delete(orgID, secretID)
 			if err == nil {
-				t.Error("readonly secret deleted..")
-				t.FailNow()
```
Abirdcfly, line 55: Initially, this test was found to be faulty because the subsequent code was unreachable because of `t.FailNow()` in line 55, and during the modification process, it was found that the test should be refactored directly

pregnor (approved):
LGTM

PR #3637, opened 2022-09-08
Option to disable PKE on Azure and AWS

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Added config options to disable starting PKE clusters on Azure or AWS. If PKE is disabled on either of them, the cluster creation request now returns 400 Bad Request with the message `Not supported distribution type`.

### Why?

It is nice to have more configuration options for Pipeline CP, also PKE on Azure is currently misbehaving, and we decided to drop support for now.

### Additional context

Tested locally.

A similar solution is going to be implemented in Cloudinfo to remove the disabled options from the GUI as well.

### Checklist

- ~OpenAPI and Postman files updated (if needed)
- [X] Related Helm chart(s) updated (if needed)

### DIFF
--- a/charts/pipeline/values.yaml
+++ b/charts/pipeline/values.yaml
@@ -170,7 +170,7 @@ configuration:
     #           default: true
 
   # Distribution settings
-  # distribution:
+  distribution:
 
   #   eks:
   #     # EKS Cloud Formation template location
@@ -190,12 +190,15 @@ configuration:
   #     # Enable create & update of EKS addons like coredns
   #     enableAddons: false
 
-  #   pke:
-  #     amazon:
+    pke:
+      amazon:
+        enabled: true
   #       globalRegion: us-east-1
   #       defaultImages: {}
   #       defaultNetworkProvider: "cilium"
   #       defaultNodeVolumeSize: 0 # GiB, 0/fallback: max(50, AMISize)
+      azure:
+        enabled: true
 
   # Database configuration
   database:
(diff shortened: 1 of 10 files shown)


charts/pipeline/values.yaml
```diff
@@ -196,6 +196,9 @@ configuration:
   #       defaultImages: {}
   #       defaultNetworkProvider: "cilium"
   #       defaultNodeVolumeSize: 0 # GiB, 0/fallback: max(50, AMISize)
+  #       disabled: false
+  #     azure:
+  #       disabled: false
```
pregnor, line 201: **Recommended**: We **SHOULD** use `enabled: true (default)` instead, because double negation increases cognitive complexity and enabled chart attributes are the norm throughout our Helm charts. These can be uncommented and left in to make sure the default is properly `enabled: true` without extra logic.

pregnor (changes requested):

charts/pipeline/values.yaml
akijakya: Done! 🙂

charts/pipeline/values.yaml
```diff
@@ -190,12 +190,15 @@ configuration:
...
 
-  #   pke:
-  #     amazon:
+    pke:
+      amazon:
   #       globalRegion: us-east-1
   #       defaultImages: {}
   #       defaultNetworkProvider: "cilium"
   #       defaultNodeVolumeSize: 0 # GiB, 0/fallback: max(50, AMISize)
+        enabled: true
```
pregnor, line 199: **Recommendation**: let's move this enabled attribute to the top of the node.

Same in config/config.yaml.dist and internal/cmd/config.go

pregnor (approved):

charts/pipeline/values.yaml
akijakya: Valid remark, done!

pregnor (dismissed):

pregnor (approved):

PR #3638, opened 2022-09-13
PKE K8s version validation

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Added PKE Kubernetes version validation with hard-coded minimum and maximum version numbers. Starting a PKE cluster with higher or lower Kubernetes version than supported now results in a 400 Bad Request error response from the Pipeline API in the case of the following cloud providers:
- Amazon
- Azure
- vSphere

### Why?

Previously if someone started a PKE cluster with higher K8s version than available, there were no feedback from the API that this is not allowed, although the cluster creation failed later on.

### Checklist

### DIFF
--- a/internal/pke/BUILD.plz
+++ b/internal/pke/BUILD.plz
@@ -7,6 +7,9 @@ go_library(
     visibility = ["PUBLIC"],
     deps = [
         "//internal/global",
+        "//pkg/errors",
+        "//third_party/go:emperror.dev__errors",
+        "//third_party/go:github.com__Masterminds__semver__v3",
         "//third_party/go:github.com__sirupsen__logrus",
     ],
 )
--- a/internal/pke/kubernetes.go
+++ b/internal/pke/kubernetes.go
@@ -18,16 +18,42 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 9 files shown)


pregnor (approved):
This PR looks okay to me, it can be merged.

However I would ask you András to think about whether we can improve the error message for the unsupported k8s version to include min and max versions related to the distribution used, because that could result in a more helpful error message where the boundaries and also the set k8s version are reported for clarity purposes. If we want to and can do that, it is sufficient to put it into another PR.

pregnor (dismissed):

akijakya:
> This PR looks okay to me, it can be merged.
> 
> However I would ask you András to think about whether we can improve the error message for the unsupported k8s version to include min and max versions related to the distribution used and the user specified value, because that could result in a more helpful error message where the boundaries and also the set k8s version are reported for clarity purposes. If we want to and can do that, it is sufficient to put it into another PR.

Valid point, implemented! 🙂

pregnor (approved):

PR #3650, opened 2022-10-21
Updated golangci-lint version 1.30.0->1.50.0

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Updated golangci-lint to latest version (1.30.0->1.50.0).

### Why?

So we use a version that has build for darwin/arm64 systems for local linting.

### Additional context

There were a lot of linting errors initially, as a lot of new linters and linter updates have been introduced since the previous versions we used.

Followed strategy:
- if a linter was existing in the time of the previously used linter version (1.30.0), errors were fixed
- if a linter was deprecated, the linter was disabled and removed in code
- if a linter was newly introduced, it was disabled for now.

### Checklist


src/api/organization.go
```diff
@@ -189,6 +189,17 @@ func (a *OrganizationAPI) DeleteOrganization(c *gin.Context) {
...
+		message := "error getting organization: " + err.Error()
+		log.Info(message)
+		statusCode := auth.GormErrorToStatusCode(err)
+		c.AbortWithStatusJSON(statusCode, common.ErrorResponse{
+			Code:    statusCode,
+			Message: message,
+			Error:   message,
+		})
+	}
+
```
pregnor, line 202: **Note**: I think it is debatable whether we want to return an error on trying to delete a non-existent entity or just return a success, because the entity "is deleted" already.

pregnor (approved):

src/api/organization.go
akijakya: I guess the error here could be any kind of GORM error, like the connection to the database lost, which would be good to get... but if a `_` in place of the `err` would suffice here, I can change it, no problem 🙂

pregnor: We can leave it as is from me, were just curious whether we thought this through, but it is fine for me.

PR #3656, opened 2023-01-09
Updated code generators

akijakya:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| API breaks?     | no
| Deprecations?   | no
| Related tickets | -

### What's in this PR?

Updated packages that generate code, and ran them.

### Why?

To use the latest versions that have builds for darwin/arm systems so code generation could be done on modern Macs.

### Checklist

### DIFF
--- a/cmd/pipeline/main.go
+++ b/cmd/pipeline/main.go
@@ -437,7 +437,7 @@ func main() {
 		Debug:         true,
 		Servers: cloudinfoapi.ServerConfigurations{
 			{
-				URL:         url.Path,
+				URL: url.Path,
 			},
 		},
 		OperationServers: map[string]cloudinfoapi.ServerConfigurations{},


internal/security/anchore_client.go
```diff
@@ -180,16 +190,25 @@ func (a anchoreClient) DeleteAccount(ctx context.Context, accountName string) er
...
+	disabled := "disabled"
+	desiredState.State = &disabled
+
+	updateAccountRequest := a.getRestClient().UserManagementApi.UpdateAccountState(a.authorizedContext(ctx), accountName)
+	updateAccountRequest.DesiredState(desiredState)
 
-	if err != nil || ur.StatusCode != http.StatusOK || s.State != "disabled" {
+	s, ur, err := a.getRestClient().UserManagementApi.UpdateAccountStateExecute(updateAccountRequest)
+
+	if err != nil || ur.StatusCode != http.StatusOK || s.State != &disabled {
```
pregnor, line 202: **Question**: Is it intentional we are comparing the state pointers instead of their values? I would expect the pointers to mismatch even when they are pointing to the same values but this request execution could have a custom behavior implemented.

pregnor (approved):

internal/security/anchore_client.go
akijakya: Thanks for catching this, it would indeed make more sense!

pregnor (approved):

